PR #7, opened 2022-05-06
Remove unnecessary files from ui

ikusteu:
Remove some unnecessary stuff from `@librocco/ui` package. 

At least one ui component branch should be merged in before this one, as it deletes the dummy `Button` which we can use as reference while developing components. As the first ui component branch gets merged, the actual components can then be used as reference.


ikusteu:
Stale

Issue #12, opened 2022-05-08
Add lint-staged to pre-commit


ikusteu:
Using rush's pre-commit functionality, should add lint-staged. Can be a copy/paste from eisbuk. The lint-staged should actually run `lint` and `prettier --write`. EisBuk could be used for reference (or simple copy/paste) on the matter.

PR #14, opened 2022-05-09
Include `js-search`

silviot:
This PR adds an amended version of `js-search` to the repository.

The changes from [the upstream code](https://github.com/bvaughn/js-search) are:

* switch to Typescript (from flow)
* update Jest
* no support for building (since this incarnation is meant to be built by a depending package)


pkg/js-search/.gitignore
```diff
@@ -0,0 +1,8 @@
+.DS_Store
+/.bower
+/.idea
+/bower_components
+/node_modules
+/npm-debug.log
+/dist
+junit.xml
```
ikusteu, line 8: I would say that only `junit.xml` is necessary here, as `node_modules`, `npm-debug.log`, `.DS_Store` and `dist` are matched by root `.gitignore`'s `**/dist`, (...etc) and I don't think we're using bower, bower components, idk what `.idea` is for.

.github/workflows/js-search.yml
```diff
@@ -0,0 +1,34 @@
...
+          path: |
+            ~/.rush
+            ~/.pnpm-store
+            common/temp
+            **/node_modules
+          key: ${{ runner.os }}-modules-node16-v1-${{ hashFiles('**/pnpm-lock.yaml') }}
+      - name: Install packages
+        run: node common/scripts/install-run-rush.js update
+      - name: Run tests
+        run: cd pkg/js-search/ && ./node_modules/.bin/jest --ci --reporters=default --reporters=jest-junit
```
ikusteu, line 27: I would prefer if we had a `test` script in each package's `package.json` (like you've already created for js-search), so that they can be ran using a rush bulk command `rush test`. This would could, but doesn't have to interfere with this workflow: It would make sense to run test on all of the packages using the aforementioned bulk command, or this workflow could stay as it is, running the js-search-specific `rushx test` command.

ikusteu: Opened up a ticked related to the matter: #17

pkg/js-search/rollup.config.js
```diff
@@ -0,0 +1,38 @@
...
+    name: "emit-module-package-file",
+    generateBundle() {
+      this.emitFile({
+        type: "asset",
+        fileName: "package.json",
+        source: `{"type":"module"}`
+      });
+    }
+  };
+};
```
ikusteu, line 15: I would prefer if we used Vite for this (for consistencies sake): as Vite already uses rollup under the hook (among other things), the transition should be almost seamless.

pkg/js-search/source/IndexStrategy/ExactWordIndexStrategy.test.ts
```diff
@@ -0,0 +1,16 @@
+import { ExactWordIndexStrategy } from "./ExactWordIndexStrategy";
+describe('ExactWordIndexStrategy', function () {
+  var indexStrategy;
+  beforeEach(function () {
```
ikusteu, line 4: Nitpicky with the dose of `var` alergies:

I would replace all occurrences of `var` with `const` (probably using `find` + `sed`), run `tsc` to see if it complains about reassigning to a `const` variable, at which point, I would replace all error occurrences with `let`.

Furthermore, for future readability I would replace all callbacks with `() => { ... }` rather than `function () { ... }`.
I believe this could partly be done with `sed` as well...or even the eslint set up to prefer arrow function callbacks and run `eslint --fix` (I will look more into it)

Finally, this is a preference when writing new code, and since, I understand, this has been generated using some flow->TS transpiler, it's understandable if you've used the "If it works don't touch it" logic (and it most certainly does)

pkg/js-search/source/IndexStrategy/ExactWordIndexStrategy.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  it('should not expand empty tokens', function () {
+    var expandedTokens = indexStrategy.expandToken('');
+    expect(expandedTokens.length).toEqual(0);
+  });
+  it('should not expand tokens', function () {
+    var expandedTokens = indexStrategy.expandToken('cat');
+    expect(expandedTokens.length).toEqual(1);
+    expect(expandedTokens).toContain('cat');
+  });
+});
```
ikusteu, line 16: We really need prettier on pre-commit (issue already opened: #12 )

ikusteu (approved):
LGTM all in all. Left some nitpicky comments and suggestions.

pkg/js-search/source/IndexStrategy/ExactWordIndexStrategy.test.ts
silviot: The codebase is now allergen-free! Arrow functions are now preferred over `function () {}`.

pkg/js-search/rollup.config.js
silviot: Vite is now used to build (and also, building is now possible).

pkg/js-search/source/IndexStrategy/ExactWordIndexStrategy.test.ts
silviot: I upgraded `prettier` to the latest version and ran it on `js-search`.

silviot:
@ikusteu Thanks for the feedback! I did the changes you suggested.

pkg/js-search/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "files": [
+    "dist",
+    "source"
+  ],
+  "devDependencies": {
+    "@babel/core": "7.x",
+    "@babel/preset-env": "^7.8.7",
+    "babel-jest": "^25.1.0",
+    "jest": "^27.5.1",
+    "prettier": "^2.6.2",
```
ikusteu, line 22: I don't think prettier is necessary here. If you want format on save I believe you need to have the prettier installed globally (+ the extension). And if you don't use it, and just want to format pre commit, that will be done through pre-commit (with prettier installed in `common/autoinstallers/pre-commit` for that specific use case.

ikusteu: Sry: nitpicky minimalism

silviot: I'm not 100% convinced: if we rely on a system installed `prettier` we can't pin a specific version, so it might unpredictably introduce changes if different people have a different `prettier` version.

Anyway, I removed the dependency for now.

ikusteu: Yeah, but in [autoinstaller](https://github.com/librocco/librocco/blob/main/common/autoinstallers/lint-staged/package.json) we're specifying the right version (and that is the one used to run pre-commit). Locally installed (well globally but on the developer's machine) is only for development comvenience.

ikusteu: Yeah, but in [autoinstaller](https://github.com/librocco/librocco/blob/main/common/autoinstallers/lint-staged/package.json) we're specifying the right version (and that is the one used to run pre-commit). Locally installed (well globally but on the developer's machine) is only for development comvenience.

PR #15, opened 2022-05-09
Import CSV file into CouchDB

fadwamahmoud:
New module for couchdb APIs and create function that bulk inserts docs


pkg/couchdb/couchdb.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * @TODO handle all cases
+ *
+ * 201 Created – Document(s) have been created or updated
+ * 400 Bad Request – The request provided invalid JSON data
+ * 404 Not Found – Requested database not found
+ */
+
+/** @TODO type docs array according to input docs columns */
+export const bulkDocs = async (
```
silviot, line 33: Let's include the `HTTP` verb we're using in the function name: `postBulkDocs` better suggests what this funciton does.

pkg/scripts/csv-to-couchdb.ts
```diff
@@ -0,0 +1,164 @@
...
+/**
+  * Counts lines in a csv file for progress bar calculations
+  * @param {string} filePath - csv file path 
+  * @return {Number} - Number of lines in file
+  */
+
+const howManyLines = async function (filePath: string) {
+    let lineCount: number = 0;
+    var file;
+    console.log({ filePath })
```
silviot, line 18: This can be removed. Also, why print
```
{
    "filepath": "path/to/the/csv_file.csv"
}
```

instead of 
```
path/to/the/csv_file.csv
```
?

I think I get it: the logging is more descriptive this way.

pkg/scripts/csv-to-couchdb.ts
```diff
@@ -0,0 +1,164 @@
...
+    return lineCount;
+}
+
+//#region progress bar
+/**
+  * Renders progress bar
+  * @param {ProgressBar} progress - progress bar instance 
+  * @param {Number} progress - chunk size for incrementation
+  */
+let completed = 0;
```
silviot, line 37: The `completed` variable should not be a top level var, attached to the global object. It would be better to pass it to the `downloading` function as a parameter.

pkg/scripts/csv-to-couchdb.ts
```diff
@@ -0,0 +1,164 @@
...
+function downloading(progress: ProgressBar, chunkSize: number) {
```
silviot, line 38: This function prints a progress bar. It does this while the CSV file is being processed. There is no download going on here. A better name could be `printProgress`. It could take `completed` as a parameter, instead of `chunkSize`.

pkg/scripts/csv-to-couchdb.ts
```diff
@@ -0,0 +1,164 @@
...
+
+
+    try {
+        await createDatabase(couchdbName, couchdbURL)
+        const f = await Deno.open(filePath);
+        console.log("Saving docs bulk...")
+
+        for await (const chunk of inChunks(readCSV(f, options), chunkSize, columnsArr)) {
+            await bulkDocs(chunk, couchdbName, couchdbURL)
+            downloading(progress, chunkSize);
```
silviot, line 82: We could initialize and update a `completed` variable in this funciton, and pass it to ~`downloading`~ `printProgress`.

silviot (changes requested):

pkg/scripts/csv-to-couchdb.ts
fadwamahmoud: We could initialize `completed` here but it has to be updated inside the `printProgress` function, right? @silviot

silviot: I don't think `printProgress` should update a counter: it should only print info about the counter it receives. Updating the counter should happen here IMHO.

silviot (approved):

silviot (approved):

PR #18, opened 2022-05-19
Setup/pre commit & rush ci

### LINKED ISSUE #12: Add lint-staged to pre-commit

### LINKED ISSUE #17: Set up rush CI commands
We should set up "command-proxies" for `rush` and `rushx` in the CI so that we can run said commands in CI, without having to type `node common/scripts/install-run-rush.js <command>` and just use `rush <command>` instead. This should be a quick one as we've already solved this problem in eisbuk [like so](https://github.com/eisbuk/EisBuk/blob/master/common/scripts/rush)

ikusteu:
- adds `pre-commit` setup (lint staged + prettier write) Fixes #12 
- adds `rush` and `rushx` CI commands Fixes #17

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,10 +116,9 @@ To remedy this (and make our lives easier) we've created the "command proxies" f
 
 ```yaml
 # ...rest of the workflow setup (installing node, caching modules, etc.)
-    - name: Add rush directory to PATH
-        # Enable command proxies by adding their directory (command/scripts) to the PATH variable
-        run: echo PATH=$PWD/common/scripts/:$PATH >> $GITHUB_ENV
+- name: Add rush directory to PATH
+  run: echo "${PWD}"/common/scripts >> $GITHUB_PATH
 # ...rest of the workflow code
 ```
 
-This way, we don't need to worry about our scripts (or some other process' pre-deploy scripts and such) calling `rushx <command>` in the CI and are confident everything will work similarly to the way it works instance.
+This way, we don't need to worry about our scripts (or some other process' pre-deploy scripts and such) calling `rushx <command>` in the CI and are confident everything will work similarly to the way it works locally.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -98,3 +98,28 @@ Also, both are configured to remove the `width` and `height` attributes. Because
...
+- our `package.json` specified scripts would not be used the same way locally and CI: locally, our scripts could run `rushx <command>`, while in the CI we would need to specify the full path (like the one above). This would require us to use different scripts for local usage and respective CI versions
+
+To remedy this (and make our lives easier) we've created the "command proxies" for [rush](./common/scripts/rush) and [rushx](./common/scripts/rush). Both are ran using simple `rush` (and `rushx` respectively) commands, and in turn they run the corresponding `install-run-rush(x).js` passing on the arguments. We don't use this locally, but run globally installed `rush` (and `rushx`) commands. For CI, we simply need to enable the provided "command proxies" by adding them to the `PATH` variable of CI process. That is done by adding something like this in the workflow setup:
+
+```yaml
+# ...rest of the workflow setup (installing node, caching modules, etc.)
+    - name: Add rush directory to PATH
+        # Enable command proxies by adding their directory (command/scripts) to the PATH variable
+        run: echo PATH=$PWD/common/scripts/:$PATH >> $GITHUB_ENV
```
silviot, line 121: This `run` key is too much to the right: it should be indented less.
Also, [in a different branch](https://github.com/librocco/librocco/pull/14/commits/ba994f8199c6504847ba7fe209752efe2b1b5a5a#diff-514696a0fb5aa77266f335d28975e7f6a374863e6cb7d55e4a45db7ebbc1456cR25) I used the `echo "${PWD}"/common/scripts >> $GITHUB_PATH` command. It's slightly easier than this IMO.

README.md
```diff
@@ -98,3 +98,28 @@ Also, both are configured to remove the `width` and `height` attributes. Because
...
+# ...rest of the workflow code
+```
+
+This way, we don't need to worry about our scripts (or some other process' pre-deploy scripts and such) calling `rushx <command>` in the CI and are confident everything will work similarly to the way it works instance.
```
silviot, line 125: Probably should be "the way it works locally" at the end of this line.

silviot (approved):

PR #58, opened 2022-05-26
NoteButton component

### LINKED ISSUE #46: Feature: NewNoteButton
## Component: NewNoteButton
**Description:** 
- Button for adding new Notes.
- Two variants: 
  a. IN Button
    - Requires Warehouse name to be passed onClick, so that Note is added to correct Warehouse context
  b. OUT button
    - Add global OUT Note => does not require Warehouse context

**Mockup**: TODO: Standard component reference in Figma. Note: For now refer to green button in [this view](https://www.figma.com/file/yjspFXBKBIFJlC3wFv1FJL/The-Brocco?node-id=101%3A4320), and red button in [this view](https://www.figma.com/file/yjspFXBKBIFJlC3wFv1FJL/The-Brocco?node-id=101%3A5155) for variants.

### Requirements:
**Props:**
- variant/style: Enum of type
- Label: Button name. For IN Notes = Warehouse, for OUT Notes can = "New Note" | "Remove Books"
- onClick handler

**Styles:**
- IN = green
- OUT = red
 - OUT on UtilityNav has rounded corners on all sides

**Storybook**
- Default

**Functionality**
Adds a new Note

**Possible Enhancements**
Null

fadwamahmoud:
Fixes #46

### DIFF
--- a/pkg/ui/src/components/NoteButton/NoteButton.stories.tsx
+++ b/pkg/ui/src/components/NoteButton/NoteButton.stories.tsx
@@ -1,0 +1,37 @@
+import React from "react";
+import { ComponentMeta } from "@storybook/react";
+
+import { StorybookGrid, StorybookItem } from "../../utils/storybook";
+
+import NoteButton, { NoteButtonType, NoteButtonLabel } from "./NoteButton";
+
+export default {
+  component: NoteButton,
+  title: "NoteButton",
+} as ComponentMeta<typeof NoteButton>;
+
+export const Default = (): JSX.Element => (
+  <>
+    <h1 className="text-lg font-bold mb-4">Note In:</h1>
+    <StorybookGrid className="mb-16">
+      <StorybookItem>
+        <NoteButton type={NoteButtonType.In}></NoteButton>
+      </StorybookItem>
+    </StorybookGrid>
+    <h1 className="text-lg font-bold mb-4">Note Out:</h1>
+    <StorybookGrid className="mb-16">
+      <StorybookItem>
+        <NoteButton type={NoteButtonType.Out}></NoteButton>
+      </StorybookItem>
+    </StorybookGrid>
+    <h1 className="text-lg font-bold mb-4">Global Out:</h1>
+    <StorybookGrid>
+      <StorybookItem>
+        <NoteButton
+          type={NoteButtonType.Out}
+          label={NoteButtonLabel.RemoveBooks}
+        />
+      </StorybookItem>
+    </StorybookGrid>
+  </>
+);
--- a/pkg/ui/src/components/NoteButton/NoteButton.stories.tsx
+++ b/pkg/ui/src/components/NoteButton/NoteButton.stories.tsx
@@ -1,9 +1,11 @@
 import React from "react";
 import { ComponentMeta } from "@storybook/react";
 
-import { StorybookGrid, StorybookItem } from "../../utils/storybook";
-
-import NoteButton, { NoteButtonType, NoteButtonLabel } from "./NoteButton";
+import NoteButton, {
+  NoteButtonType,
+  NoteButtonLabel,
+  NoteButtonShape,
+} from "./NoteButton";
 
 export default {
   component: NoteButton,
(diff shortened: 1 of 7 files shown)


pkg/ui/src/components/NoteButton/NoteButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+  [NoteButtonType.Out]: ["text-white", "bg-red-400", "hover:bg-red-500"],
+};
+
+// Classes for focus pseudo-class are the same regardless of color prop
+const focusClasses = [
+  "focus:ring-2",
+  "focus:ring-indigo-500",
+  "focus:ring-offset-2",
+  "focus:ring-offset-white",
+];
```
ikusteu, line 42: This seems unnecessary as it's not specified in the spec (nor used in this context I reckon) and (I'm not a designer, but) I think this color, indigo, wouldn't really work well with the other colors)

pkg/ui/src/components/NoteButton/NoteButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+ * @param {JSX.Element | JSX.Element[]} props.children
+ */
+const NoteButton: React.FC<NoteButtonProps> = ({
+  children,
+  type = NoteButtonType.In,
+  label = Warehouse.Varia,
+  onClick = () => {},
+  ...props
+}) => {
+  const size = "md";
```
ikusteu, line 63: TLDR: This doesn't achieve anything.

More in-depth explanation: You've probably seen spacing in `tailwind.config.js`. Extending spacing with new key-values means: Extend all spacing classes to include the provided key. In this case:
```js
// ...theme, extend and all that jazz
{
  spacing: {
    md: "1024px"
  }
}
```
creates new css classes:
```css
.w-md {
  width: 1024px;
};
// ...
.h-md {
  height: 1024px;
};
// ...
.p-md {
  padding: 1024;
};
// ...
.ml-md {
  margin-left: 1024px;
}
// ...etc
```

ikusteu: and all variants of tailwind utility classes related to spacing (margin, width, height, padding, left, right, top, etc..) can be addressed using `<one-of-spacing-classes>-md` and provide the `1024px` as a value for the given property.

ikusteu: In order to use it you need to pair it with utility prefix (`m-`, `ml-`, `mr`, `p-`, ...)

pkg/ui/src/components/NoteButton/NoteButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const shape =
+    label === NoteButtonLabel.RemoveBooks ? "rounded-md" : "rounded-r-md";
```
ikusteu, line 65: kudos 👍 However, I would rather generalise the shape property, but I will update the spec. (my bad)

pkg/ui/src/components/NoteButton/__tests__/NoteButton.test.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+
+import NoteButton, { Warehouse } from "../NoteButton";
+
+describe("An example unit test", () => {
```
ikusteu, line 6: Thank you for providing this example 😂

pkg/ui/src/components/NoteButton/index.ts
```diff
@@ -0,0 +1,3 @@
+import NoteButton from "./NoteButton";
+
+export default NoteButton;
```
ikusteu, line 3: You can also add `export * from "./NoteButton"` here as you want the consumer of `NoteButton` (in web package perhaps) be able to use the above specified enums (NoteButtonType, etc.) and it's always ok to export the Props interface for a component.

pkg/ui/src/components/NoteButton/NoteButton.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+} as ComponentMeta<typeof NoteButton>;
+
+export const Default = (): JSX.Element => (
+  <>
+    <h1 className="text-lg font-bold mb-4">Note In:</h1>
+    <StorybookGrid className="mb-16">
+      <StorybookItem>
+        <NoteButton type={NoteButtonType.In}></NoteButton>
+      </StorybookItem>
+    </StorybookGrid>
```
ikusteu, line 20: No need to use the grid if there's only one element in the row (or grid for that matter)

ikusteu: This can just be `<NoteButton {...}>`

ikusteu (changes requested):
I'm not too happy with the spec, but it was implemented more/less ok. Aside from the small stuff I mentioned, there are two points you should fix:
- You need to add an export for `NoteButton` in `pkg/ui/index.ts` so that the lib bundles it (and it can be used in other packages)
- You need to pass `className` from props to the wrapping element of `NoteButton` so that we can position it at least

After the small fixes are made, this can go in, AFAIC, as it's ok as a first pass and we can update later.

pkg/ui/src/components/NoteButton/NoteButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+export enum NoteButtonType {
+  In = "in",
+  Out = "out",
+}
+
+export enum Warehouse {
+  Scolastica = "scolastica",
+  Nuovo = "nuovo",
+  Varia = "varia",
+}
```
delasous, line 27: Warehouses are not limited to these names, they can openly be defined by the users

ikusteu: I completely agree and have mentioned it in #46, but would really love to merge this in ASAP, so that we can possibly add the updates/corrections on svelte-rewritten version of this component.

pkg/ui/src/components/NoteButton/__tests__/NoteButton.test.tsx
fadwamahmoud: guilty of copy/pasting :joy:

Issue #59, opened 2022-05-26
Enhancement: PaginationLinks


ikusteu:
## Component: PaginationLinks

**Description:** PaginationLinks should be an enhancement on [Pagination](https://github.com/librocco/librocco/issues/23) component. The core `Pagination` is stateless (and lets the display be controlled through props and the state be kept outside). This was convenient for multiple reasons (stories are easier to display), simpler testing and as a result, more generic component. `PaginationLinks` should wrap the `Pagination` component and provide the stateful logic using React router and pass the props to `Pagination`

**Mockup:** Purely stateful wrapper, no style involved, for Mockup see #23 

#### Props:

The props are a subset of #23 props and should get passed down to the `Pagination` component

- maxItems (optional: number): 
- links 
- className (optional: string):

#### Storybook (stories):

- null

#### Functionality (should be tested):

- Route matching: 
  - should match the pathname from URL with appropriate entry in `links` array and infer `currentItem`
- Routing:
  - routing should be done by passing the React router `Link` component as a `Wrapper` to `Pagination` component

### Possible enhancements:

- null

PR #60, opened 2022-05-26
Feature/pagination

### LINKED ISSUE #23: Feature:  Pagination (NumberLinks)
## Component: Pagination

**Description:** A navbar with numbers for links (used for some sort of pagination). The first pass should be stateless (both `onButtonClick` and `currentItem` should be passed as props). As an enhancement, we can add a component wrapping the stateless one with React router (route matching and `Link` as element `Wrapper`). Implements #40, but no need for separate component as the #40 is defined with a simple set of classNames and trivial button functionality.

**Mockup:** [Figma](https://www.figma.com/file/yjspFXBKBIFJlC3wFv1FJL/The-Brocco?node-id=20%3A2337)

#### Props:

- maxItems (optional: number):
  - default: 7
  - I think the minimum should be 5, as 3 would just show the current and the dots
  - should probably fall down to previous odd number in case an even number has been provided
- links (required: Array<string>):
  - if none provided - component is not rendered
- wrapper (optional: React component):
  - to be used to possibly wrap each element with react router Link (or such)
  - should be able to accept `to`, `disabled` (as those props will be passed from the component to each wrapping element, according to internal state)
- onButtonClick (optional: Function):
  - should accept `linkTo: string` and `i: number` (0 - based index of clicked button/link)
- className (optional: string):
  - shouldn't override anything from the core component (enhancement: only accept positioning)
- currentItem (number):
  - index (0 - based) of currently selected item

#### Storybook (stories):

- Default
- Try out different combinations with different `...` positions as that functionality might be a bit tricky

#### Functionality (should be tested):

- rendering:
  - should respect the `maxItems` prop while rendering
  - should display odd number of items (fall back to `maxItems - 1` if the odd number was provided)
  - should throw on `maxItems < 5`
  - should hide the correct links behind `...` (if more than `maxItems` provided)
- onChange:
  - should call the `onChange` function with the right value for `url` on link click
  - should call the `onChange` function with the right value for `url` on next/previous button click
- wrapper:
  - should wrap the element and pass the right values for `disabled` and `to`

### Possible enhancements:

- create a helper to filter all non position-related Tailwind classes
- create a `PaginationLinks` component wrapping `Pagination` into react router logic

ikusteu:
Fixes #23 
Without the enhancements:
- #59


ikusteu:
This seems to already be in. I even wrote this in svelte as part of ui rehaul.

PR #74, opened 2022-05-29
Enhancement/ci 1

### LINKED ISSUE #69: Create a storybook CI workflow
Initially, storybook should be built on PR using GitHub action workflow, to assure there are no issues with storybook build.

As an enhancement, this workflow should also publish the build to a visual regression testing tool (we can use Chromatic for starters and then maybe migrate to self hosted version)

### LINKED ISSUE #70: Add pkg/ui tests to test workflow
- [ ] Merge `js-search` workflow with linting workflow as a workflow which will run on push and PR
- [ ] Add a job to test the `ui` package

ikusteu:
Fixes #70 
Fixes #69


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,110 @@
+name: Build and test packages
+
+on:
+  - push
+  - pull_request
```
silviot, line 5: I did this with Eisbuk but it's probably useless: `push` should be enough. My understanding of what is checked out when the target is `pull_request` has changed: I used to think that the result of the PR merge was checked out, but now I believe some manual work would be needed to achieve that (and it wouldn't work in case of conflicts).

ikusteu:
Closing as it has been addressed in #88

PR #89, opened 2022-08-18
Data model tests

fadwamahmoud:


pkg/web/db-actions.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const newNote = (note: Note) => {
+  const currentDatabase = getDB();
+
+  const newDB = { notes: [...currentDatabase.notes, note] };
+
+  setDB(newDB);
+
+  const promise = new Promise<{ id: string; ok: boolean }>((resolve) => {
```
silviot, line 29: This works, but I feel it would be more readable in an `async` function, as

```
return { id: note._id, ok: true };
```

pkg/web/db-actions.ts
```diff
@@ -0,0 +1,78 @@
...
+    currentDatabase.notes.find((note) => note._id === id) || ({} as Note);
+
+  const promise = new Promise<Note>((resolve) => {
+    resolve(note);
+  });
+
+  return promise;
+};
+
+// set DB
```
silviot, line 49: Instead of having a "currently active" DB we could have a function `createDB` that returns an empty database.
The returned object would then be needed to invoke `newNote`, `updateNote` etc.

pkg/web/db-actions.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  setDB({ notes: [...newDatabase.notes, newNote] });
+
+  const promise = new Promise<{ id: string; ok: boolean }>((resolve) => {
+    resolve({ id, ok: true });
+  });
+  return promise;
+};
+
+export const deleteDatabase = () => {
```
silviot, line 72: I don't think we need this.

For in-memory database (when the DB is a Javascript object) we'll let the garbage collector pick it up (or not - it's not the end of the world).
For PouchDB/CouchDB we can just leave the DB there (expecting the db server to be ephemeral and/or expecting the user to clean it up if they so desire).

pkg/web/pouchdb-actions.ts
```diff
@@ -0,0 +1,46 @@
...
+  title: string;
+  author: string;
+  editor: string;
+  price: number;
+  quantity: number;
+  note?: string;
+  date?: string;
+}
+
+export const database = new Pouchdb("notes");
```
silviot, line 19: I think database creation should be part of a `createDB` function. It can use a random ID to make sure an empty database is returned on each invocation.

silviot:
Good work! We're getting there!
I believe the API for `updateNote` should change. [I left a comment about it here](https://github.com/librocco/librocco/discussions/85#discussioncomment-3423600).

I think a good test case could be:

* List all notes - make sure there's none
* Create a note
* List all notes - make sure there's one
* List all books in the newly created note - make sure there's none
* Add a book to the note
* List all books in the newly created note - make sure there's one
* Add another book to the note
* List all books in the newly created note - make sure there are two
* Remove a book from the note
* List all books in the newly created note - make sure there's one, with the expected ISBN
* Remove a non-present book from the note - make sure we get an exception

pkg/scaffold/.eslintrc.js
```diff
@@ -149,6 +149,9 @@ module.exports = {
     // we're using empty functions as fallback for undefined props
     "@typescript-eslint/no-empty-function": "off",
 
+    // For pouchdb and couchdb snake case conventions
+    "@typescript-eslint/camelcase": "off",
+
```
ikusteu, line 154: Can we do this on a per-line/per-file basis, rather than in all of the packages in the entire monorepo?

rush.json
```diff
@@ -386,6 +386,10 @@
     { "packageName": "@librocco/svg", "projectFolder": "pkg/svg" },
     { "packageName": "@librocco/ui-react", "projectFolder": "pkg/ui-react" },
     { "packageName": "@librocco/shared", "projectFolder": "pkg/shared" },
-    { "packageName": "js-search", "projectFolder": "pkg/js-search" }
+    { "packageName": "js-search", "projectFolder": "pkg/js-search" },
+    {
+      "packageName": "@librocco/web",
+      "projectFolder": "pkg/web"
+    }
```
ikusteu, line 393: 1. I feel that the db interfaces can be their own separate package, only assembled in the app. 
2. The `pkg/web` is deprecated and will be replaced with `apps/web-client`, so this should be moved out (my suggestion is 1. ☝️ )

I can help out with setting up of the package for db interfaces.

ikusteu (changes requested):
Didn't go through the code line by line. I like the class usage for the variable db (ability to create as many new ones as needed).
I wrote some repo related comments, which I'd like addressed before this is merged in. I'm happy to help out.

pkg/database/src/types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export interface BookInterface {
+	isbn: string;
+	title: string;
+	author: string;
+	editor: string;
+	price: number;
+	quantity: number;
+	note?: string;
+	date?: string;
```
ikusteu, line 14: This is a bit loose. We can create two interfaces: 
- `BookStock` (I believe that was the term @delasous suggested and we've agreed on it): This interface without the optional fields
- `BookStockInNote` in: Extends `BookStock` and adds required `note` and `date`

ikusteu (approved):
LGTM

pkg/database/src/types.ts
silviot: After today's discussion we settled on `BookData` for title, author, etc. and `Volume`, when it also has date, quantity and an indication of where the book is (might be one of {`note`, `location`, `where`}).

silviot (approved):

PR #95, opened 2022-08-21
Feature/test ci ytt templates

ikusteu:
I've created workflow template for an entire test job. 
I've included the `rush build` step in all test jobs generated by a template as `js-search` already does a build check, and build with rush gets parallelised as it is, so I don't think there's too much of an overhead. 
I believe that using a conditional `needs_build` step is doable, but didn't find it necessary atm, @silviot let me know what you think.

### DIFF
--- a/.github/workflow.templates/on-push.yml
+++ b/.github/workflow.templates/on-push.yml
@@ -10,7 +10,7 @@ name: Cheks
   - push
 
 jobs:
-  js-search-test: #@ test_job("JS-Search Tests", "pkg/js-search")
+  js-search-test: #@ test_job("JS-Search Tests", "pkg/js-search", "jest-junit")
 
   ui-test: #@ test_job("UI Tests", "pkg/ui")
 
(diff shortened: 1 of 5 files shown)


.github/workflow.templates/on-push.yml
```diff
@@ -2,54 +2,18 @@
...
-      - uses: actions/setup-node@v2
-        with:
-          node-version: "16"
-      -  #@ cache_node()
-      -  #@ rush_add_path()
-      -  #@ rush_install()
-      -  #@ rush_build()
-      - name: Run tests
-        run: cd pkg/ui && rushx test:ci
+  js-search-test: #@ test_job("JS-Search Tests", "pkg/js-search", "jest-junit")
```
silviot, line 13: Very small thing, and debatable. But I'd like to point out that this can also be expressed by providing parameter names for improved readability:
```
  js-search-test: #@ test_job("JS-Search Tests", workdir="pkg/js-search", reporter="jest-junit")
```

silviot (approved):

PR #99, opened 2022-09-23
Feature/data model tests

ikusteu:
Here it is. Please go over it and let me know if it's unclear/hard to use and if you find something missing (along with the standard PR review stuff).

TODOs for second pass:
- add a way to load design documents into the test db
- add benchmarking
- create and integrate a docker couchdb image and use that data as test data


pkg/datamodel-tests/src/implementations/version-1.1/note.ts
```diff
@@ -0,0 +1,105 @@
+import { VolumeStock, VolumeTransactionTuple } from '@/types';
+import { sortBooks } from '@/utils/misc';
+import { NoteInterface, WarehouseInterface, NoteData } from './types';
+
+export class Note implements NoteInterface {
```
silviot, line 5: Should we export the class? It means that code importing `Note` needs to use `new`, which might be unexpected. I think it might be good to consider this an implementation detail, and use `newNote` as the public API.

ikusteu: Makes sense...I completely missed that, and I had the `newNote` function to begin with...the export is completely unnecessary.

PR #101, opened 2022-09-27
Initial version of TextFIeld component

### LINKED ISSUE #63: Feature: TextField
## Component: TextField

**Description:** A styled text input field as specified in the mockup. 

**Mockup:** [Figma](https://www.figma.com/file/yjspFXBKBIFJlC3wFv1FJL/The-Brocco?node-id=21%3A2629)

### Requirements:

Should accept state through props and be agnostic of the environment (used within Formik as well as a standalone), but shoud conform to the Formik's spec for the field, for seamless integration.

#### Props (should extend React.HTMLAttributes<HTMLInputElement>): 

- **startAdornment** (optional: string or React.FC):
  - shown inside the field, before the text
  - could be string (e.g. `https://`), React.FC (e.g. additional dropdown, as shown in figma, or a simple icon as an SVGComponent)
- **endAdornment** (optional: string or React.FC):
  - shown inside the field, after the text
  - see **startAdornment** ☝️ ☝️ 
- **startAddon** (optional: string or SVGComponent):
  - shown as an add-on before the field itself
- **endButton** (optional: { **label**: string, **onClick**?: Function, **type**?: string for button type, **Icon**?: SVGComponent } ):
  - used to render a button as an add-on after the field itself
- **error** (optional: string): if present, the field should be red (signaling validation error state), and the show the error icon before the **endAdornmant** on icon hover, the error message should be shown as alt-text @delasous please chime in here
- **meta** (optinal: string): if present, the question mark icon should be shown before the **endAdornment**  and the meta message shown on icon hover (as alt-text). Error icon ☝️ ☝️  should take precedence over meta icon. @delasous please chime in here

#### Storybook (stories):
- Default
- StartEndAdornment
- StartAddOn (string, icon, dropdown)
- EndButton

#### Functionality (should be tested):
- onChange

### Possible enhancements:
- null

### Progress:
**TODO:**
- [ ] Create a brutalist component with appropriate functionality
- [ ] Implement styles

fadwamahmoud:
Fixes #63

### DIFF
--- a/apps/web-client/svelte.config.js
+++ b/apps/web-client/svelte.config.js
@@ -8,11 +8,7 @@ const config = {
 	preprocess: preprocess(),
 
 	kit: {
-		adapter: adapter({ fallback: 'index.html' }),
-
-		prerender: {
-			default: false
-		}
+		adapter: adapter({ fallback: 'index.html' })
 	}
 };
 
--- a/common/autoinstallers/lint-staged/package.json
+++ b/common/autoinstallers/lint-staged/package.json
@@ -15,6 +15,8 @@
     "eslint-plugin-unused-imports": "^1.0.1",
     "lint-staged": "^12.3.8",
     "prettier": "^2.6.2",
+    "prettier-plugin-svelte": "^2.7.0",
+    "svelte": "^3.44.0",
     "typescript": "~4.4.2"
   }
 }
--- a/common/autoinstallers/lint-staged/pnpm-lock.yaml
+++ b/common/autoinstallers/lint-staged/pnpm-lock.yaml
@@ -13,6 +13,8 @@ specifiers:
   eslint-plugin-unused-imports: ^1.0.1
   lint-staged: ^12.3.8
   prettier: ^2.6.2
+  prettier-plugin-svelte: ^2.7.0
+  svelte: ^3.44.0
   typescript: ~4.4.2
 
 dependencies:
@@ -28,6 +30,8 @@ dependencies:
   eslint-plugin-unused-imports: 1.1.5_bf92fac4e3857d5b28bcad3e8c0922df
   lint-staged: 12.4.1
   prettier: 2.6.2
+  prettier-plugin-svelte: 2.7.1_prettier@2.6.2+svelte@3.50.1
+  svelte: 3.50.1
   typescript: 4.4.4
 
 packages:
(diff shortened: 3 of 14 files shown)


delasous (changes requested):
I have added a this [tailwind plugin](https://github.com/tailwindlabs/tailwindcss-forms) which resets the default styles of form elements. This was why the stories didn't look quite like the tailwindUI examples.

I have a few requests before this goes in:

### Props
1. I would avoid exporting all props with `{ ... }`. Its repetitive, but the `export let` before each prop helps it stand out as part of the components public interface. The script block isn't busy right now, but if it grows and we add more declarations within, it will hard to tell them apart from the internal ones.
2. `className` - let's avoid using this unless its unavoidable. As a user of the component its difficult to tell which element the classes will target / what effect they'll have mixing with the other styles defined within. In this case the `w-1/2` can be set on a parent element, the top div in TextField will just grow to fill it.
3. You can drop `string` type annotations with `= ""`, TS can figure this out.
4. `startAdornment` and `endAdornment` should be handled as [named slots](https://svelte.dev/tutorial/named-slots) not props.
5. `onChange` is exported but not passed to the markup? I see that the OG spec mentions it, but we don't need it now. As a side, [event handlers can be passed with much less effort](https://svelte.dev/repl/00f1ceb442614501bfa1edeb87d38255?version=3.24.0) 
6. Add a new `name` string prop. This should be passed to the `for` attribute of `<label>` and the `name` + `id` attribute of `<input>`

### Markup
1. Remove `class={className}` from the top level div
2. The divs surrounding start and end adornment slots should not be rendered if slot content isn't passed. Right now its creating unused space on the Default story. See [checking for slot content](https://svelte.dev/tutorial/optional-slots).
3. The inputs focus should be `teal-500`, but I'll address this more in next comment...
4. The start and end adornment containers are not responsive their contents width. This means the "start" story is overlapping the input content. This means deviating from the TailwindUI styles slightly so that the adornement/input containing div is `flex`. We've figured this out in the Eisbuk TextInput component already, have a look at that and bring the styles over. A consequence of this is that the input's focus needs to be removed, and the containing div needs to use`focus-within`. This is part of styles you'll see in TextInput, but the focus should be updated to `teal-500`

### Stories
I would update the stories to reflect the Figma mocks:
1. Default
2. Default with placeholder
3. Start Adornment, with the mail icon and email placeholder
4. Start & End, with the dollar sign and USD 

We are going to have to add disabled and error props/styles too, but I can see this depending on the interface of whatever form state management library we use - Formik is only for React, so we'll need to investigate an alternative. I will make a note of this and write it up as an enhancement.

delasous (approved):
Touched up a few styles and (think) I fixed .svelte file prettying. I'd say its good to go now

PR #104, opened 2022-09-30
Feat/checkbox

delasous:
Add [mocked checkbox](https://www.figma.com/file/aKxt2b7lVOivTZ5jgrQwe2/Librocco?node-id=1575%3A29867) component (no ticket for this)


pkg/ui/src/FormFields/Checkbox/Checkbox.svelte
```diff
@@ -0,0 +1,41 @@
...
+		'rounded',
+		'disabled:text-gray-200'
+	].join(' ');
+</script>
+
+<div class="relative flex items-start">
+	<div class="flex items-center h-5">
+		<input
+			type="checkbox"
+			id={name}
```
silviot, line 26: The `id` of an HTML element needs to be unique in the page. Using the `name` could lead to problems.
Say we have a boolean field `available` in a book form. If we ever use two instances of the book form on the page, there will be two checkboxes with the same id. I think (I'm not sure) in that case all labels would drive the first checkbox.
To solve the same problem in Eisbuk we created a unique id on the fly at component creation, and use that in the `checkbox` `id` and the `label` `for` attributes.

delasous: Updated to use uuid for checkbox id

fadwamahmoud (approved):

PR #106, opened 2022-10-04
Feature/dropdown trigger

fadwamahmoud:
Dropdown trigger component.

I'm facing a weird issue with the svg package; adding any new svg component doesn't work(old ones work just fine). I even tried copying one of the old svg files to a different file under a different component name and still nothing. So I left the svg element inside the component for now.


delasous (commented):
## Dropdown
1. `Dropdown.svelte`:
    a. Should manage visibility of "items" slot via click of "trigger" slot: when the trigger toggles open state, the items should transition in/out. Look at [Transitions directive tutorial](https://svelte.dev/tutorial/transition).
 2. `DropdownItems.svelte`:
    a. `items` prop should be required, remove default `[' ']` and provide typing instead
    b. Each item is a link => should also be passed an `href` attribute and optionally an`on:click` handler
3. `DropdownTrigger.svelte`:
    a. `label` should be a required prop
    b. Remove Icon as prop and pass as named "icon" slot
    c. (Pending merge of Histoire/svg PR) - use equivelant svgs from `lucide-svelte`
    d. `aria-expanded` and `aria-haspopup` should be related to open/close state of trigger, related to comment above under `Dropdown.svelte`
    e. Focus ring colour should be `teal-500`

## SelectMenu
1. `SelectMenu.svelte`:
    a. Should manage a current value which should be the "label" of `SelectMenuTrigger`. Current should also be settable as an `initialValue`
    b. `trigger` slot component state should manage visibility of `items` component and transition in / out
    c. Remove `fixed top-16 w-72` from top level div
2. `SelectMenuTrigger.svelte`:
    a. `aria-expanded` and `aria-haspopup=listbox` should reflect state
    b. Add icons from `lucide-svelte`
    c. Focus colours should be `teal-500`
 3. `SelectMenuItems.svelte`:
     a. `items` should be required, add typing instead
     b. `selected` should reflect `current` being managed by `SelectMenu.svelte`, should be required.
    c. List item styles/focus should reflect [mock](https://www.figma.com/file/aKxt2b7lVOivTZ5jgrQwe2/Librocco?node-id=1554%3A31578)
    d. Add tick-svg from `lucid-svelte`
    e. Not sure what aria label `aria-activedescendant` is doing, check how it works. All items are being rendered with id `listbox-option-0`, update this with index so that there is only a single 0.

Also, assuming Histoire/SVG work is in: migrate `.stories` to Histoire `.story`s

delasous (approved):

PR #108, opened 2022-10-07
Badge Component

fadwamahmoud:


delasous (commented):
Looks good.

Two small updates:
1. `rounded-full` => `rounded`
2. BadgeColor.Primary => BadgeColor.Neutral

Before this goes in I would also love to have Histoire work on main, so we will need to update the storybook `.stories` file to `.story` and use e.g Histoire markup/features in `Button.story.svelte`

delasous (approved):

PR #110, opened 2022-10-11
Feature/svg lib

delasous:
Includes #109.

Replaces self-managed SVG package with [lucide](https://lucide.dev/) / [lucide-svelte](https://lucide.dev/docs/lucide-svelte).

Removes unused NoteButton component (ditched in last mock updates)


pkg/ui/package.json
```diff
@@ -44,13 +38,16 @@
...
+		"histoire": "~0.11.2",
+		"@histoire/plugin-svelte": "~0.11.2",
+		"@histoire/controls": "~0.11.3",
+		"autoprefixer": "~10.4.12"
 	},
 	"type": "module",
 	"dependencies": {
-		"uuid": "~9.0.0"
+		"uuid": "~9.0.0",
+		"lucide-svelte": "~0.91.0"
```
ikusteu, line 51: I would move `lucide-svelte` from `dependencies` and add it to both `peerDependencies` and `devDependencies` + externalise it when bundling with Vite. 

The reasoning is: the library shouldn't be packaged with the UI as it will still be loaded when packaging `web-client` (and we're using the UI package solely for `web-client` and not publishing it, so there's no need to bundle the svg's in)

pkg/ui/src/Button/Button.svelte
```diff
@@ -106,24 +96,18 @@
...
-	$: iconClasses = ['block', iconSizeClasses, iconSpacingClasses, iconColorClass].join(' ');
 </script>
 
 <svelte:element this={as} class={containerClasses} on:click>
-	<span class="flex items-center">
-		{#if showIcon && iconPosition === IconPosition.Start}
-			<span class={iconClasses}><svelte:component this={Icon} /></span>
+	<span class="flex items-center gap-x-2">
+		{#if $$slots.startAdornment && !ButtonShape.Circular}
+			<slot name="startAdornment" />
```
ikusteu, line 104: Loving this 👌 👌

ikusteu (approved):
Love it!!!

Some suggestions for enhancements:
- it would be nice to get some JSDoc parser: showing props in docs, storybook had it and I believe that might have been causing an issue with pnpm (some dependency of a dependency of that package 😆 )
- @delasous you've mentioned some upload of the built stories to a bucket...did I miss it, or is that an enhancement you're working on?

pkg/ui/package.json
delasous: Thanks for this! Never would have thought about it and its v. important.

PR #119, opened 2022-10-29
Feature/add prettier plugin tailwindcss

ikusteu:
- adds prettier plugin tailwindcss: 
  - I've been using this on my other project and I think it provides a lot of benefit
  - it has an opinionated way of sorting tailwind classes (in `class` attribute), as they say: "one less thing to fight with your team about"
  - it conflicts with svelte prettier plugin (if installed side by side), therefore, they've bundled the svelte plugin into it, so you automatically get both

- additionally, I've temporarily removed the `lint-staged` from `pre-commit` as it's flaky and downright broken at the moment (setup is still there, it only doesn't get ran on commit


ikusteu:
Merging as it includes some fixes. It kills `lint-staged` but it's broken anyhow. New ticket opened regarding the matter: #129

PR #125, opened 2022-11-01
InventoryTable component

fadwamahmoud:


pkg/ui/src/InventoryTable/InventoryTable.svelte
```diff
@@ -0,0 +1,48 @@
...
+		'ISBN',
+		'Title',
+		'Author',
+		'Editor',
+		'Quantity',
+		'Price',
+		'Year',
+		'OutOfPrint',
+		'EditedBy',
+		'Classification'
```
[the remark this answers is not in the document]
ikusteu: Why the `export let`?

pkg/ui/src/InventoryTable/InventoryTable.svelte
```diff
@@ -0,0 +1,48 @@
...
+								{#each headerRows as headerRow (headerRow)}
+									<th
+										scope="col"
+										class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6"
+									>
+										<!-- svelte-ignore a11y-invalid-attribute -->
+										<a href="#" class="group">
+											<div class="flex justify-center">
+												{headerRow}
+											</div>
```
ikusteu, line 33: Why the `<a>`?

ikusteu (changes requested):
Looks ok, but do apply the small changes. 
I don't particularly like the x-scroll...but we should put our heads together to figure out the appropriate solution for it...could you just hide the scroll bar?

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+	import type { TableData } from './types';
+
+	export let rowData: TableData;
+	export let className = '';
+</script>
+
+<tr class={className}>
+	{#each Object.values(rowData) as cell, index}
```
ikusteu, line 9: What happens if I, for instance omit the editor, but add quantity, would the quantity then be shown in place of editor?

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -0,0 +1,21 @@
...
+		<td
+			class={[
+				'whitespace-nowrap py-2 text-sm',
+				index ? 'px-2 text-gray-500' : 'pl-4 pr-3 font-medium text-gray-900 sm:pl-6'
+			].join(' ')}
+		>
+			<div>
+				{cell}
```
ikusteu, line 17: I like the way "N/A" is shown for editor, can we make it a default fallback here: if any field is not defined, it should render "N/A"

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+	import type { TableData } from './types';
+
+	export let rowData: TableData;
+	export let className = '';
```
ikusteu, line 5: Can we use `class` rather than `className` to keep with Svelte conventions

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+	import type { TableData } from './types';
+
+	export let rowData: TableData;
```
ikusteu, line 4: Can flatten this into named props (have `isbn`, `title`, etc...instead of them being properties on `rowData`) That's easier to provide fallbacks for and more efficient if only one field changes (we don't rerender the entire row, only the field)

ikusteu (changes requested):
Seems ok, but let's not use JS in place of CSS and make the styling consistent throughout (table row cells and table heading cells).

Plus, I feel `Object.values` is not really the way to go when rendering table cells.

pkg/ui/tailwind.config.cjs
```diff
@@ -3,5 +3,19 @@ const theme = require('../scaffold/tailwind.theme');
...
+		function ({ addUtilities }) {
+			addUtilities({
+				'.scrollbar-hide::-webkit-scrollbar': {
+					display: 'none'
+				},
+
+				'.scrollbar-hide': {
+					'-ms-overflow-style': 'none',
+					'scrollbar-width': 'none'
+				}
```
ikusteu, line 17: Could have used nesting (instead of having `.scrollbar-hide::--webkit-scrollbar` could have simply used `&::--webkit-scrollbar` inside of `.scrollbar-hide`)

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -1,21 +1,66 @@
...
-			<div>
-				{cell}
-			</div>
-		</td>
-	{/each}
+<tr class="even:bg-gray-50 whitespace-nowrap  text-sm  text-gray-500 ">
+	<td class="first:text-gray-900 first:font-medium py-2 pl-4">
+		<div>
+			{ISBN}
+		</div>
```
ikusteu, line 18: Is the `div` necessary here?

pkg/ui/src/InventoryTable/InventoryTable.svelte
```diff
@@ -24,17 +24,17 @@
 								{#each columnTitles as columnTitle (columnTitle)}
 									<th
 										scope="col"
-										class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6"
+										class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900"
 									>
-										<div class="flex justify-center">
+										<div>
 											{columnTitle}
 										</div>
```
ikusteu, line 31: Is the `div` necessary here

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -1,21 +1,66 @@
 <script lang="ts">
-	import type { TableData } from './types';
-
-	export let rowData: TableData;
-	export let className = '';
+	export let ISBN = 0;
```
ikusteu, line 2: can we name this `isbn` as per JS conventions

pkg/ui/tailwind.config.cjs
```diff
@@ -7,11 +7,10 @@ module.exports = {
...
 		function ({ addUtilities }) {
 			addUtilities({
-				'.scrollbar-hide::-webkit-scrollbar': {
-					display: 'none'
-				},
-
 				'.scrollbar-hide': {
+					'&::--webkit-scrollbar': {
+						display: 'none'
+					},
```
ikusteu, line 13: 👌

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -1,21 +1,66 @@
...
-				'whitespace-nowrap py-2 text-sm',
-				index ? 'px-2 text-gray-500' : 'pl-4 pr-3 font-medium text-gray-900 sm:pl-6'
-			].join(' ')}
-		>
-			<div>
-				{cell}
-			</div>
-		</td>
-	{/each}
+<tr class="even:bg-gray-50 whitespace-nowrap  text-sm  text-gray-500 ">
```
ikusteu, line 14: Can you remove excess whitespaces here

ikusteu (changes requested):
A couple of small changes + I would stick to named props, but would group them before rendering:
```typescript
cells: {
  isbn,
  title,
  // ...etc
}
```

and then render them using `{#each}`. I wouldn't apply 'N/A' as a fall back for string fields only, but use it while rendering, e.g.
```svelte
<script>
  export let title = ""
  // ...props

  let colls = {
    title,
    // ...etc
  }
</script>

{#each colls as coll}
  <td>{coll || "N/A"}</td>
{/each}
```

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -0,0 +1,34 @@
...
+	export let isbn;
+	export let title;
+	export let author;
+	export let editor;
+	export let quantity;
+	export let price;
+	export let year;
+	export let outOfPrint;
+	export let editedBy;
+	export let classification;
```
ikusteu, line 12: Could you, please initialise the values. That makes them optional. You can do `undefined` for each.

pkg/ui/src/InventoryTable/InventoryTableRow.svelte
```diff
@@ -0,0 +1,34 @@
...
+		outOfPrint,
+		editedBy,
+		classification
+	};
+</script>
+
+<tr class="even:bg-gray-50 whitespace-nowrap text-sm text-gray-500 ">
+	{#each Object.values(cols) as col}
+		<td class="first:text-gray-900 first:font-medium py-2 pl-4">
+			{col || 'N/A'}
```
ikusteu, line 31: Let's check if it's `undefined` and then use "N/A" as out of print = false will show "N/A" for instance

ikusteu: You can use `col ?? 'N/A'`

ikusteu (approved):

PR #127, opened 2022-11-03
Header component

### LINKED ISSUE #126: Component: Header
I think the initial idea was to use the header in `+layout.svelte`, but I would like for us to have a `Header` component represent the entire  https://www.figma.com/file/aKxt2b7lVOivTZ5jgrQwe2/Librocco?node-id=1552%3A29671 element.

In that case, we could reuse the nav part, but would be able to pass in the `title` prop and `actionButton` slot. That, however wouldn't be a part of `+layout.svelte`, but rather passed in as a slot in #67 

```svelte
<InventoryPage>
  <Header slot="header" title="Stock">
    <button slot="actionButton">New Warehouse</button>
  </Header>

  <!-- Rest of the page slots -->
</InventoryPage>
```

It would have the nav baked in:
- stock
- inbound
- outbound

It would accept a `title: string` prop and `<slot name="actionButton" />`

fadwamahmoud:
Fixes #126


pkg/ui/src/Header/HeaderTabContext.svelte
```diff
@@ -0,0 +1,35 @@
...
+
+	setContext<TabContext>(TABS, {
+		tabs,
+		currentTab,
+		registerTab,
+		selectTab
+	});
+</script>
+
+<slot />
```
ikusteu, line 35: I don't think context is necessary at all here...

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,93 @@
+<script lang="ts">
+	import { getContext, onMount } from 'svelte';
+
+	import { TABS, type TabContext } from './HeaderTabContext.svelte';
+	import HeaderTab from './HeaderTab.svelte';
+
+	export let title = 'Stock';
+	export let tabNames: string[];
+	export let initialTabIx = 0;
+	export let ariaLabel = 'Select a tab';
```
ikusteu, line 10: When I said 
> It would have the nav baked in
I meant something like this:
```typescript
// Notice the 'const', not 'let'
const links = [
  {
    label: "Stock",
    href: "/inventory/stock"
  },
  {
    label: "Inbound",
    href: "/inventory/inbound"
  },
  {
    label: "Outbound",
    href: "/inventory/outbound"
  }
]
```

ikusteu (changes requested):
Visually it looks ok, but implementationally it seems as if a lot of code is (hastily) copied over from somewhere else (namely classes like `sm:truncate` which are completely out of place here).

Could you please go over the code and reduce the unnecessary stuff. 

Additionally, the idea is to implement a navigation as discussed in #118

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
...
+		{
+			label: 'Inbound',
+			href: '/inventory/inbound'
+		},
+		{
+			label: 'Outbound',
+			href: '/inventory/outbound'
+		}
+	];
+	export let currentUrl = 'stock';
```
ikusteu, line 18: 1. I can see how this could go wrong:
- passing a current "url", which isn't a proper url string, from the outside
- matching a passed "url" to a label (which might get translated at some point)

2. "stock" isn't a proper url string, what you want to use is `$page.url.pathname` and match it against each `href` to get the current location (the same way you would use react router context)

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
...
+			<nav class="flex space-x-4">
+				{#each links as { label, href }}
+					<a
+						{href}
+						class={[
+							'px-3 py-2 text-sm rounded-md',
+							currentUrl === label.toLowerCase()
+								? 'bg-teal-500 text-gray-900'
+								: 'text-white hover:text-teal-500'
+						].join(' ')}
```
ikusteu, line 35: I know I've suggested this for react, but could we use `"px-3 py-2 text-sm rounder-md {/* current url check here */ ? "bg-teal-500 text-gray-900" : "text-white hover:text-teal-500"}"`. That way we would still get nice formatting, but also class sorting (to some extent). For full reasoning, check #121

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
...
+						aria-current={currentUrl === label.toLowerCase() ? 'page' : 'false'}
+					>
+						<span>{label}</span>
+					</a>
+				{/each}
+			</nav>
+		</div>
+	</div>
+	<div class="mt-2 md:flex md:items-center md:justify-between">
```
ikusteu, line 44: You forgot the responsive classes here as well

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
...
+		<div class="min-w-0 flex-1">
+			<h2 class="text-3xl font-bold leading-7 text-white">
+				{title}
+			</h2>
+		</div>
+		<div class="mt-4 pt-4 flex flex-shrink-0 md:mt-0 md:ml-4">
```
ikusteu, line 50: `md` here as well

ikusteu: `ml` is unnecessary since you have `justify-between` on the container

ikusteu: `flex-shrink` is unnecessary since the content wont touch the other content (title and action button) as they're on the opposite sides of the header

ikusteu: Instead of using `mt` here, can we set the fixed height on the row and have `items-center` (already set on the row element: parent to this element) be in charge of aligning everything vertically?

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
...
+						aria-current={currentUrl === label.toLowerCase() ? 'page' : 'false'}
+					>
+						<span>{label}</span>
+					</a>
+				{/each}
+			</nav>
+		</div>
+	</div>
+	<div class="mt-2 md:flex md:items-center md:justify-between">
+		<div class="min-w-0 flex-1">
```
ikusteu, line 45: Is the this div at all necessary in this case?? Can't we just use the `h2`?

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
...
+		}
+	];
+	export let currentUrl = 'stock';
+</script>
+
+<div class="bg-gray-900 p-10 pt-5 ">
+	<div class="border-b border-sky-900 pb-5 flex sm:items-center ">
+		<Logo />
+
+		<div class="pl-4">
```
ikusteu, line 25: I think setting the `Logo` in a `<div class="mr-4">` makes more sense that setting the following nav into additional container and applying `ml-4`, let alone `pl-4`. 

The flow always goes from top to bottom and from left to right, so it's the easiest to organise the layout (html/css) in that way.

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
...
+		},
+		{
+			label: 'Outbound',
+			href: '/inventory/outbound'
+		}
+	];
+	export let currentUrl = 'stock';
+</script>
+
+<div class="bg-gray-900 p-10 pt-5 ">
```
ikusteu, line 21: Instead of using `pt` here, can't we have:
```html
<!-- Main container -->
<header class="w-full px-[70px] bg-gray-900">
  
  <!-- First row -->
  <div class="h-16 border-b border-sky-900 ...other classes">
    <!-- Logo and nav -->
  </div>

  <!-- Second row -->
  <div class="h-[168px] ...other classes">
    <!-- Title and action button -->
  </div>

</header>
```
This way the outermost container is in charge of setting the background, the full width and the padding on the sides.

The children: first and second row can stretch for the full width (as the padding is already set by the parent element and are in charge of setting their own height, aligning their children elements and first row renders the border bottom

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,54 @@
+<script lang="ts">
+	import Logo from './Logo.svelte';
+	export let title = 'Stock';
+	const links = [
```
ikusteu, line 4: Don't be afraid to add whitespace here and there 😂

ikusteu (changes requested):

pkg/ui/src/Header/Header.svelte
fadwamahmoud: can I import page outside of a sveltekit page?

pkg/ui/src/Header/Header.svelte
fadwamahmoud: are you referring to making the title have a different hover style?

ikusteu: No...`md:` prefixes

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,52 @@
...
+					<span>{label}</span>
+				</a>
+			{/each}
+		</nav>
+	</div>
+	<div class="flex h-24 items-center justify-between">
+		<h2 class="text-3xl font-bold leading-7 text-white">
+			{title}
+		</h2>
+
```
ikusteu, line 47: Why `h2`...isn't this the top title (`h1`)

pkg/ui/src/Header/Header.svelte
```diff
@@ -0,0 +1,52 @@
...
+		<div class="flex">
+			<slot name="actionButton" />
+		</div>
```
ikusteu, line 50: Can we simply render the `slot` here without the `flex` container....it already is in one flex container (the second row), so I don't see the benefit of `<div class="flex" >` wrapping this

ikusteu (changes requested):
Quite nice, just a couple of very small fixes and we should be there.

ikusteu (approved):

PR #128, opened 2022-11-03
Feature/data stores

### LINKED ISSUE #122: Create dummy store(s) with some fixtures
We should create a store to serve as a middle layer (for data transformation) and initially provide the (hard coded) test data. This store would be used for two purposes:
1. Provide test data which the pages would hook into (so that we can have the visual preview when writing the UI as well as provide some basic functionality of page components)
2. The store would be a middle layer of transforming the data from the db for frontend usage:
  - frontend would have a certain interface to communicate with the middle layer (said store)
  - "backend" (pouchdb) would have a certain interface to supply the data to the store as well as actions to update the db
  - this way when (and if) only one or the other changes, only that side of the middle layer would change
  - we can easily switch out the test data with an actual pouchdb once the "backend" side of things is developed
_The store will not be just a single store, but rather a collection of stores or streams (standardised over time)._

## Data needed by the frontend:
- a list of warehouses
- a list of in notes grouped by warehouses
- a list of out notes
- stock for each warehouse
- full stock
- note data for each note

_As a first pass we only need some data...so full stock doesn't have to reflect the aggregate of all warehouses, nor does any warehouse stock have to reflect the actual state when you play all the notes. This should, ofc change in the future._

_As a first pass, there need not be any action from the frontend back to the store (such as updating the note, stock etc. Navigation is done in a flux way: frontend navigates to a different route, thus specifying a different view or a different warehouse/note to be shown and updating the `page` store. The store(s) in charge of passing the data back to the frontend listen to route (`page`) updates and change the data accordingly._

### Data stores
_these stores are kept internally, no need to export as derived stores are exported for frontend usage_

- **bookStore** - a store containing a map of books, keyed by isbn, a single entry looking something like this:
  - **[ isbn:** string **]**:  
    - **authors**: string array
    - **price**: number
    - **publisher** (I'm guessing this would be the `editore` field): string
    - **year**: string
    - **title**: string

#### Content stores
The following stores are used to read the entries for a table view and, therefore have similar interface. 
They all have an **entries** field, an array with each entry having the following structure:
- **isbn**: string
- **quantity**: number

In `stock` view, those **entries** represent a stock for a volume, while in `inbound`/`outbound` note view, each entry is a volume transaction in the note.

**warehouseStore** - a store containing a map of warehouses, keyed by warehouse name:
- **[ warehouse:** string **]**:
  - **entries**: see entries description ☝️ 
  - **inNotes** - an array of strings (ids of inbound notes belonging to the warehouse)
   
**inNoteStore** - a map of inbound notes, keyed by note id, each entry:
- **[ noteId:** string **]**:
  - **entries**: see entries description ☝️ 

**outNoteStore** - a map of outbound notes, keyed by note id, each entry:
- **[ noteId:** string **]**:
  - **entries**: see entries description ☝️ 

### Derived stores
_these store would be exported for frontend usage_

#### Navigation
Each of the following stores is used for side navigation items for a particular view: 
- **warehouses** (string array) - a list of all warehouse keys (names) - derived from **warehouseStore** (keys)
- **inNotes** (record) - a list of in note names grouped by warehouses - derived from **warehouseStore** 
  - **[warehouse]**: string array (see **warehouseStore.inNotes**)
- **outNotes** - string array of note ids - derived from **outNoteStore** directly (keys)

#### Derived content
There should be a derived store with table content, for a particular view. Each store should derive its state from the corresponding content store (**warehouseStore**, **inNoteStore**, **outNoteStore**), with **entries** picked out for an appropriate warehouse/note, the `warehouse`/`noteId` should be derived from svelte's `page` store (as described in #118).

A content store for each view should return the same content structure (an array of following entries):
- **isbn** (from current note/warehouse **entries** field)
- **title** (from **bookStore** entry, found by **isbn**)
- **authors** (from **bookStore** entry, found by **isbn**)
- **publisher** (from **bookStore** entry, found by **isbn**)
- **price** (from **bookStore** entry, found by **isbn**)
- **year** (from **bookStore** entry, found by **isbn**)
- **quantity** (from current note/warehouse **entries** field) - _this doesn't exist in the mockup_

## Enhancements (next steps)
- this schema will definitely evolve over time as we implement the frontend and backend
- switch out hard coded test data with the backend implementation, syncing with couchdb containing test data (to simulate production as much as possible)
- we will want to test this at some point
- upgrade the front facing interface with actions to update the notes and transactions commit/delete note, etc.

ikusteu:
Fixes #122 
With some simple tests


ikusteu:
Merging this as it's open for a while and I need it for a different PR. Even if something is wrong, it will get iterated over either way.

PR #130, opened 2022-11-07
Feature/inventory page

### LINKED ISSUE #67: Feature: PageTemplate
## Component: InventoryPage

_This would be a template for page styling, but I think it would make more sense to call is simply InventoryPage`

**Description:** This component would be used as presentational template for a page component. To clarify on @delasous comment about seeing how this fits into svelte's layout practices:
I'm **not** suggesting we make this a `+layout.svelte` type of component. This component would merely be a complex (stateless) style container. Example usage:
```svelte
<InventoryPage>

<svelte:fragment name="sidebar">
  <SideBarComponent(s) />
</svelte:fragment>

<svelte:fragment name="table-section">
  {#each tableRows as row}
    <TableRowComponent {...row} />
  {/each}
</svelte:fragment>

</InventoryPage>
```

There are certain points why I feel this makes sense:
- it would abstract away the positioning of main content containers on the page
- by abstracting away the container css logic, we avoid code redundancy per page basis
- when mobile version mockup is created, it would be easier to apply automatic repositioning of said containers

#### Elements (containers):

Thanks to Svelte's extremely convenient way of rendering elements passed as named slots (`<Container><slot name="section-1"/></Container>`), the implementation becomes trivial and requires minimum to no props (compared to a would-be React implementation)

- Header:
  - The entire `Header` component could get passed into this slot:
    - that way we can reuse the header component, but have the designated space where it would be rendered
- Sidebar:
  - shown on the leftmost side bar
  - should span the entire height of the page and scroll on `overflow-y`
  - `overflow-x` should be hidden (if an element overflows, we could show the ellipsis)
- Table heading: 
  - would show title, search bar, action buttons (depending on the page/view)
- Table section:
  - shows table rows for a view (stock, transactions)
  - even though number of table entries per page would be limited, we might want to enable overflow-y-scroll
- Table Footer:
  - for now it seems it will be the same across views (show number of items shown, number of total items and pagination)
  - even though it's the same across pages it still makes sense to pass it as a named slot (rather than abstracting it away in `Page` template to allow for the caller (the `+page.svelte`) direct control over the pagination functionality

### Possible enhancements:
- Small screen: this really depends on the future small screen mockup: the existing containers could be extended to simply rearrange on smaller screens, or extend the props and have sm version of each container at which point the content in large screen containers would be hidden and sm content shown instead

### LINKED ISSUE #123: Start implementation per page basis
I've described the idea for routing/folder structure in #118.
While still waiting for confirmation, I think we can start the implementation and, maybe, change the folder structure later if needed.

Without complicating stuff too much, I feel a lot of the atoms have been developed, so it might be good to start thinking about the frontend top-down: Start on implementing the pages using the existing atoms and hooking those pages to the middle layer, described in #122.

As we work on the pages we will see the would-be requirements for atoms or, perhaps, a way to group components into mote complex ones to avoid redundancy.

But...to start from the beginning, I suggest we implement each `+page.svelte` file as if it's the only one, meaning, we can afford redundancy (copy/paste the structure from another `+page.svelte` file) and maybe reduce the redundancy on second, or any following, pass.

## The pages we need to implement

- `/inventory/stock/[[warehouse]]/+page.svelte`
- `/inventory/inbound/[[note]]/+page.svelte`
- `/inventory/outbound/[[note]]/+page.svelte`

### Elements

- side nav (each page should implement its own version of side nav, according to mockup)
- table heading (as a first pass, each page could have only the Title (note title, warehouse)
- table section

### Functionality

- each page should hook to corresponding store(s)
- each page should be able to navigate using the side nav (and show appropriate content in the table section)

## Enhancements

- Full heading
- Editable title
- Pagination

ikusteu:
Fixes #67 
Fixes #123


delasous (approved):
Looks excellent: Store logic is clean and easy to follow, and InventoryPage layout does a solid job of minimizing redundancy. Really enjoyed playing with a live version.

Only issue I came across is `rushx build:prod` fails to produce a static bundle. I have fixed this in a branch called`experiment/InventoryPage-loading`, it:
1. Moves `hooks.server.ts` redirect logic to `routes/inventory/+layout.ts`;
2. Fixes prerender config by adding `prerender = true` to this +layout.ts file, and defining entries in `svelte.config.js` in place of `fallback: index.html`

There is more discussion to be had around this prerendering / page loading logic, but I will table it in our meetings, and we can prospectively fold it in to "App Implementation 3".

PR #139, opened 2022-12-11
Create a TextEditable component:

ikusteu:
* Create the component
* Apply quick styling
* Tests and implement the desired (bind, save, update) functionality

The component should be used to enable updates to note titles:
- when not editing it shows text and an on-hover edit icon
- on click it changes to input field
- the props are structured in such a way that binding to the `value` prop applies updates from the bound value into the component immediately, while updates from inside the component are propagated to the bound value only on save (as we'll probably use these updates to asynchronously update the db)


pkg/ui/src/TextEditable/TextEditable.svelte
```diff
@@ -0,0 +1,60 @@
...
+	 * - the user edits the input - no change is propagated to the parent component
+	 * - the user saves the input - the parent component is updated with the new value
+	 */
+	export let value = '';
+
+	/** This is the internal value, used to store the current state of the input */
+	$: text = value;
+	export let isEditing = false;
+
+	console.log('Is editing: ' + isEditing);
```
delasous, line 18: Delete

pkg/ui/src/TextEditable/TextEditable.svelte
```diff
@@ -0,0 +1,60 @@
...
+
+	/** Enter edit mode */
+	function edit() {
+		isEditing = true;
```
delasous, line 22: An alternative to `autofocus` would be:
1. Bind to an inputRef `bind:this={inputRef}`
2. After toggling `isEditing = true` , add `tick().then(() => (inputRef as HTMLElement).focus())`

pkg/ui/src/TextEditable/TextEditable.svelte
```diff
@@ -0,0 +1,60 @@
...
+	/** Save the input and propagate the new value to the parent component */
+	function save() {
+		if (value != text) {
+			value = text;
+		}
+		isEditing = false;
+	}
+</script>
+
+<div {...$$restProps}>
```
delasous, line 38: Not required, just FYI: In Edible's version I ended up spreading `$$restProps` on the input as this component was being used in a form and there were some additional one-off props that I had to pass through from the form state library

pkg/ui/src/TextEditable/TextEditable.svelte
```diff
@@ -0,0 +1,60 @@
...
+	{#if !isEditing}
+		<p on:click={edit} on:keydown class="group relative inline-block h-[38px] cursor-pointer pt-4 pr-8" on:focus={edit}>
+			<span class="inline-block align-middle text-lg font-medium leading-6">{text}</span>
+			<button class="absolute top-0 right-0 hidden h-10 w-10 -translate-y-1/4 p-2 group-hover:block" on:click={edit}>
+				<Pencil class="h-6 w-6 text-cyan-700" />
```
delasous, line 43: Not required, just my taste: can we shrink this pencil a little to `h|w-4`

delasous (approved):
LGTM

pkg/ui/src/TextEditable/TextEditable.svelte
ikusteu: 👍

pkg/ui/src/TextEditable/TextEditable.svelte
ikusteu: Wow...cool

pkg/ui/src/TextEditable/TextEditable.svelte
ikusteu: 👍

PR #152, opened 2023-02-21
Feature/use db v1

ikusteu:
This is a slightly bugged version of database integration into the main app...Most of the core functionality is there, unless it's chocked by max TCP connections. 

Most of the aforementioned issues should get solved when we have the local pouchdb syncing to the CouchDB (instead of having multiple connections open against CouchDB directly).

@silviot @delasous let me know if you think it can be merged as is, or if you think it would block the development of other features in this state.

@fadwamahmoud I'm writing up the ticket for the `db.init` function which should take care of most of the issues (as explained above), so then that PR could be opened either against `main` or against this branch (depending on the merge consensus)

**Edit** 
All of the aforementioned issues are solved and included in this branch. The creation of notes warehouses in not integrated as of yet, but I'm merging this in and will apply updates against master (as this is already a considerable diff)


apps/web-client/src/lib/stores/inventory/__tests__/display_name.test.ts
```diff
@@ -1,84 +1,141 @@
...
+		const note = await warehouse.note().create();
+
 		// Test for note
-		const inNoteStore = writable<NoteStore>({
-			'note-1': defaultNote
+		await note.setName('Note 1', {});
+
+		const ndn$ = createDisplayNameStore(note, null, {});
+		let noteDisplayName: string | undefined;
+		ndn$.subscribe((ndn) => (noteDisplayName = ndn));
```
delasous, line 29: Subscriptions to `ndn$ | wdn$ | pd$ | de$ ...` are never `unsubscribed`?

delasous (approved):

apps/web-client/src/lib/stores/inventory/__tests__/display_name.test.ts
ikusteu: Hmm...interesting point. I honestly don't know if this is a memory leak or if it gets garbage collected when all of the referenced vars are out of scope...

PR #154, opened 2023-02-22
Set up db.init() function

### LINKED ISSUE #153: Set up db.init() function
There are a couple of things we should do before the app starts (as the db is "loading"). I think it makes sense to put all of those in the `init()` function and have that function be called before the app becomes interactive. The features are:
- design document should be loaded
- the default warehouse should be created
- [a new one]: the db should be initalised as puchdb instance in indexed db, rather than an SDK for CouchDB communication

Tasks:
- Create an `init` method on the db interface:
  - [x] Update `db/src/types/index.ts` (types applicable to the database interface, regardless of the implementation) to have to include `db.init()` method ( should return a promise...whether or not that'll be void or return the db interface is up to you )
  - [x] Utilise `db.updateDesignDoc` to load the design documents into the db 
  - [x] Update the test runner to run (and await) `db.init()` before passing it (as a param) to each test task
  - [x] Create the default warehouse
  - **the method should not return (resolve) until all of the operations are done (both the operations above and below)**

- Set up replication
  - [x] Start the db as a db inside the browser
  - [x] Connect the db to the CouchDB instance (running on `localhost:5000`, we will update that later...I think @silviot also had a UI for that functionality in mind)
  - [x] Set up two way replication between local pouch and "remote" couch
  - [x] The init function shouldn't resolve until the initial data replication is complete (this might be a tricky one)

fadwamahmoud:
Fixes #153

### DIFF
--- a/pkg/datamodel-tests/src/implementations/version-1.1/db.ts
+++ b/pkg/datamodel-tests/src/implementations/version-1.1/db.ts
@@ -2,7 +2,7 @@ import { DbStream, DesignDocument, DocType, InNoteList, NavListEntry, utils } fr
 import { debug } from '@librocco/shared';
 
 import { DatabaseInterface, WarehouseInterface } from './types';
-
+import designDocs from './designDocuments';
 import { newWarehouse } from './warehouse';
 
 const { newViewStream } = utils;
@@ -16,6 +16,25 @@ class Database implements DatabaseInterface {
 		// Initialize the default warehouse (this makes sure the "0-all" warehouse exists, otherwise it will be created)
 		// All of this is done automatically when running db.warehouse('0-all')
 		this.warehouse('0-all').create();
+	}
+	async init(remote?: { database: PouchDB.Database }) {
+		//  Utilise db.updateDesignDoc to load the design documents into the db
+
+		let promises = [];
+
+		// Upload design documents if any
+		const ddUpdates = designDocs.map((dd) => this.updateDesignDoc(dd));
+		if (ddUpdates?.length) {
+			promises.push(ddUpdates);
+		}
+
+		// create default warehouse
+		const whPromise = this.warehouse().create();
+		promises.push(whPromise);
+
+		// Set up replication between local pouch and "remote" couch
+		remote && promises.push(replicate({ local: this._pouch, remote: remote.database }));
+		return Promise.all(promises);
 	}
 
 	warehouse(id?: string): WarehouseInterface {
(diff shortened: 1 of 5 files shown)


pkg/datamodel-tests/src/implementations/version-1.1/db.ts
```diff
@@ -17,6 +17,25 @@ class Database implements DatabaseInterface {
 		// All of this is done automatically when running db.warehouse('0-all')
 		this.warehouse('0-all').create();
 	}
+	async init(remote?: { database: PouchDB.Database }) {
```
ikusteu, line 20: I know this makes sense to you when writing a function, but think about it also making sense when calling it: Imagine you're calling this `db.init({ database })` -> it's kinda confusing...rename it so that it communicates that it's remote db you're passing in.

pkg/datamodel-tests/src/implementations/version-1.1/db.ts
```diff
@@ -92,4 +111,26 @@ export const newDatabase = (db: PouchDB.Database): Database => {
...
+			.to(database.remote)
+			.on('complete', function () {
+				// yay, we're done!
+				console.log('Replication complete');
+				resolve();
+			})
+			.on('error', function (err) {
+				// boo, something went wrong!
+				console.log('could not replicate to remote db', err);
+				reject();
```
ikusteu, line 132: Will this also replicate from the remote? (when I add a document in the remote db, I would like to see the update locally)

pkg/datamodel-tests/src/implementations/version-1.1/db.ts
ikusteu: You can just name it `params` and have a property `remoteDb` (or just `remote`, like you have in the helper function)

pkg/db/src/db.ts
```diff
@@ -20,6 +24,20 @@ class Database implements DatabaseInterface {
 		this.warehouse('0-all').create();
 	}
 
+	async init(remote?: { database: PouchDB.Database }) {
+		//  Utilise db.updateDesignDoc to load the design documents into the db
+		const designdDocsPromise = designDocs.forEach((designDoc) => {
```
ikusteu, line 29: *designDocsPromise

pkg/datamodel-tests/src/implementations/version-1.1/db.ts
ikusteu: This is a bit of a redundancy (you can simply import from `utils/pouchdb`

ikusteu: Is this replication live...does it keep replicating as either of the dbs changes?

ikusteu (changes requested):
Left some comments, additionally:
- initDB (in test runner) should initialise the db with only `dbName` and options `{ adapter: "memory" }`
- when calling `db.init` (in task setup, test runner), it should check and pass couch instance to replicate to (if the `__withDocker__` (`test:full`) is true (so kinda need to move the logic from previous point ☝️ here

ikusteu (approved):

PR #162, opened 2023-02-28
Display row stream

fadwamahmoud:


apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +25,41 @@ describe('tableContentStore', () => {
 			totalItems: 0
 		} as PaginationData);
 	});
+	test('should create display store from db and entries streams', async () => {
+		const db = await newTestDB();
+		await db.upsertBook({
+			isbn: '0007149522',
+			title: 'The Age of Wonder',
+			authors: ['Richard Holmes'],
```
ikusteu, line 33: This should be a (comma separated) string

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +25,41 @@ describe('tableContentStore', () => {
...
+			publisher: 'HarperCollins UK',
+			year: '2008',
+			price: 69.99
+		});
+		const note = await db.warehouse().note('note-1').create();
+		const books = await db.getBooks(['0007149522']);
+		await note.update({
```
ikusteu, line 40: `note.update` doesn't exist (I obviously have to fix type system...). What you want to do is do `note.addVolumes` (and pass an array of entries)

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +25,41 @@ describe('tableContentStore', () => {
...
+					warehouse: 'jazz',
+					quantity: 5
+				}
+			]
+		});
+
+		const de$ = createDisplayEntriesStore(note, readable(0), {});
+		let displayEntries: FullTableRow[];
+		de$.subscribe((de) => (displayEntries = de));
+		expect(displayEntries).toEqual([]);
```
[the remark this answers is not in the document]
ikusteu: Also, for actual test, you'll want to do `await waitFor(() => { ...assertions})`

apps/web-client/src/lib/stores/inventory/index.ts
```diff
@@ -16,7 +14,7 @@ interface NoteDisplayStores {
 	displayName: Writable<string | undefined>;
 	state: Writable<NoteAppState>;
 	updatedAt: Readable<Date | null>;
-	entries: Readable<DisplayRow[]>;
+	entries: Readable<FullTableRow[]>;
```
ikusteu, line 17: Let's update the `DisplayRow` to include the current type for the `BookEntry`

ikusteu: i.e. `DisplayRow` should be the type for the full table row

apps/web-client/src/lib/stores/inventory/table_content.ts
```diff
@@ -3,43 +3,81 @@ import { debug } from '@librocco/shared';
...
+	): Readable<FullTableRow[]>;
 }
+
+interface createDisplayRowStream {
+	(entity: NoteInterface<object> | WarehouseInterface<NoteInterface<object>, object>, ctx: DebugCtx): Observable<
+		FullTableRow[]
+	>;
+}
+
+const createDisplayRowStream: createDisplayRowStream = (entity, ctx) => {
```
ikusteu, line 27: Types should be PascalCased

apps/web-client/src/lib/stores/inventory/table_content.ts
```diff
@@ -3,43 +3,81 @@ import { debug } from '@librocco/shared';
...
+	const fullTableRow = entity?.stream(ctx).entries.pipe(
+		switchMap((valueFromEntryStream) => {
+			const isbns = valueFromEntryStream.map((entry) => entry.isbn);
+			// map entry to just isbns
```
ikusteu, line 31: I gave you leads on why you did't receive anything (in the test) from `entity?.stream(ctx).entries` (note wasn't updated properly). For further checks, you can add `.tap` above the `switchMap` here

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
ikusteu: On a semi-related note: you can always check that the note was updated properly by doing `console.log(note)` (thanks to our carefully created class, only the note fields will get logged to the console, without the methods and private values...should make for easier debugging)

apps/web-client/src/lib/stores/inventory/table_content.ts
```diff
@@ -3,43 +3,81 @@ import { debug } from '@librocco/shared';
...
+
+				map((booksFromDb) => {
+					// for each value from entry stream (volume stock client)
+					return valueFromEntryStream.map((individualEntry) => {
+						// find the corresponding book in booksFromDb
+						// instead of find => look in the same index then look at first element (maybe book is not in yet???)
+						const bookData = booksFromDb.find((book) => book.isbn === individualEntry.isbn);
+						// account for books not found
+						if (!bookData) return;
+						// return bookData + entry value
```
ikusteu, line 43: If no book data found, we should return the (`VolumeStockClient`) entry (without book data)

apps/web-client/src/lib/stores/inventory/table_content.ts
```diff
@@ -1,45 +1,81 @@
...
  * Creates a store that streams the entries to be displayed in the table, with respect to the content in the db and the current page (set by pagination element).
  * @param entity a note or warehouse interface
  * @param currentPageStore a store that containing the current page index
  * @param bookStore a store with book data - used to extend each entry from the entity store with the rest of the book data
  * @returns
  */
-export const createDisplayEntriesStore: CreateDisplayEntriesStore = (entity, currentPageStore, bookStore, ctx) => {
-	const entriesStore = readableFromStream(entity?.stream(ctx).entries, [], ctx);
+export const createDisplayEntriesStore: CreateDisplayEntriesStore = (db, entity, currentPageStore, ctx) => {
+	const fullTableRowStream = createDisplayRowStream(db, entity, ctx);
```
ikusteu, line 64: Make sure to update the JSDoc

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +29,139 @@ describe('tableContentStore', () => {
 			totalItems: 0
 		} as PaginationData);
 	});
+	test('should create display store from entries only if some books were not found', async () => {
+		const db = await newTestDB();
```
ikusteu, line 33: A whitespace between tests PLEASE 🙏

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +29,139 @@ describe('tableContentStore', () => {
...
+				title: 'Twelve Bar Blues',
+				authors: 'Patrick Neate',
+				publisher: 'Penguin UK',
+				year: '2002',
+				price: 39.86
+			}
+		];
+
+		// toMatchObject because of _rev
+		await waitFor(() => expect(displayEntries).toMatchObject(res));
```
ikusteu, line 163: This (toMatchObject) will become obsolete when you return only the book data (no couch document stuff)...so you'd want to assert they're equal (not matched)

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +29,139 @@ describe('tableContentStore', () => {
...
+			['019976915X', 10, `v1/jazz`],
+			['0195071409', 5, `v1/jazz`]
+		);
+
+		const de$ = createDisplayEntriesStore(db, note, readable(0), {});
+		let displayEntries: DisplayRow[];
+		de$.subscribe((de) => (displayEntries = de));
+		const res = [
+			{
+				isbn: '0195071409',
```
ikusteu, line 127: Not really a `res`...I've seen people naming this as `want`...

ikusteu: Or `expect`...but let's not clash with vitest API

ikusteu (changes requested):
Wrote a couple of comments to fix. Ping me if you've forgotten the suggestion for matching of the entries with book data

pkg/db/src/implementations/version-1.1/db.ts
```diff
@@ -53,6 +53,36 @@ class Database implements DatabaseInterface {
...
+
+	async getBooks(isbns: string[]): Promise<CouchDocument<BookEntry>[]> {
+		const rawBooks = await this._pouch.allDocs({ keys: isbns, include_docs: true });
+		// The rows are returned in the same order as the supplied keys array.
+		// The row for a nonexistent document will just contain an "error" property with the value "not_found".
+
+		const bookDocs = rawBooks.rows.reduce((values: CouchDocument<BookEntry>[], value) => {
+			if (value.doc) values.push(value.doc as CouchDocument<BookEntry>);
+			return values;
+		}, []);
```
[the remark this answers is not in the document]
fadwamahmoud: what about books that don't exist

pkg/db/src/implementations/version-1.1/db.ts
```diff
@@ -53,6 +53,36 @@ class Database implements DatabaseInterface {
...
+
+	async upsertBook(bookEntry: BookEntry): Promise<void> {
+		try {
+			const bookDocument = await this._pouch.get(bookEntry.isbn);
+			await this._pouch.put({ ...bookDocument, ...bookEntry });
+			return;
+		} catch (err) {
+			console.log('document not found');
+		}
+		await this._pouch.put({ ...bookEntry, _id: bookEntry.isbn });
```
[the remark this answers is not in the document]
fadwamahmoud: err here is of type unknown so I can't check for err.status etc
Should I cast it?
```
type SerializedError = {
	"docId":string;
	"error": boolean;
	"reason": string;
	"status":number;
  }
```

Then
```

catch (err) {
			
			if ((err as SerializedError).status !== 404) throw Error((err as SerializedError).reason)
			// not sure what to do if it's 404 so I threw an Error
			throw Error()

		}

```

pkg/db/src/implementations/version-1.1/db.ts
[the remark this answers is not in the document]
ikusteu: What about them?

pkg/db/src/implementations/version-1.1/db.ts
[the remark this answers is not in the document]
ikusteu: Nooo…you can just do ‘as any’ (it shouldn’t be used, but here we can get away with it)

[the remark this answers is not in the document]
ikusteu: It would be good if pouchdb has the type…then definitely do use the specific type, bit no need to write your own here

pkg/db/src/implementations/version-1.1/db.ts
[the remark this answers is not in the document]
ikusteu: I’m thinking…if we can guarantee that the request will return the same number of results as there are ids requested, with the ones not found not having doc, we can return doc…

pkg/db/src/implementations/version-1.1/db.ts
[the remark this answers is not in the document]
ikusteu: Aaand `throw new Error(err as SerialisedError)` is unnecessary:
1. I don't think the error type would be inferred for outer catch step (so type casting is unnecessary)
2. It's already an error, so no need for `new Error` constructor, you can just `throw err`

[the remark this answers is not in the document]
ikusteu: About 404, that means not-found = the document (book) we've requested is not found, return `undefined` (from `db.getBook()`, as you've specified in type signature)

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -1,15 +1,20 @@
...
 import { readable } from 'svelte/store';
+import { newTestDB } from '$lib/__testUtils__/db';
+const { waitFor } = testUtils;
 
 describe('tableContentStore', () => {
 	test("should not explode if no 'entity' is provided", async () => {
-		const de$ = createDisplayEntriesStore(undefined, readable(0), readable({}), {});
+		const db = await newTestDB();
+
+		const de$ = createDisplayEntriesStore(db, undefined, readable(0), {});
```
ikusteu, line 17: This should also work if `db` is undefined (should have the same behaviour as if the entiry is not defined)

ikusteu: I would prefer to test for both cases:
1. If both are undefined
2. If db is defined, but not the entity
 _(we won't have a case where db is undefined, and entity is)_

ikusteu: This could be a single `test()` case ☝️

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +29,187 @@ describe('tableContentStore', () => {
...
+
+		await db.upsertBook({
+			isbn: '0195399706',
+			title: 'The Age of Wonder',
+			authors: 'Richard Holmes',
+			publisher: 'HarperCollins UK',
+			year: '2008',
+			price: 69.99
+		});
+
```
ikusteu, line 44: LGTM, can we just:
- rename this to: "should stream book data with the entires (matching their isbn)"
- reduce the number of books to two (it's a lot of code, and we don't really need three entries)
- save `book1` and `book2` as variables and expect `{ ...entry, ...book }` (to reduce the code further)
- also use `Promise.all()` for setup (slightly quicker tests)

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +29,187 @@ describe('tableContentStore', () => {
...
+			}
+		];
+
+		await waitFor(() => expect(displayEntries).toEqual(want));
+	});
+
+	test('should create display store only from entries streams if corresponding books do not exist', async () => {
+		const db = await newTestDB();
+
+		await db.upsertBook({
```
ikusteu, line 116: - "should stream just the volume stock if for isbns with no book data found"
- also reduce to two: one book and one undefined
- also use variables so "compress" the test code
- also use `Promise.all` for setup

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -24,4 +29,187 @@ describe('tableContentStore', () => {
...
+			}
+		];
+
+		await waitFor(() => expect(displayEntries).toEqual(want));
+	});
+
+	test('should return book if found and undefined otherwise', async () => {
+		const db = await newTestDB();
+
+		const book1 = {
```
ikusteu, line 182: Let's remove this test: it should be in `db` package, but let's just remove it for now (I'll write it up for enhancement ticket)

apps/web-client/src/lib/types/inventory.ts
```diff
@@ -2,6 +2,7 @@ import type { NoteTempState } from '$lib/enums/inventory';
 import type { NoteState } from '$lib/enums/db';
 
 import type { VolumeQuantity } from './db';
+import type { VolumeStockClient, CouchDocument, BookEntry as BookEntryFromDB } from '@librocco/db';
```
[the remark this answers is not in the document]
ikusteu: Not resolved ☝️

apps/web-client/src/lib/stores/inventory/table_content.ts
```diff
@@ -1,45 +1,85 @@
...
+			);
+		})
+	);
+
+	return fullTableRow;
+};
+
 /**
  * Creates a store that streams the entries to be displayed in the table, with respect to the content in the db and the current page (set by pagination element).
+ * @param db current context database
```
ikusteu, line 61: Just database interface (let's not create a confusion with "context")

apps/web-client/src/lib/stores/inventory/table_content.ts
```diff
@@ -1,45 +1,85 @@
...
  * @param currentPageStore a store that containing the current page index
- * @param bookStore a store with book data - used to extend each entry from the entity store with the rest of the book data
+ * @param ctx debug context
  * @returns
  */
-export const createDisplayEntriesStore: CreateDisplayEntriesStore = (entity, currentPageStore, bookStore, ctx) => {
-	const entriesStore = readableFromStream(entity?.stream(ctx).entries, [], ctx);
+export const createDisplayEntriesStore: CreateDisplayEntriesStore = (db, entity, currentPageStore, ctx) => {
+	const fullTableRowStream = createDisplayRowStream(db, entity, ctx);
+	const displayRowStore = readableFromStream(fullTableRowStream, [], ctx);
```
ikusteu, line 69: Let's call this `entriesStore` like before, because here you're:
1. Creating `displayRows`
2. Creating `displayEntries` from `displayRows`...

The former aren't displayed (they're used to create the displayed ones) and, if any should be `displayRows` (implying they're displayed in the UI as table rows), those should be the latter.

apps/web-client/src/lib/stores/inventory/table_content.ts
```diff
@@ -1,45 +1,85 @@
...
-			}));
-			debug.log(ctx, 'display_entries:derived:res')(res);
-			return res;
-		}
-	);
+		const res = $displayRowStore.slice(start, end).map((entry) => ({
+			...entry
+		}));
+		debug.log(ctx, 'display_entries:derived:res')(res);
+		return res;
```
ikusteu, line 81: Can just do `return $displayRowStore.slice(start, end)` (unnecessary map, also, no need for additional variable here: it's clear what we're returning)

pkg/db/src/implementations/version-1.1/db.ts
```diff
@@ -53,6 +53,45 @@ class Database implements DatabaseInterface {
...
+
+	async upsertBook(bookEntry: BookEntry): Promise<void> {
+		let bookDoc: BookEntry | undefined = undefined;
+		try {
+			bookDoc = await this._pouch.get(bookEntry.isbn);
+		} catch (err) {
+			if ((err as any).status !== 404) throw err;
+			this._pouch.put({ ...bookEntry, _id: bookEntry.isbn });
+		} finally {
+			if (bookDoc) this._pouch.put({ ...bookDoc, ...bookEntry });
```
ikusteu, line 91: What happens if `bookDoc` is `undefined` ?

ikusteu (changes requested):
Minor fixes, other that that, LGTM

apps/web-client/src/lib/types/inventory.ts
```diff
@@ -21,8 +22,8 @@ export interface BookStore {
 	[isbn: string]: BookEntry;
 }
 
-/** The properties of a book + quantity row shown in the note/warehouse table. */
-export type DisplayRow = BookEntry & { quantity: number };
+/** The result of merging an entry with the equivalent book data */
+export type DisplayRow = VolumeStockClient & Partial<CouchDocument<BookEntryFromDB>>;
```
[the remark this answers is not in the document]
ikusteu: Still not resolved

PR #170, opened 2023-03-09
Readme updates

ikusteu:


delasous (approved):

apps/web-client/README.md
```diff
@@ -1,38 +1,282 @@
...
 
-```bash
-npm run build
+const paginationData = createPaginationDataStore(warehouse, currentPage, {});
 ```
 
-You can preview the production build with `npm run preview`.
+##### Context
+
+One final thing to mention. So far we've been showing usage of factory functions with the last parameter being and empty object (`{}`), this is a placeholder for the debug context (and an empty object is a valid parameter if we don't want to use it).
```
fadwamahmoud, line 265: being and empty object => an

apps/web-client/README.md
```diff
@@ -1,38 +1,282 @@
...
+-   if we're committing the note, we set the state as "Committed", the temp state ("Committing") is displayed, the update is propagated to the db (running `note.commit()`) and committed note streams the update, so the state "Committed" is displayed
+-   if deleting the note, the temp state is set to "Deleting", whilst the db update is performing `note.delete()`
+
+To achieve this behaviour, the note `state` store is composed of two parts, `internalState` store and the synthetic, `displayStateStore` store:
+
+-   `internalState` store:
+    -   an actual Svelte, writable, store
+    -   on subscribe, it opens subscription to the note's (in db) state
+    -   each time the note in db updates, the update is streamed to update the internal state store
+    -   additionally, it it intentionally a writable store, allowing for explicit updates, for UI only states, as we'll explore below
```
fadwamahmoud, line 128: it it => it is

apps/web-client/README.md
```diff
@@ -1,38 +1,282 @@
...
+
+| UI state  | UI temp state |
+| --------- | ------------- |
+| Draft     | Saving        |
+| Committed | Committing    |
+| Deleted   | Deleting      |
+
+The way temp states work:
+
+-   if we're updating the draft note (this is normally an outside update), the state displayed is "Saving", then when the note gets updated, the note's current state in db (Draft) is streamed back to the display store
```
fadwamahmoud, line 118: Is this meant to be state in db (committed : false)? since "draft" is not a db state

pkg/db/src/test-runner/README.md
```diff
@@ -0,0 +1,303 @@
...
+Before going to actual usage, let's quickly go over the data loaders for test data (fixtures).
+
+The test data loader is provided to the `runner` on initialisation: when the runner is initialised, it will load the test data (using the provided loader), transform the data to the right format (on `setup` step) and make that data avilable to the tests (through `getNotesAndWarehouses` passed to each test).
+
+_Note: the transform step is unnecessary and will be omitted in the future when we rewrite our test data to the right format._
+
+The loader must be an object with two methods:
+
+-   getNotes - should return a list of notes in their chronological order
+-   getSnaps - should return a list of snapshots of all of the warehouses, after each note has has been committed
```
fadwamahmoud, line 181: after each note has has => has

apps/web-client/README.md
```diff
@@ -1,38 +1,282 @@
...
+-   `/stock`:
+
+    -   **default view**
+    -   the load function instantiates the specific warehouse (read from route) and passes the warehouse interface down to the page component
+    -   the view is used to inspect the current stock in a particular warehouse
+    -   if warehouse not provided, or doesn't exist, defaults to: `/invntory/stock/0-all` (the default warehouse)
+
+-   `/inbound`:
+
+    -   the load function runs `db.findNote`, returning the note and its parent warehouse, passing the both down to the page component (as note interface and warehouse interface respectively)
```
fadwamahmoud, line 45: passing them both

pkg/shared/README.md
```diff
@@ -0,0 +1,281 @@
...
+	tap((val) => console.log('source_stream:mapped_value: ', val))
+);
+
+// Destination
+source.pipe(tap((val) => console.log('subscriber:received_value: ', val))).subscribe(() => {
+	/* Do something */
+});
+```
+
+In this case the the logging is trivial and we can track the logs of everything streamed from the source, through the pipeline, all the way to the subscriber. There is a simple problem though: If we wanted to "turn on" the logging, we'd have to manually add each `.tap` and `console.log` each time.
```
fadwamahmoud, line 36: the the logging

apps/web-client/README.md
ikusteu: Yes...here I meant the state from the db that gets streamed back (not the exact implementation). This is not an error, but I'll revisit if it should be explained better.

PR #183, opened 2023-03-10
Create book interface

fadwamahmoud:


pkg/db/src/implementations/version-1.1/db.ts
```diff
@@ -66,44 +67,9 @@ class Database implements DatabaseInterface {
...
-		} catch (err) {
-			if ((err as any).status !== 409) throw err;
-
-			const bookDoc = await this._pouch.get(bookEntry.isbn);
-
-			await this._pouch.put({ ...bookDoc, ...bookEntry });
-		}
+	books() {
+		const bookInterface = newBookInterface(this);
+		return bookInterface;
```
ikusteu, line 72: Why not simply `return newBooksInterface(this)`

pkg/db/src/types.ts
```diff
@@ -195,11 +195,17 @@ export interface DatabaseInterface<W extends WarehouseInterface = WarehouseInter
...
-	getBooks: (isbns: string[]) => Promise<(BookEntry | undefined)[]>;
-	getBook: (isbn: string) => Promise<BookEntry | undefined>;
-	upsertBook: (bookEntry: BookEntry) => Promise<void>;
+	books: () => BookInterface;
 }
 
+/**
+ * An interface for books in a db
+ */
+export interface BookInterface {
```
ikusteu, line 204: BooksInterface

ikusteu (changes requested):

pkg/db/src/__tests__/tests.ts
```diff
@@ -514,3 +514,57 @@ export const streamsShouldFallBackToDefaultValueForTheirType: TestFunction = asy
...
+		year: '2002',
+		price: 39.86
+	};
+
+	const bookInterface = db.books();
+
+	let bookEntries: (BookEntry | undefined)[] = [];
+
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	const [_, __, ___] = await Promise.all([
```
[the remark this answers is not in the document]
fadwamahmoud: Yeah, my initial intention was to write a test for the upsert as well as stream. I fixed it so that I first check for original book state then after it's updated.

ikusteu (changes requested):
A couple of small wrinkles, otherwise it looks good. (haven't tried it out though)

PR #188, opened 2023-03-16
Enhancement/add volumes

### LINKED ISSUE #171: Update 'note.addVolumes'
# `note.addVolumes` updates:

## Parameters

Currently we have a bit of a discrepancy between `note.addVolumes` and `note.updateTransaction`:

- `note.addVolumes` accepts `VolumeTransactionTuple` or an array of `VolumeTransactionTuple`s, like so:
```typescript
// Single transaction
note.addVolumes("12345678", 2, "warehouse-1")

// Multiple transactions
note.addVolumes(["12345678", 2, "warehouse-1"], ["11111111", 5, "warehouse-1"])
```

- `note.updateTransaction` accepts a `VolumeStock`, like so:
```typescript
note.updateTransaction({ isbn: "12345678", warehouseId: "warehouse-1", quantity: 2 })
```

We should fix this so that `note.addVolumes` accepts the same type (`VolumeStockQuantity`). We can leave the ...rest functionality, so the behaviour should be something like this:
```typescript
// Single transaction
note.addVolumes({ isbn: "12345678", warehouseId: "warehouse-1", quantity: 2 })

// Multiple transactions
note.addVolumes(
  { isbn: "12345678", warehouseId: "warehouse-1", quantity: 2 },
  { isbn: "11111111", warehouseId: "warehouse-1", quantity: 5 },
  // ...and so on
)
```

## Warehouse id versioning

When passing a warehouse id to `note.addVolumes`, we should be able to pass just the warehouse id, without the version number and the method should automatically prepend it:
```typescript
// Instead of doing 
note.addVolumes({ isbn: "12345678", quantity: 2, warehouseId: "v1/warehouse-1" })

// We should be able to simply do
note.addVolumes({ isbn: "12345678", quantity: 2, warehouseId: "warehouse-1" }) // and have 'v1/' be automatically prepended
```

The best (and easy) way this could be tested is by simply removing `versionId()` call on warehouse ids, when calling `note.addVolumes` in the (existing) tests

Tasks:

- [x] Delete the `VolumeTransactionTuple` type (as it won't be used anymore)
- [x] Update the `note.addVolumes` to accept `VolumeStock`
- [x] Update the `note.addVolumes` to automatically prepend the warehouse id with version string
- [x] Update the tests to only use `versionId` when making assertions (not when running `note.addVolumes`

fadwamahmoud:
Fixes #171

### DIFF
--- a/pkg/db/src/implementations/version-1.1/note.ts
+++ b/pkg/db/src/implementations/version-1.1/note.ts
@@ -217,22 +217,24 @@ class Note implements NoteInterface {
 	 */
 	addVolumes(...params: Parameters<NoteInterface['addVolumes']>): Promise<NoteInterface> {
 		return runAfterCondition(async () => {
-			const updateQuantity = (isbn: string, quantity: number, warehouseId: `v${number}/${string}`) => {
-				const matchIndex = this.entries.findIndex((entry) => entry.isbn === isbn && entry.warehouseId === warehouseId);
+			params.forEach((update) => {
+				const warehouseId = versionId(update.warehouseId || '');
+
+				const matchIndex = this.entries.findIndex(
+					(entry) => entry.isbn === update.isbn && entry.warehouseId === update.warehouseId
+				);
 
 				if (matchIndex === -1) {
-					this.entries.push({ isbn, warehouseId, quantity });
+					this.entries.push({ isbn: update.isbn, warehouseId, quantity: update.quantity });
 					return;
 				}
 
 				this.entries[matchIndex] = {
-					isbn,
+					isbn: update.isbn,
 					warehouseId,
-					quantity: this.entries[matchIndex].quantity + quantity
+					quantity: this.entries[matchIndex].quantity + update.quantity
 				};
-			};
-
-			params.forEach((update) => updateQuantity(update.isbn, update.quantity, versionId(update.warehouseId || this.#w._id)));
+			});
 
 			return this.update(this, {});
 		}, this.#initialized);
@@ -242,13 +244,19 @@ class Note implements NoteInterface {
 		// Create a safe copy of volume entries
 		const entries = [...this.entries];
 		// handle edge case where we have multiple books with the same isbn, but belonging to different warehouses
-		const i = entries.findIndex(({ isbn, warehouseId }) => isbn === transaction.isbn && warehouseId === transaction.warehouseId);
+		// match isbn only in case entry has no whId but transaction does (not undefined or empty string)
+		const i = entries.findIndex(
+			({ isbn, warehouseId }) =>
+				isbn === transaction.isbn &&
+				(warehouseId === versionId(transaction.warehouseId || '') || (warehouseId === versionId('') && transaction.warehouseId))
+		);
 
+		const versionedTransaction = { ...transaction, warehouseId: versionId(transaction.warehouseId || '') };
 		// If the entry exists, update it, if not push it to the end of the list.
 		if (i !== -1) {
-			entries[i] = transaction;
+			entries[i] = versionedTransaction;
 		} else {
-			entries.push(transaction);
+			entries.push(versionedTransaction);
 		}
 		// Post an update, the local entries will be updated by the update function.
 		return this.update({ entries }, {});
@@ -259,6 +267,9 @@ class Note implements NoteInterface {
 	 * when calculating the stock of the warehouse.
 	 */
 	commit(ctx: debug.DebugCtx): Promise<NoteInterface> {
+		if (this.entries.findIndex((entry) => entry.warehouseId === versionId('')) !== -1)
+			throw new Error('Cannot commit a note with empty warehouseId entries');
+
 		debug.log(ctx, 'note:commit')({});
 		return this.update({ committed: true }, ctx);
 	}
@@ -294,7 +305,7 @@ class Note implements NoteInterface {
 				map(([entries, warehouses]) => {
 					// Create a record of warehouse ids and names for easy lookup
 					const warehouseNames = warehouses.reduce((acc, { id, displayName }) => ({ ...acc, [id]: displayName }), {});
-					return entries.map((e) => ({ ...e, warehouseName: warehouseNames[e.warehouseId] || 'not-found' }));
+					return entries.map((e) => ({ ...e, warehouseName: e.warehouseId ? warehouseNames[e.warehouseId] : 'not-found' }));
 				}),
 				tap(debug.log(ctx, 'note:entries:stream:output'))
 			),
--- a/pkg/db/src/implementations/version-1.1/note.ts
+++ b/pkg/db/src/implementations/version-1.1/note.ts
@@ -4,7 +4,7 @@ import { debug } from '@librocco/shared';
 
 import { DocType, NoteState } from '@/enums';
 
-import { NoteType, VolumeStock, VersionedString } from '@/types';
+import { NoteType, VolumeStock, VersionedString, PickPartial } from '@/types';
 import { NoteInterface, WarehouseInterface, NoteData, DatabaseInterface } from './types';
 
 import { isVersioned, runAfterCondition, sortBooks, uniqueTimestamp, versionId } from '@/utils/misc';
@@ -211,15 +211,14 @@ class Note implements NoteInterface {
 	}
 
 	/**
-	 * Add volumes accepts an object or multiple objects of VolumeStockOptionalWarehouse updates (isbn, quantity, warehouse?) for
+	 * Add volumes accepts an object or multiple objects of VolumeStock updates (isbn, quantity, warehouseId?) for
 	 * book quantities. If a volume with a given isbn is found, the quantity is aggregated, otherwise a new
 	 * entry is pushed to the list of entries.
 	 */
 	addVolumes(...params: Parameters<NoteInterface['addVolumes']>): Promise<NoteInterface> {
 		return runAfterCondition(async () => {
 			params.forEach((update) => {
-				const warehouseId = versionId(update.warehouseId || '');
-
+				const warehouseId = this.noteType === 'inbound' ? this.#w._id : update.warehouseId ? versionId(update.warehouseId) : '';
 				const matchIndex = this.entries.findIndex(
 					(entry) => entry.isbn === update.isbn && entry.warehouseId === update.warehouseId
 				);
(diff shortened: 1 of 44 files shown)


pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -237,11 +232,7 @@ class Note implements NoteInterface {
 				};
 			};
 
-			if (isTuple(params)) {
-				updateQuantity(params[0], params[1], params[2]);
-			} else {
-				params.forEach((update) => updateQuantity(update[0], update[1], update[2]));
-			}
+			params.forEach((update) => updateQuantity(update.isbn, update.quantity, versionId(update.warehouseId || this.#w._id)));
```
ikusteu, line 235: There's one problem with falling back to `this.#w._id` and that is: we don't want any volume transaction to be assigned to `0-all` (default warehouse)...
In terms of UX, we should allow adding of transactions with no warehouses (and updated later), only we wouldn't allow for committing of notes without warehouse.

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -211,18 +211,13 @@ class Note implements NoteInterface {
...
 	addVolumes(...params: Parameters<NoteInterface['addVolumes']>): Promise<NoteInterface> {
-		// A check to see if the update function should be ran once or multiple times
-		const isTuple = (params: Parameters<NoteInterface['addVolumes']>): params is VolumeTransactionTuple => {
-			return !Array.isArray(params[0]);
-		};
-
 		return runAfterCondition(async () => {
-			const updateQuantity = (isbn: string, quantity: number, warehouseId = this.#w._id) => {
+			const updateQuantity = (isbn: string, quantity: number, warehouseId: `v${number}/${string}`) => {
 				const matchIndex = this.entries.findIndex((entry) => entry.isbn === isbn && entry.warehouseId === warehouseId);
```
ikusteu, line 221: This is ok, only you have `VersionedString` type for this usage (alias for `${number}/${string}`)

ikusteu: Also, since we don't need to check for `isTuple`, we don't need to declare an `updateQuantity` function, could just run its body inside the top level function.

pkg/db/src/types.ts
```diff
@@ -38,6 +38,13 @@ export interface VolumeStock {
 	warehouseId: VersionedString;
 }
 
+/**
+ * Used as param(s) for adding volumes to a note where the warehouseId could be optional and a plain string (non-versioned)
+ */
+export interface VolumeStockOptionalWarehouseId extends Omit<VolumeStock, 'warehouseId'> {
+	warehouseId?: string;
+}
```
ikusteu, line 46: With respect to some comments left already (being completely fine with some note transactions, temporarily, not having `warehouseId`, let's simply make the `warehouseId` optional in `VolumeStock` (this will also trickle down to `VolumeStockClient` (which is ok)

pkg/db/src/types.ts
```diff
@@ -83,11 +90,6 @@ export interface NoteStream {
 	entries: Observable<VolumeStockClient[]>;
 }
 
-/**
- * A tuple used as param(s) for adding volumes to a note: [isbn, quantity, warehouse?]
- */
-export type VolumeTransactionTuple = [string, number, VersionedString] | [string, number];
```
ikusteu, line 89: 🤘

pkg/db/src/types.ts
```diff
@@ -102,12 +104,7 @@ export interface NoteProto<A extends Record<string, any> = {}> {
...
 	/** Set name updates the `displayName` of the note. */
 	setName: (name: string, ctx: debug.DebugCtx) => Promise<NoteInterface<A>>;
-	/**
-	 * Add volumes accepts an array of volume stock transactions and adds them to the note.
-	 * If any transactions (for a given isbn/warehouse) already exist, the quantity gets aggregated.
-	 */
-	addVolumes: (...params: VolumeTransactionTuple | VolumeTransactionTuple[]) => Promise<NoteInterface<A>>;
-	/** Explicitly update an existing transaction row. */
+	addVolumes: (...params: VolumeStockOptionalWarehouseId[]) => Promise<NoteInterface<A>>;
 	updateTransaction: (transaction: VolumeStock) => Promise<NoteInterface<A>>;
```
ikusteu, line 108: Can you update the comments here (the previous comments were fine, just update them with the updates you've applied to the function)

ikusteu (changes requested):

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -250,12 +243,20 @@ class Note implements NoteInterface {
 	updateTransaction(transaction: VolumeStock): Promise<NoteInterface> {
 		// Create a safe copy of volume entries
 		const entries = [...this.entries];
-		const i = entries.findIndex(({ isbn }) => isbn === transaction.isbn);
+		// handle edge case where we have multiple books with the same isbn, but belonging to different warehouses
+		// match isbn only in case entry has no whId but transaction does (not undefined or empty string)
+		const i = entries.findIndex(
+			({ isbn, warehouseId }) =>
+				isbn === transaction.isbn &&
+				(warehouseId === versionId(transaction.warehouseId || '') || (warehouseId === versionId('') && transaction.warehouseId))
```
ikusteu, line 251: Yeah...this should be updated like the `addVolumes` `warehouseId` logic

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -250,12 +243,20 @@ class Note implements NoteInterface {
...
+		);
+
+		const versionedTransaction = { ...transaction, warehouseId: versionId(transaction.warehouseId || '') };
```
ikusteu, line 254: Maybe do this before trying to match the entry...less `warehouseId` conversions on the fly. 
Also: we don't want `v1/` as `warehouseId`

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -266,6 +267,9 @@ class Note implements NoteInterface {
 	 * when calculating the stock of the warehouse.
 	 */
 	commit(ctx: debug.DebugCtx): Promise<NoteInterface> {
+		if (this.entries.findIndex((entry) => entry.warehouseId === versionId('')) !== -1)
+			throw new Error('Cannot commit a note with empty warehouseId entries');
```
ikusteu, line 271: Just remove this, it's done in a different PR

pkg/db/src/implementations/version-1.1/warehouse.ts
```diff
@@ -256,7 +256,7 @@ class Warehouse implements WarehouseInterface {
 					const warehouseNames = warehouses.reduce((acc, { id, displayName }) => ({ ...acc, [id]: displayName }), {});
 					const res = entries.map((e) => ({
 						...e,
-						warehouseName: warehouseNames[e.warehouseId] || 'not-found'
+						warehouseName: (e.warehouseId && warehouseNames[e.warehouseId]) || 'not-found'
```
ikusteu, line 259: Just revert to the old one

ikusteu: `warehouseId` should always be defined

pkg/db/src/types.ts
```diff
@@ -35,7 +35,7 @@ export type DesignDocument = {
 export interface VolumeStock {
 	isbn: string;
 	quantity: number;
-	warehouseId: VersionedString;
+	warehouseId?: string;
 }
```
ikusteu, line 39: Ok...so going through the code, I see the confusion and complexity arising from `warehouseId` being optional. Let's make it a requited `string`, like so:
```ts
export interface VolumeStock {
  isbn: string
  quantity: number
  warehouseId: string
}
```

As for `addVolumes` / `updateTransaction`, you can use `PickPartial<VolumeTransaction, "warehouseId">[]` as params type (no need to create a special type).

`PickPartial` is a utility type, found in `@/types`

pkg/db/src/types.ts
```diff
@@ -103,11 +98,15 @@ export interface NoteProto<A extends Record<string, any> = {}> {
...
-	 * Add volumes accepts an array of volume stock transactions and adds them to the note.
-	 * If any transactions (for a given isbn/warehouse) already exist, the quantity gets aggregated.
+	 * Add volumes accepts an array of volume stock entries and adds them to the note.
+	 * If any transactions (for a given isbn and warehouse) already exist, the quantity gets aggregated.
+	 */
+	addVolumes: (...params: VolumeStock[]) => Promise<NoteInterface<A>>;
+	/** Explicitly update an existing transaction row.
+	 * the transaction is matched with both isbn and warehouseId.
+	 * If entry with the same isbn previously has no warehouseId and
+	 * a warehouseId was provided, the empty warehouseId will be overwritten
```
ikusteu, line 108: I wouldn't bother with non existing warehouse id being updates, we can do it as a second pass, let's, for now just match by isbn and warehouseId (as is).

Also, multiline JSDoc should start on the second line (not after `/**`)

pkg/db/src/utils/misc.ts
```diff
@@ -2,7 +2,7 @@ import { distinctUntilChanged, firstValueFrom, Observable, Subject, Subscription
 
 import { VersionedString, VolumeStock } from '../types';
 
-export const sortBooks = ({ isbn: i1, warehouseId: w1 }: VolumeStock, { isbn: i2, warehouseId: w2 }: VolumeStock) =>
+export const sortBooks = ({ isbn: i1, warehouseId: w1 = '' }: VolumeStock, { isbn: i2, warehouseId: w2 = '' }: VolumeStock) =>
```
ikusteu, line 5: Yeah...let's restore this (after the type is updated, as mentioned)

ikusteu (changes requested):

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -211,51 +211,52 @@ class Note implements NoteInterface {
...
 		const entries = [...this.entries];
-		const i = entries.findIndex(({ isbn }) => isbn === transaction.isbn);
+
+		const versionedTransaction = { ...transaction, warehouseId: transaction.warehouseId ? versionId(transaction.warehouseId) : '' };
+		// handle edge case where we have multiple books with the same isbn, but belonging to different warehouses
+		// match isbn only in case entry has no whId but transaction does (not undefined or empty string)
+		const i = entries.findIndex(
+			({ isbn, warehouseId }) =>
+				isbn === versionedTransaction.isbn &&
+				(warehouseId === versionedTransaction.warehouseId || (!warehouseId && transaction.warehouseId))
```
ikusteu, line 252: This is not handling all the warehouse update cases...I'll remove it (simplify so that it only performs updates requested in this ticket) and we should think about how to handle warehouse updates.

But it's a valid point and a functionality we'll definitely need.

ikusteu (approved):

PR #189, opened 2023-03-17
Enhancement/update transaction

### LINKED ISSUE #174: Update note.updateTransaction
Small edge case with updating of note transaction(s):

When we run `note.updateTransaction`, we pass the entire transaction (`VolumeStock`), but the transaction is matched by the `isbn`. We should update that so that the transaction is matched by `isbn` and `warehouseId` to handle a case where we have multiple books with the same `isbn`, but belonging to different warehouses.

_Note: this should really be a rarely occurring edge case, but I don't think it would need a lot of work to implement_

Tasks:

- [ ] Implement the behaviour
- [ ] Test the behaviour
- [ ] Accept unversioned `warehouseId` in `note.updateTransaction` and automatically prepend the warehouseId with version string internally, as mentioned for `note.addVolumes` in #171

fadwamahmoud:
Fixes #174 
Depends on #188


ikusteu:
Closing as it's included in #188

PR #202, opened 2023-03-22
Implement updating warehouseId on outbound notes

### LINKED ISSUE #201: Enhancement: updateTransaction should be able to also update warehouseId
With current implementation `note.updateTransaction` matches transactions by their `isbn` and `warehouseId`. This is satisfactory only in part, as it allows us to only update the `quantity` for a given transaction. Additionally, we would like to be able to update the `warehouseId` (for warehouse selection o outbound notes).

### To clarify the requirements: 

User should be able to click on any transaction in the note, change the warehouse for a given transaction and have the db support that kind of update. If we had the following:

```ts
[
  { isbn: "12345678", quantity 2, warehouseId: "wh-1" },
  { isbn: "12345678", quantity 2, warehouseId: "wh-2" },
  { isbn: "12345678", quantity 2, warehouseId: "" },
]
```
we should support update to any of those transactions, whether updating the quantity or the warehouse (or both), regardless of the order in which those transactions appear.

### Implementation

TBD

fadwamahmoud:
Fixes #201


pkg/db/src/__tests__/tests.ts
```diff
@@ -140,16 +140,26 @@ export const noteTransactionOperations: TestFunction = async (db) => {
 
 	// Update transaction should overwrite the existing transaction (and not confuse it with the same isbn, but different warehouse)
 	await note.updateTransaction({ isbn: '11111111', quantity: 3, warehouseId: wh1._id });
+	// Update transaction should not overwrite the existing warehouseId if it's empty
+	await note.updateTransaction({ isbn: '0123456789', quantity: 3 });
+
+	// Update transaction should overwrite the existing warehouseId
```
ikusteu, line 146: Can you add an assertion after each operation...it's easier to track the expected state in the db

pkg/db/src/__tests__/tests.ts
```diff
@@ -140,16 +140,26 @@ export const noteTransactionOperations: TestFunction = async (db) => {
...
+	await note.updateTransaction({ isbn: '0123456789', quantity: 4, warehouseId: wh2._id });
```
ikusteu, line 147: Can we  replace this line to, instead of updating transaction with isbn "012345678", update the transaction
```ts
{ isbn: "11111111", quantity: 10, warehouseId: versionId(wh2._id) }
```
to 
```ts
{isbn: "11111111", quantity: 10, warehouseId: "warehouse-3" }
```

ikusteu (changes requested):

pkg/db/src/__tests__/tests.ts
ikusteu: Bump

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -240,7 +240,13 @@ class Note implements NoteInterface {
...
 	}
 
-	updateTransaction({ isbn, quantity, warehouseId }: PickPartial<VolumeStock, "warehouseId">): Promise<NoteInterface> {
+	updateTransaction({
+		match: { isbn, quantity, warehouseId },
+		update
+	}: {
+		match: PickPartial<VolumeStock, "warehouseId">;
+		update: VolumeStock;
+	}): Promise<NoteInterface> {
```
ikusteu, line 249: Can you apply that signature, like we've discussed (adjusted with partial warehouseId):
```typescript
updateTransaction(match: PickPartial<Omit<VolumeStock, "quantity">, "warehouseId">, update: VolumeStock): Promise<NoteInterface>
```

ikusteu: It's less messy to consume, as well as less code in the declaration

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -250,11 +256,19 @@ class Note implements NoteInterface {
 			warehouseId: warehouseId ? versionId(warehouseId) : this.noteType === "inbound" ? this.#w._id : ""
 		};
 
-		const i = entries.findIndex((e) => e.isbn === transaction.isbn && e.warehouseId === transaction.warehouseId);
+		const updateTransaction = {
+			isbn: update.isbn,
+			quantity: update.quantity,
+			warehouseId: update.warehouseId ? versionId(update.warehouseId) : this.noteType === "inbound" ? this.#w._id : ""
+		};
```
ikusteu, line 263: Can we name this something else...it's a bit confusing (same name as the method)

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -250,11 +256,19 @@ class Note implements NoteInterface {
...
+
+		const i = entries.findIndex(
+			(e) => e.isbn === transaction.isbn && e.warehouseId === transaction.warehouseId && e.quantity === transaction.quantity
+		);
```
ikusteu, line 267: Quantity shouldn't be used to match the transaction...it's completely unnecessary

ikusteu (changes requested):
The tests seem ok (representative), just please update the signature

PR #213, opened 2023-03-29
Feature/hook up the app

### LINKED ISSUE #205: Hook up the functionality from UI to the db
I'm opening up this ticket more as a checklist for app integration. As @delasous mentioned, we should work towards hooking up the entire app with the "backend" (db interface), which would allow us to see all of the parts in action and pinpoint some additional features necessary (in the UI, or db interface), if any.

_Note: while writing this, I might have oversimplified some of the functionality, but the goal is to have it working the full round trip (add, delete, update, books, notes, etc...) and then extend with some eye/UX candy + additional features. One of additional features is definitely the search, which I've omitted all functionality and support for it in these tasks as I believe it should be handled in the next pass._

### TODO:

- [x] we should be able to add now books to a note:
  - the simplest solution I see here is to add the 'Scan to add books...' text input, have it accept (for now only user typed in) isbn and, on 'Create' button click run `note.addVolumes({ isbn, quantity: 1 })`:
     -  `isbn` would be the one provided to the input
     -  `quantity: 1` (as it could be specifically set on the transaction row, more below :point_down: )
     - `warehouseId` omitted as, that way, the db will automatically fall back to parent warehouse id (for inbound notes) or `""` for outbound notes (letting the user select the warehouse on transaction row more below 👇 
  - I understand that 'Create' button might have a different use case when the functionality is fully integrated (open up a create book dialog, allowing user to input the book data for the book), but for simplicity's sake, I think we can use it in this way now and then, later on, add an additional step (checking if book should be created or so)

- [ ] we should be able to update the volume transaction on the note:
  - this goes for both transaction `quantity` (as mentioned above) as well as `warehouseId` (in case of outbound notes)
  - `note.updateTransaction` can be used for both updates
  - my suggestion for quantity update is to toggle between quantity badge and input element for quantity (and maybe provide a submit button, used to update the quantity) 
  - outbound note -> warehouse selection -> available warehouses - I think, as a first pass, we can simply have all the warehouses as 'available'...as a further enhancement:
    - the transaction, added on outbound note, would use `warehouseId` of the only outbound warehouse that book is available in (if there's only one available warehouse for the book)
    - if there are multiple warehouses (or none) where the book is available, it should fall back to `""`
  - we can live without the aforementioned enhancements for now, as user won't be able to commit the note in case of any kind of incompatible `warehouseId`, moreover, the same functionality used to verify the transactions when committing (already in place in the db interface) can be extended to stream book availability in warehouses and used to achieve said enhancements (default fallback/available warehouse list)

- [ ] we should be able to remove transaction rows from the note:
  - in the simplest form, this would require adding `onDeleteRows` (or something like that) to the `InventoryTable` component and have it run `note.removeTransactions` for specified rows, whereas the table itself would be updated by db update (direct addition/removal to/from the table can be added as a second pass, but I don't see it being overly complicated either)

- [ ] we should signal why a particular note can't be committed:
  - we have db guards in place for this, throwing a serialised error with info about invalid transaction rows, blocking the commit, so we should use those to show a notification and/or highlight the invalid rows

ikusteu:
Fixes #205

### DIFF
--- a/pkg/db/src/__tests__/tests.ts
+++ b/pkg/db/src/__tests__/tests.ts
@@ -133,7 +133,7 @@ export const noteTransactionOperations: TestFunction = async (db) => {
 	await note.addVolumes(
 		// The add volumes operation should not confuse the transaction with the same isbn, but different warehouse
 		{ isbn: "0123456789", quantity: 3, warehouseId: wh1._id },
-		// This should also wor if warehouse is not provided (falls back to "", in case of outbound note)
+		// This should also work if warehouse is not provided (falls back to "", in case of outbound note)
 		{ isbn: "11111111", quantity: 7 }
 	);
 	await waitFor(() => {
--- a/pkg/db/src/__tests__/tests.ts
+++ b/pkg/db/src/__tests__/tests.ts
@@ -115,27 +115,28 @@ export const noteTransactionOperations: TestFunction = async (db) => {
 		{ isbn: "0123456789", quantity: 2, warehouseId: wh1._id },
 		// Having the same isbn for different warehouses will come in handy when testing update/remove transaction
 		{ isbn: "11111111", quantity: 4, warehouseId: wh1._id },
-		{ isbn: "11111111", quantity: 3, warehouseId: wh2._id }
+		{ isbn: "11111111", quantity: 3 }
 	);
 	await waitFor(() => {
 		expect(entries).toEqual([
 			{ isbn: "0123456789", quantity: 2, warehouseId: versionId(wh1._id), warehouseName: "Warehouse 1" },
-			{ isbn: "11111111", quantity: 4, warehouseId: versionId(wh1._id), warehouseName: "Warehouse 1" },
-			{ isbn: "11111111", quantity: 3, warehouseId: versionId(wh2._id), warehouseName: "Warehouse 2" }
+			{ isbn: "11111111", quantity: 3, warehouseId: "", warehouseName: "not-found" },
+			{ isbn: "11111111", quantity: 4, warehouseId: versionId(wh1._id), warehouseName: "Warehouse 1" }
 		]);
 	});
 
 	// Adding volumes to the same ISBN/warheouseId pair should simply aggregate the quantities
 	await note.addVolumes(
 		// The add volumes operation should not confuse the transaction with the same isbn, but different warehouse
-		{ isbn: "11111111", quantity: 3, warehouseId: wh1._id },
-		{ isbn: "11111111", quantity: 7, warehouseId: wh2._id }
+		{ isbn: "0123456789", quantity: 3, warehouseId: wh1._id },
+		// This should also wor if warehouse is not provided (falls back to "", in case of outbound note)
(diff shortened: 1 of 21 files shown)


pkg/db/src/__tests__/tests.ts
```diff
@@ -115,27 +119,28 @@ export const noteTransactionOperations: TestFunction = async (db) => {
...
 		]);
 	});
 
 	// Adding volumes to the same ISBN/warheouseId pair should simply aggregate the quantities
 	await note.addVolumes(
 		// The add volumes operation should not confuse the transaction with the same isbn, but different warehouse
-		{ isbn: "11111111", quantity: 3, warehouseId: wh1._id },
-		{ isbn: "11111111", quantity: 7, warehouseId: wh2._id }
+		{ isbn: "0123456789", quantity: 3, warehouseId: wh1._id },
+		// This should also wor if warehouse is not provided (falls back to "", in case of outbound note)
```
fadwamahmoud, line 136: work*

pkg/db/src/implementations/version-1.1/warehouse.ts
```diff
@@ -23,6 +23,14 @@ class Warehouse implements WarehouseInterface {
 
 	#initialized = new BehaviorSubject(false);
 	#exists = false;
+
+	// Update stream receives the latest document (update) stream from the db. It's multicasted using plain RxJS Subject (no repeat or anything).
+	// This stream is used to signal the update has happened (and has been streamed to update the instance). Subscribers needing to be notified when
+	// and update happens should subscribe to this stream.
```
fadwamahmoud, line 29: an update*

pkg/db/src/implementations/version-1.1/note.ts
```diff
@@ -19,6 +19,14 @@ class Note implements NoteInterface {
 
 	#initialized = new BehaviorSubject(false);
 	#exists = false;
+
+	// Update stream receives the latest document (update) stream from the db. It's multicasted using plain RxJS Subject (no repeat or anything).
+	// This stream is used to signal the update has happened (and has been streamed to update the instance). Subscribers needing to be notified when
+	// and update happens should subscribe to this stream.
```
fadwamahmoud, line 25: an update*

pkg/ui/src/InventoryTable/__tests__/InventoryTable.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+		const table = createTable(tableOptions);
+
+		const { component } = render(OutNoteTable, { table });
+		component.$on("transactionupdate", (e) => mockOnUpdate(e.detail));
+
+		// Update the warehouse of the first row (wh1 -> wh2)
+		screen.getAllByRole("combobox")[0].click();
+		await act(() => {
+			// Even though other comboboxes are closed, the selection is merely hidden my CSS (so still in the DOM),
```
fadwamahmoud, line 32: by*

fadwamahmoud (approved):

PR #236, opened 2023-05-24
Feature/sveltekit pwa plugin

fadwamahmoud:


apps/web-client/vite.config.js
```diff
@@ -1,13 +1,62 @@
...
+}
+
 /** @type {import('vite').UserConfig} */
 const config = {
-	plugins: [sveltekit()],
+	plugins: [
+		sveltekit(),
+		SvelteKitPWA({
+			buildBase: `${BASE_PATH}/`,
+			// workbox: {
```
silviot, line 28: This should probably be removed

apps/web-client/package.json
```diff
@@ -40,7 +40,10 @@
 		"@esbuild-plugins/node-modules-polyfill": "~0.1.4",
 		"rollup-plugin-node-polyfills": "~0.2.1",
 		"autoprefixer": "~10.4.13",
-		"@types/uuid": "~9.0.0"
+		"@types/uuid": "~9.0.0",
+		"vite-plugin-pwa": "~0.14.7",
+		"@vite-pwa/sveltekit": "~0.2.2",
+		"workbox-window": "~6.5.4"
```
silviot, line 46: Do we really need to depend on this? Since we're not using it directly [the transient dependency](https://github.com/vite-pwa/vite-plugin-pwa/blob/main/package.json#L108) via `vite-plugin-pwa` should be enough, right?

silviot (changes requested):
Just a couple of minor changes needed IMO.

apps/web-client/package.json
fadwamahmoud: We need it because when I try uninstalling it, I get this error
`Rollup failed to resolve import "workbox-window" from "/@vite-plugin-pwa/virtual:pwa-register"`
I tried removing node_modules and re-installing but I get the same error

silviot: I see. I'd love to include this info into the package.json, next to that line, but comments are not allowed there unfortunately.

apps/web-client/vite.config.js
```diff
@@ -1,13 +1,56 @@
...
+
+const dev = process.env.NODE_ENV === "development";
+const CURRENT_SHA = process.env.CURRENT_SHA;
+let BASE_PATH = process.env.BASE_PATH;
+
+if (typeof BASE_PATH === "undefined") {
+	// If no BASE_PATH was passed as environment variable, we default to either
+	// `/preview` or (if CURRENT_SHA is defined) to `/preview/<CURRENT_SHA>`
+	BASE_PATH = dev || !CURRENT_SHA ? DEFAULT_BASE_PATH : `${DEFAULT_BASE_PATH}/${CURRENT_SHA}`;
+	process.env.BASE_PATH = BASE_PATH;
```
delasous, line 19: Be helpful to leave a small comment here explaining why we are doing this: so that the default base path will be communicated to the `svelte.config`

apps/web-client/src/routes/+layout.svelte
```diff
@@ -1,8 +1,38 @@
...
+				onRegistered(r) {
+					r &&
+						setInterval(() => {
+							console.log("Checking for sw update");
+							r.update();
+						}, 20000);
+					console.log(`SW Registered: ${r}`);
+				},
+				onRegisterError(error) {
+					console.log("SW registration error", error);
```
delasous, line 23: We should do something with this. Once toast library is added, it could potentially be communicated to client, or is it unimportant to them? What are main reasons it could fail? Should we retry?

fadwamahmoud: Registration could fail if the script fails to download, parse, or throws an error in its initial execution, so, it should be successful unless there's something really wrong like syntax error or trying to access a non-existing resource in the cache, in which case, I don't think retrying would be of much use. If we decide to communicate this to the client, we could display a toast of something along the lines of "The web app cannot work offline if connection is lost" @delasous

delasous: I think we should definitely toast the successful install so that the user knows the app will continue to work offline. But I'm not sure I see a real reason for them to know when it fails as they can't do much about it. Best thing for now would probably be to log the error with Sentry so that we're made aware of it. @silviot is this already setup/available for Librocco?

delasous (approved):

silviot (approved):

PR #247, opened 2023-06-15
Feature/existing isbn book form

### LINKED ISSUE #245: Feature: Existing ISBN on book detail form
Suggested solution:

- Create new find method on book interface that checks whether isbn exists in db and returns the book if it does
- it gets called when user clicks away from isbn field
- If isbn exists disable save button and display an validation message saying `book with this isbn already exists`
- we could also provide a link in the error message so that it says `book with this isbn already exists, Edit`? which if clicked fills in the form and disables the isbn field

Filling in the form with the book details right after they enter an existing isbn could complicate things as the isbn field will be editable and the book details could be overwritten

What do you guys think? @silviot @delasous @ikusteu  

Tasks:

- [x] Create validation on debounce
- [x] disable submit button on isbn error
- [x] Create cache store for same isbn (to prevent multiple api calls)
- [x] Fill in form on edit click in error message
    Make helpText clickable

fadwamahmoud:
Fixes #245


silviot:
I found this problem while testing this:

https://github.com/librocco/librocco/assets/396425/6609e126-e6cf-4fe2-9093-335646328d4f


i.e. after clicking to edit the book the form populated correctly, but then the error message appeared again.

pkg/ui/src/Forms/BookDetailForm.svelte
```diff
@@ -12,23 +12,103 @@
...
+				const errors: {
+					isbn?: string;
+					title?: string;
+					price?: string;
+					year?: string;
+					authors?: string;
+					publisher?: string;
+					editedBy?: string;
+					outOfPrint?: string;
+				} = { isbn: isbnExistsError };
```
delasous, line 49: I would work with the errors object here, return it, then subscribe to and use e.g `$errors.isbn` store returned from `createForm`, rather than managing `isbnExistsError`. https://felte.dev/docs/svelte/validation. Note for some reason `$errors` returns arrays for each field, not sure why...

pkg/ui/src/Forms/BookDetailForm.svelte
```diff
@@ -12,23 +12,103 @@
 	export let book: BookEntry;
 	export let publisherList: string[] = [];
 	export let onSubmit: (values: BookEntry) => void = () => {};
+	export let editMode: boolean = false;
+	export let openEditMode = () => {};
 	export let onCancel = () => {};
+	export let onISBNField: (isbn: string) => Promise<BookEntry[]> | undefined = () => undefined;
```
delasous, line 18: I would generalise this to `onValidate` and pass all "values" as param

ikusteu (commented):
LGTM, other than some very minor semantic conventions (which I've mentioned). I've pointed out where I think might lie the problem for @silviot's reported bug, but I'm not really sure.

pkg/ui/src/Forms/BookDetailForm.svelte
```diff
@@ -12,23 +12,103 @@
...
 		}
 	});
 
+	const onEdit = () => {
+		openEditMode();
+
+		// reset isbn error on edit
+		isbnExistsError = "";
+
+		setFields({ year: "", authors: "", publisher: "", editedBy: "", outOfPrint: false, ...bookCache[getValue($data, "isbn")] });
```
delasous, line 83: Do you need to use `getValue` here? Could you not just create a new variable which references current state of `$data`?

pkg/ui/src/FormFields/TextField/TextField.svelte
```diff
@@ -20,11 +21,13 @@
 	export let variant: TextFieldSize = TextFieldSize.Base;
 	export let error = false;
 	export let value = "";
+	export let onHelpTextClick = () => {};
 	$: error = error;
 
 	const labelBaseClasses = ["block", "text-sm", "font-medium", "text-gray-700"];
 	const helpTextBaseClasses = ["ml-[2px]", "text-sm", "min-h-[20px]"];
 	const inputBaseClasses = ["block", "w-full", "border-0", "text-sm", "focus:outline-0", "focus:ring-0", variant];
+	const clickableHelpText = onHelpTextClick ? "cursor-pointer underline" : "";
```
delasous, line 30: This will always evaluate to true because you've assigned default `() => {}` - even this empty function is truthy.

pkg/ui/src/FormFields/TextField/TextField.svelte
```diff
@@ -61,6 +64,6 @@
 		{/if}
 	</div>
 	{#if helpText}
-		<p class={helpTextClasses}>{helpText}</p>
+		<p on:keypress={onHelpTextClick} on:click={onHelpTextClick} class={helpTextClasses}>{helpText}</p>
```
delasous, line 67: This component is definitely overloaded with responsibility now. Can we define a "helpText" slot and manage all of this in the book form itself? Keep only the `error` prop (renamed to `isValid`) for now so that the rest of the component knows how to style itself when there's a validation error

pkg/ui/src/Forms/BookDetailForm.svelte
```diff
@@ -12,23 +12,103 @@
...
 		<div class="flex grow flex-col flex-wrap gap-y-4 lg:flex-row">
 			<div class="basis-full">
-				<TextField name="isbn" label="ISBN" required={book.isbn === ""} disabled={book.isbn !== ""} />
+				<TextField
+					name="isbn"
+					label="ISBN"
+					required={getValue($data, "isbn") === ""}
+					disabled={editMode}
+					error={Boolean(isbnExistsError)}
+					helpText={isbnExistsError}
```
delasous, line 109: Pass clickable error message here as `helpText` slot. Use this aria attribute on element: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live

delasous (changes requested):

pkg/ui/src/Forms/BookDetailForm.svelte
```diff
@@ -12,23 +12,103 @@
...
+
+				if (values.isbn.trim() === "" || $interacted !== "isbn") return errors;
+
+				// reset isbn error if isbn field interacted with
+				isbnExistsError = "";
+
+				if (bookCache[values.isbn]) {
+					errors.isbn = isbnError;
+					isbnExistsError = isbnError;
+				} else {
```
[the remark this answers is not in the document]
fadwamahmoud: yeah, the book is still cached, but this would only be checked for if the isbn field was interacted with which it won't be when the user clicks edit/autofill

delasous:
Code updates look good... only nit-picky thing would be to re-introduce `error` prop and input error styles to TextField - but I'm still seeing same behaviour in preview that Silvio highlighted

fadwamahmoud:
> Code updates look good... only nit-picky thing would be to re-introduce `error` prop and input error styles to TextField. I'm also still seeing same behavior in preview that Silvio highlighted - is this an old preview or does the bug still exist?

So change isValid to error and add another prop for the error styles? @delasous 
 I tried checking out the bug again and it's gone, so maybe you checked an older preview link.

delasous:
Ok yeah, sorry! I missed the `error -> isValid` change, and I was looking at an old link. Nice work!

delasous (approved):

PR #248, opened 2023-06-21
Fix/optimisations 1

ikusteu:
The main goal of the ticket was to improve the speed of navigation between views (and I managed to reduce the bottleneck from 1300ms to 300ms) + added some small fixes here and there (don't load the design docs on each view switch, preventing larger and larger choke of JS engine)


silviot (approved):
🖱️🦭: "I clicked around" Seal of Approval

And I clicked around fast! And the UI was updating faster than I can click. That's how all UIs should be!

PR #254, opened 2023-06-23
Cache the full stock in the db instance

### LINKED ISSUE #249: Optimisation: cache the stock inside the db instance
In #248 I've reduced the bottleneck of querying the design document to get entries (from 1300ms to 300ms).
However, at this point all of the 'stock' streams query the entire 'addDocs' (namsepaced with the current version -> to process only note/warehouse documents) and the stock (for a warehouse) is kept inside the warehouse instance (as cache) and multicasted to all subscribing streams.
I think we can optimise this further and cache the entire state inside the db instance (kept alive throughout the app usage):
- the number of document queries is always the same (regardless of warehouse)
- we can calculate the stock for '0-all' (default warehouse) and simply pick out the entries belonging to a particular warehouse
- there might be an issue of memory size, but I don't imagine it being significantly larger than what we have now (with only the particular warehouse stock "cached")
- the cache would get updated when there's a change in stock (as it would still be a multicasted observable)
- the book data is not cached - it's quite fast as we only take the 10 rows to display and then do `allDocs` query with 10 explicitly specified document ids (isbns in this case)

The following diagrams represent the current and proposed data flows.
The strong line indicates the data flow for "current" warehouse in view, whereas a dotted line indicates data flow which would be processed on each change of warehouse in view:

## Current implementation
```mermaid
graph TD;
db("db")
aa(" ")
ba(" ")
ca(" ")
ab(" ")
bb(" ")
cb(" ")
ac(" ")
bc(" ")
cc(" ")
ad(" ")
bd(" ")
cd(" ")
ae("Multicast warehouse-stock cache")
be("Multicast warehouse-stock cache")
ce("Multicast warehouse-stock cache")
af(" ")
bf(" ")
cf(" ")
ag(" ")
bg(" ")
cg(" ")
ah(" ")
bh(" ")
ch(" ")

db == "allDocs result (note and warehouse documents)" ==> aa
db -. "allDocs result (note and warehouse documents)" .-> ba
db -. "allDocs result (note and warehouse documents)" .-> ca
aa == "filter out warehouse and non-committed-note documents" ==> ab
ab == "flat map entries (transactions)" ==> ac
ac == "filter out irrelevant transactions (0-quantity, non current warehouse)" ==> ad
ad == "reduce the transactions to a { isbn-warehouseId => quantity } Map (current warehouse stock)" ==> ae
ae ==> af
ae ==> ag
ae ==> ah

ba -. "filter out warehouse and non-committed-note documents" .-> bb
bb -. "flat map entries (transactions)" .-> bc
bc -. "filter out irrelevant transactions (0-quantity, non current warehouse)" .-> bd
bd -. "reduce the transactions to a { isbn-warehouseId => quantity } Map (current warehouse stock)" .-> be
be -..-> bf
be -..-> bg
be -..-> bh

ca -. "filter out warehouse and non-committed-note documents" .-> cb
cb -. "flat map entries (transactions)" .-> cc
cc -. "filter out irrelevant transactions (0-quantity, non current warehouse)" .-> cd
cd -. "reduce the transactions to a { isbn-warehouseId => quantity } Map (current warehouse stock)" .-> ce
ce -..-> cf
ce -..-> cg
ce -..-> ch
```

## Proposed implementation:

```mermaid
graph TD;
a("allDocs result (note and warehouse documents)")
b(" ")
c(" ")
d(" ")
e("Multicast full-stock cache")
f("warehouse 1")
g("warehouse 2")
h("warehouse 3")

a == "filter out warehouse and non-committed-note documents" ==> b
b == "flat map entries (transactions)" ==> c
c == "filter out irrelevant transactions (0-quantity)" ==> d
d == "reduce transactions to a { isbn-warehouseId => quantity } Map (entire stock)" ==> e
e == "pick particular stock entries" ==> f
e -. "pick particular stock entries" .-> g
e -. "pick particular stock entries" .-> h
```

So...the difference in processing is filtering out on the last step vs doing the entire process anew each time. 
Considering the entire process takes around 300-350ms and the most expensive parts are the `allDocs` (around 170-200ms) and the process until the stock is calculated (the stock Map is created).
If we go with the proposed solution, I imagine the first process (when the app is loaded) would take the same (300-350ms) and absorb most of the processing time, providing for seamless transitions between views as the processing time of picking out the relevant stock entries (from cached stock Map) would be insignificant compared to the entire pipeline.

In terms of the memory, we wouldn't get much of the overhead:
- In case of `0-all` (default) warehouse, displaying full stock, the cached stock (in current implementation) is exactly the same as the stock cached in proposed solution at all times
- In case of particular warehouse, the cached stock is significantly smaller (only the entries for a particular warehouse are cached)
- additionally, when switching from default warehouse to a particular one, we might end up with both structures in memory, until the GC has had time to collect the default warehouse (full stock), providing for unnecessary duplication in the mean time (warehouse "foo"'s stock is a subset of default warehouse's stock)

With respect to the points above, my conclusion is:
- in case of default warehouse there is zero overhead (the cached structure is the same)
- in case of any particular warehouse the overhead is there (compared to caching only the particular warehouse), but no more than making the memory usage the same as in case of default warehouse
- when switching between warehouses, we would actually reduce (possible) duplication in memory as well as the need for GC to collect the difference

I've given this a lot of thought and seems like a solution with little to no cost and all the benefit. 
@delasous @fadwamahmoud @silviot, let me know your thoughts. The implementation should as trivial as moving a piece of code from one place to another.

ikusteu:
Cache the stock full stock in the db instance (significantly improving visual response times in between views)
Fixes #249

### DIFF
--- a/pkg/db/src/implementations/version-1.2/db.ts
+++ b/pkg/db/src/implementations/version-1.2/db.ts
@@ -1,9 +1,19 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { BehaviorSubject, firstValueFrom, map, Observable, share, tap } from "rxjs";
+import { BehaviorSubject, firstValueFrom, map, Observable, ReplaySubject, share, tap } from "rxjs";
 
 import { debug } from "@librocco/shared";
 
-import { BooksInterface, CouchDocument, DbStream, DesignDocument, InNoteList, MapReduceRow, NavListEntry, Replicator } from "@/types";
+import {
+	BooksInterface,
+	CouchDocument,
+	DbStream,
+	DesignDocument,
+	InNoteList,
+	MapReduceRow,
+	NavListEntry,
+	Replicator,
+	VolumeStock
+} from "@/types";
 import { DatabaseInterface, WarehouseInterface, WarehouseListRow, OutNoteListRow, InNoteListRow } from "./types";
 
 import { NEW_WAREHOUSE } from "@/constants";
@@ -15,6 +25,7 @@ import { newDbReplicator } from "./replicator";
 import { newView } from "./view";
 
 import { scanDesignDocuments } from "@/utils/pouchdb";
+import { newStock } from "./stock";
 
 class Database implements DatabaseInterface {
 	_pouch: PouchDB.Database;
@@ -24,6 +35,8 @@ class Database implements DatabaseInterface {
 	#warehouseListStream: Observable<NavListEntry[]>;
 	#outNoteListStream: Observable<NavListEntry[]>;
 	#inNoteListStream: Observable<InNoteList>;
+
+	#stockStream: Observable<VolumeStock[]>;
 
 	constructor(db: PouchDB.Database) {
 		this._pouch = db;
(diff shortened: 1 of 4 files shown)


ikusteu:
Merging this in as it's been open for awhile, produces very little change to the code, but a big improvement and I would like to reconcile everything.

PR #255, opened 2023-06-23
Fix/small fixes 5

### LINKED ISSUE #207: Small updates
### Some small low hanging fruit which can be handled as a single unit of work

- [x] **Note should not be reachable from the wrong warehouse**

This isn't really something that's necessary for our current use case, but it shouldn't be too hard to implement (as an additional db guard of sorts). Currently If I were to do `db().warehouse("wh1").note("wh2/inbound/note-1")` I could still reach the note, what's more, the `NoteInterface` would, internally keep back reference to "wh1" as parent warehouse (which isn't the case in the db). So far we've been good citizens and have taken care of that, but this should not be possible...the simplest solution, I think, would be to add a validation function while setting up the `NoteInterface` instance and have it throw an error if internal warehouse and note warehouse mismatch (which should easily be done with `v1` data model as warehouse id is included in note's id/path)

- [x] **Create an imperative query for note/warehouse entries**

Currently all note/warehouse data, consumed by the client, is delivered in form of streams, but, sometimes, we don't need streams and, in fact, streams can cause some unnecessary performance toll (as explained in #206). After the aforementioned ticket is in, we should create a way to, imperatively, retrieve the entries data (for full text search purposes, for instance).

The method can be `getEntries()` or something like that, and it should (for now) return all entries in note/warehouse stock.

- [x] **Don't show the table when there's no table data**

Currently, if there's no table data, the table heading row is shown, which is kinda weird...this can be a simple fix, or it might need some working around...(I don't know how the table flows are organised in that regard, @delasous ?? )

ikusteu:
- Provide fallbacks to display "" or "N/A" in the table, rather than "null" or "undefined" (for books with no `year`, `authors`, `editedBy`, etc...)
- Remove trailing slash when using `findNote` (reloading note page breaks otherwise)
- Fixes #207 (points 1 and 2...3 was already solved)

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -23,7 +23,6 @@ importers:
       eslint: ^8.16.0
       eslint-config-prettier: ^8.3.0
       eslint-plugin-svelte3: ^4.0.0
-      fast-glob: ~3.2.12
       lucide-svelte: ^0.210.0
       postcss: ~8.4.13
       pouchdb: ~7.3.0
--- a/common/config/rush/repo-state.json
+++ b/common/config/rush/repo-state.json
@@ -1,5 +1,5 @@
 // DO NOT MODIFY THIS FILE MANUALLY BUT DO COMMIT IT. It is generated and used by Rush.
 {
-  "pnpmShrinkwrapHash": "aaaa092591815fd39b057b7ff46f1bdcb637d052",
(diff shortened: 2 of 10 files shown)


ikusteu:
Merging this in as it's been open for awhile, produces very little change to the code, but a big improvement (removes some small but annoying problems) and I would like to reconcile everything.

PR #260, opened 2023-06-24
Rehaul/iterators 1

### LINKED ISSUE #257: Iterables 1: Create functions to iterate over iterables
Instead of using `arr.map(mapper)`, we should be able to use `map(arr, mapper)`. Furthermore `map(iterable, mapper)` should work on any iterable (not only the array).

I've already started on this and have basic methods. Some problems we need to solve:

- be able to compose them:

```ts
// Instead of writing this
const mapped = map(iterable, mappingFn1)
const filtered = filter(mapped1, filterFn)
const transformed = map(filtered, mappingFn2)

// I would like to be able to do
const transformed = compose(
  iterable,
  map(mappingFn1),
  filter(filterFn),
  map(mappingFn2)
)
```
This is easily achievable, but the type inference for it...not so much. I've tried to copy what the RxJS dis for their `pipe` method, but to no avail.

- have the returned structures be reusable:

We should either return a generator and have the consumer spread it into an array for reusability, return a generator function (instead of a generator), so that we achieve the reusability by forcing the consumer to call the function (and receive a new generator) when iterating over it, or...create a data structure which would allow for both.

For motivation and reasoning, refer to #250

ikusteu:
Fixes #257 

The reusability has been achieved by creating an iterable from generator on each transformation:
- the resulting iterable is as lean as the generator function: has only one property (`[Symbol.iterator]`) and doesn't hold internal iterable, it might (and probably will) contain reference to the iterable the generator function had captured in it's closure
- the resulting iterable is reusable (unlike the generator) - the `[Symbol.iterator]` property of the resulting iterable is actually the generator function (passed as argument), so each time a consumer acquires the iterator, it actually produces a new instance of the same generator (returned from the generator function)

Composition was achieved by wrapping the iterable in `TransformableIterable`, providing it with `map`, `flatMap`, `filter` and `reduce`, which can be chained together (as they all return an instance of `TransformableIterable`), but this kind of iterable is much lighter than an array (for instance)

`wrapIter(iterable)` (wrapping the `iterable` into `TransformableIterable`) is used in a similar way you would add a trait to a `struct` in Rust with the difference that, in case of `wrapIter`, any only the `[Symbol.iterator]` is taken from the input iterable, whereas the other methods the input might have are lost (e.g. `Map.prototype.keys`, etc.)

The resulting api is more/less the same as on arrays:
```ts
// Transformations using array methods
const arr = [0, 1, 2, 2]
const arrRes = arr
    .map(mapper)
    .filter(predicate)
    .flatMap(mapper)
    .reduce(reducer)

// For simplicity's sake we're using an array as an Iterable here
const iterable = wrapIter(arr)
const result = iterable
    .map(mapper)
    .filter(predicate)
    .flatMap(mapper)
    .reduce(reducer)
```
As you can see, the API is almost exactly the same (save wrapping with `wrapIter`)    

I've used an array as an example, but we can do this for any iterable:
```ts
const iterable = wrapIter(new Set(arr))
const result = iterable
    .map(mapper)
    .filter(predicate)
    .flatMap(mapper)
    .reduce(reducer)
```

### DIFF
--- a/pkg/shared/src/generators.test.ts
+++ b/pkg/shared/src/generators.test.ts
@@ -1,6 +1,6 @@
 import { describe, test, expect } from "vitest";
 
-import { iterableFromGenerator, wrapIter } from "./generators";
+import { EmptyIterableError, filter, flatMap, iterableFromGenerator, map, reduce, wrapIter } from "./generators";
 
 describe("iterableFromGenerator", () => {
 	test("should allow for multiple iterations over the same generator", () => {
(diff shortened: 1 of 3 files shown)


fadwamahmoud (approved):
Looks great!

PR #266, opened 2023-06-30
Rehaul/iterators 2

### LINKED ISSUE #258: Iterables 2: Replace array methods with functions over iterables
In #257 we're creating functions able to run over iterables (the way array methods iterate over arrays). After they're created, we can replace array methods in our data processing pipelines with said functions.

We don't need to replace the data types yet. Aforementioned functions should would work on arrays as well (as arrays implement iterable interface) and, when handing the data over from function to function, we can simply spread it into an array (to satisfy call signatures/expected types).

For motivation and reasoning, refer to #250

### LINKED ISSUE #259: Iterators 3: Replace array/record data structures with their iterable counterparts
Continuation from #258 

- replace instances of object (`Record<K, V>`) with instances of `Map<K, V>`
- replace instances of `Array<T>` with instances of `Iterable<T>`

Arrays satisfy the iterable interface, so the same functionality can be reused even if we pass an array.

Record types should definitely be replaced with Maps as they're convenient for both lookups as well as iterations (map is an iterable and exposes convenient `.keys()` and `.values()` being iterables themselves.

Finally, when we want to display the data in the UI, Svelte also supports iterables (it's not only constrained to arrays) as its `{#each .. as}` control flow is akin to `for .. of`

For motivation and reasoning, refer to #250

### LINKED ISSUE #264: Create a VolumeStockMap class building on top of Map
While working on #259 I've come up with this structure. In spirit of agility, I'm separating this as its own ticket as:
- it's some amount of work
- it will provide additional functionality (beyond the aforementioned ticket)

### Abstract

When using `Map`, this type is perfectly legal: `Map<[string, string], any>` - allowing us to key the Map with an array (or, in this case a tuple. However, in this case, the tuple is compared by reference (as is understandable in JS design):

```ts
const m = new Map<[string, string], string>()

// Example of using the same key when accessing an element
const key = ["foo", "bar"]
m.set(key, "fizz")
m.set(key, "buzz")
// Results in 'm' being: Map { ["foo", "bar"] => "buzz" }

// Example using different instances of tuple with same values
m.clear()
m.set(["foo", "bar"], "fizz")
m.set(["foo", "bar"], "buzz")
// Results in 'm' being:
// Map {
//   ["foo", "bar"] => "fizz"
//   ["foo", "bar"] => "buzz"
// }
```

This is understandable as both calls to `.set` method (in the latter example) are done with different array instance <=> different reference value.

What I would like to do with `VolumeStockMap` is to make it so that we're able to access the same element if the tuple (used as key) is deeply equal as the one "stored in the map".
The most straightforward way to achieve this that comes to mind is to use `Map<string, any>` and expose all of the methods implementing the `Map<[string, string], any>` interface, where each method will simply join tuple values into a string.

```ts
class TupleIndexableMap implements Map<[string, string], any> {
    #internal = new Map<string, any>
    
    // Get accepts a tuple of '[string, string]' 
    get(key: [string, string]) {
        // Value is retrieved by joining the tuple into a string, like so: ["foo", "bar"] => "foo:bar"
        return this.#internal.get(key.join(":"))
    }

    // ...implement the rest of the Map<[string, string], any> interface in the same way
}
```

### Motivation

As mentioned, this came to my when working on converting array values in our data processing pipelines in to their iterable counterparts (preferably Map) as Maps are ordered in the same way arrays are, but are also accessible using `.get` method (removing the need for `array.find`). This takes that rationale further and allows us to access the transaction by its isbn and warehouse without too much hassle (on the surface level) and, virtually no performance overhead.

Additionally, beyond the #259 ticket, this can be used in the same way (only slightly more convenience) it already is used when caching `stock` for the entire shop, and allowing to pick only a subset of entries for a particular warehouse, but could also be used to implement the available warehouses by isbn, for which this functionality is not yet implemented (we're using design documents for that, which were proven to not be very efficient)

### Requirements

- [ ] Create a `StockMap` class mimicking the `Map<[string, string], Row>`, but evaluating tuples by value
- [ ] Create additional `.aggregate` method taking in an `Iterable<VolumeStock>` and aggregating the resulting stock

ikusteu:
Fixes #258 
Fixes #259 
Fixes #264

### DIFF
--- a/pkg/shared/src/StockMap.test.ts
+++ b/pkg/shared/src/StockMap.test.ts
@@ -1,0 +1,73 @@
+import { describe, test, expect } from "vitest";
+
+import { StockMap } from "./StockMap";
+
+describe("StockMap", () => {
+	test("should index the element using tuple as if value, not reference", () => {
+		const map = new StockMap();
+
+		map.set(["12345678", "wh1"], { quantity: 10 });
+
+		expect(map.has(["12345678", "wh1"])).toEqual(true);
+		expect(map.get(["12345678", "wh1"])).toEqual({ quantity: 10 });
+
+		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+		map.get(["12345678", "wh1"])!.quantity = 20;
+
+		expect(map.get(["12345678", "wh1"])).toEqual({ quantity: 20 });
+
+		expect(map.delete(["12345678", "wh1"])).toEqual(true);
+
+		expect(map.has(["12345678", "wh1"])).toEqual(false);
+	});
+
+	test("should keep the 'size' the same value as the internal map", () => {
+		const map = new StockMap();
+
+		expect(map.size).toEqual(0);
+
+		map.set(["12345678", "wh1"], { quantity: 10 });
+		map.set(["12345678", "wh2"], { quantity: 10 });
+
+		expect(map.size).toEqual(2);
+
+		map.delete(["12345678", "wh1"]);
+
+		expect(map.size).toEqual(1);
+
+		map.clear();
+
+		expect(map.size).toEqual(0);
+	});
+
+	// This is hiding the fact that the internal map is a Map<string, { quantity: number }>
+	test("should iterate over the elements as if the underlaying map is a Map<[string, string], { quantity: number }>", () => {
+		const map = new StockMap();
+
+		map.set(["12345678", "wh1"], { quantity: 10 });
+		map.set(["12345678", "wh2"], { quantity: 10 });
+
+		const values = [...map.values()];
+		expect(values).toEqual([{ quantity: 10 }, { quantity: 10 }]);
+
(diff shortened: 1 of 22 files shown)


fadwamahmoud (approved):

pkg/shared/src/StockMap.test.ts
```diff
@@ -0,0 +1,218 @@
...
+		// All of the iterators should be reusable
+		expect([...map]).toEqual(entries);
+		expect([...map.values()]).toEqual(values);
+		expect([...map.keys()]).toEqual(keys);
+		expect([...map.entries()]).toEqual(entries);
+	});
+});
+
+describe("StockMap '.aggregage' method", () => {
+	test("should aggregate the quantity by [isbm, warehouse] of entries passed as argument", () => {
```
fadwamahmoud, line 76: tiny typo

PR #268, opened 2023-07-04
Add bookForm to all views and smooth up animation

### LINKED ISSUE #256: Some book form related issues
## Bugs:
- [x] The `Edit` button on `/stock` page is not working (edit functionality is not wired in)
- [x] There's no `Edit` button on outbound note page (should probably be there)
- [x] When updating a book (transaction), using the book form, on `/inbound` page, the quantity increments each time
- [x] When updating a book, using the book form, on `/inbound` page, the updating of publisher is flaky (mostly it doesn't work, one time it updated to "P", instead of "Penguin (...) Publishing")

## Touch-ups:
- [ ] I would prefer for the slide-over to slide over from right to left, rather than top to bottom
- [ ] The entry animation is janky

fadwamahmoud:
Fixes #256

### DIFF
--- a/pkg/ui/src/Slideover/Slideover.svelte
+++ b/pkg/ui/src/Slideover/Slideover.svelte
@@ -41,6 +41,6 @@
 	}
 
 	.animate-unknown {
-		animation: slider 1s ease-in-out;
+		animation: slider 300ms ease-in-out;
 	}
 </style>
--- a/pkg/ui/src/Slideover/Slideover.svelte
+++ b/pkg/ui/src/Slideover/Slideover.svelte
@@ -2,12 +2,13 @@
 	import { X } from "lucide-svelte";
 	import type { MouseEventHandler } from "svelte/elements";
 
+	import { slide } from "svelte/transition";
 	export let title: string;
 	export let description: string;
 	export let handleClose: MouseEventHandler<HTMLButtonElement>;
 </script>
 
-<div class="animate-unknown flex max-w-2xl flex-col shadow-lg">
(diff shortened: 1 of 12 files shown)


pkg/ui/src/Slideover/Slideover.svelte
```diff
@@ -28,3 +27,20 @@
...
+			opacity: 0;
+		}
+		to {
+			transform: translateX(0px);
+			opacity: 1;
+		}
+	}
+
+	.animate-unknown {
+		animation: slider 1s ease-in-out;
```
silviot, line 44: Can we make this 200ms? It slows down my speedrun!

On a more serious note, I agree with [this article](https://archive.is/yCKWA#39a4) (that I found after googling five seconds) that confirms my current opinions.

I also [asked ChatGPT](https://chat.openai.com/share/c3c06a66-7f29-4014-b07a-dacbe488b76c) and it also suggests between 200 and 300 ms for "simple animations".

pkg/ui/src/Slideover/Slideover.svelte
```diff
@@ -28,3 +27,20 @@
...
+			opacity: 0;
+		}
+		to {
+			transform: translateX(0px);
+			opacity: 1;
+		}
+	}
+
+	.animate-unknown {
+		animation: slider 300ms ease-in-out;
```
silviot, line 44: Much faster, excellent! I think it should also animate on its way out. Maybe `slider` could become `slider-in` and `slider-out`.

fadwamahmoud: I implemented the easiest way to do it, because it can't be achieved with css alone as far as I know (js must be used to toggle a different class on close).

silviot (approved):

PR #270, opened 2023-07-06
Feature/set up playwright

### LINKED ISSUE #209: Add some form of e2e testing
This is fairly obvious. We could use Cypress (with which we're already familiar), or Playwright (which Svelte endorses)...I would go with the former, but would like to hear your thoughts (maybe we can investigate Playwright and see if it's worth it).

All in all, the app is really coming together at this point and we should put e2e tests in place to safeguard the functionality as the app gets more complex (or some parts are optimised for performance)

### LINKED ISSUE #271: E2E: Move Playwright to a separate package (like we have with Eisbuk and Cypress)
Self explanatory: 
- [x] Set up a new package in `apps/e2e`
- [x] Set up playwright, scaffolding and test running logic in said package
- [x] Move the e2e tests to the said package
- [x] Provide documentation:
    - running playwright scripts
    - librocco specific helpers api (`getSidebar` and so on...)

### LINKED ISSUE #272: E2E: Playwright CI workflow
Create a CI workflow for running Playwright. This might be straightforward, but might include some playing around with. We need to:
- [x] Run e2e tests in the CI
- [ ] Get some form of reporting (preferably have a UI display, like we did with Cypress)
- [ ] Get some coverage info

ikusteu:
Fixes #209 

Tests created:
- warehouse flow - warehouse page, navigation, creation and renaming
- inbound note flow - inbound note page, navigation, creation, renaming and deletion
- outbound note flow - inbound note page, navigation, creation, renaming and deletion

Additionally, I've created a special api, used as util to target certain elements more easily (as all of the views have the same dashboard structure).

For simplicity's sake, the tests (and playwright setup) are found in `apps/web-client`, but we might want to move/set up the e2e tests in a separate package (like we have in Eisbuk)

@delasous I've updated some elements to assign css `id`s or `data-<property>`. Can you take a look and see if there's a better way to do this (aria and all)

**Edit:** 

Fixes #271 
Fixes #272


ikusteu:
Merging as it introduces no breaking changes and my current work is depending on it.

PR #308, opened 2023-08-01
Setup ci-debug workflow

ikusteu:
Sets up a workflow which runs only on branches prepended with `ci-debug/*` (same branches are ignored by other workflows). 
The mentioned branches can be used to debug the CI using SSH

### DIFF
--- a/.github/workflow.templates/workflow-templates.yml
+++ b/.github/workflow.templates/workflow-templates.yml
@@ -1,21 +1,23 @@
 name: Github Actions templates
 
 "on":
-  push:
-    paths:
-      - '.github/workflow.templates/*'
-      - '.github/workflows/*'
+    push:
+        paths:
+            - ".github/workflow.templates/*"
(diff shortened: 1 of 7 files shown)


.github/workflow.templates/workflow-templates.yml
```diff
@@ -1,21 +1,23 @@
...
-    steps:
-      - uses: actions/checkout@v2
-      - name: Run script to generate Github Actions
-        run: ./.github/build-workflows.sh
-      - name: Check if there are any changes
-        run: |
-          git status
-          git diff
-          git status | grep "working tree clean" || false && echo Did you forget to run build-workflows.sh?
+    check-template-generation:
```
silviot, line 12: No need to take any action here, but this is an example of an editor automation gone slightly wrong: the file is now using 4-spaces-indentation, but that was probably not the intention. The intention was to edit the file leaving its indentation alone.

PR #325, opened 2023-08-09
Initial plugin fetching implementation

fadwamahmoud:
Implemented the logic for fetching books from the plugin when not found in database.
1. a temp mock of the plugin manager was created that will then be replaced with the real thing, it essentially returns a book after a set timeout of 300 ms
2. use mergeMap to merge both books from db and those from plugins
3. emit available books immediately (will show up in UI as books without the rest of the meta data)
4. fetch from plugin
5. when each book is fetched, it's then persisted to the db then emitted to ensure live updates of the books


pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -37,10 +37,29 @@ class Books implements BooksInterface {
 	}
 
 	async get(isbns: string[]): Promise<(BookEntry | undefined)[]> {
-		const rawBooks = await this.#db._pouch.allDocs<BookEntry>({ keys: isbns.map((isbn) => `books/${isbn}`), include_docs: true });
+		const rawBooks = await this.#db._pouch
+			.allDocs<BookEntry>({ keys: isbns.map((isbn) => `books/${isbn}`), include_docs: true })
+			.then((docs) => unwrapDocs(docs));
```
ikusteu, line 42: Just move the comment from below to above the `.then...` (as it describes that part)

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -37,10 +37,29 @@ class Books implements BooksInterface {
...
+				return;
+			}
+
+			return isbn;
+		});
+
+		const fetchedBooks = await Promise.all(fetchIsbns.map((isbn = "") => pluginManager.getBook(isbn)));
+		this.upsert(fetchedBooks);
+
+		return [...fetchedBooks, ...existingBooks];
```
ikusteu, line 62: Ok, this messes up the order and is unnecessary. What we want to do is still return the full data frame (`rawBooks` in this implementation).

The fetching of not-found books is done in the background and, when fetched, it will trigger another pass through the stream

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -37,10 +37,29 @@ class Books implements BooksInterface {
...
+		const existingBooks: BookEntry[] = [];
+		// temp fix for isbn
+		const fetchIsbns = combinedResults.map(([isbn, book]) => {
+			if (book !== undefined) {
+				existingBooks.push(book);
+				return;
+			}
+
+			return isbn;
+		});
```
ikusteu, line 57: 1. With respect to the comments below, you'll see that `existingBooks` is unnecessary
2. There's a bug: when you do `.map(() => { ... return isbn }` you're returning all isbns (both found and not found)
3. I would prefer it if we used the more functional approach: to `wrapIter().zip()` you can also chain `.filter` (filtering out the books that are known, leaving you with only ones to fetch) and `.map` to return only the isbn

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -37,10 +37,29 @@ class Books implements BooksInterface {
 	}
 
 	async get(isbns: string[]): Promise<(BookEntry | undefined)[]> {
-		const rawBooks = await this.#db._pouch.allDocs<BookEntry>({ keys: isbns.map((isbn) => `books/${isbn}`), include_docs: true });
+		const rawBooks = await this.#db._pouch
```
ikusteu, line 40: These aren't `rawBooks` anymore as the documents get unwrapped, might as well name the var `books`

pkg/db/src/implementations/version-1.2/books.ts
ikusteu: _Note: the `wrapIter` works with generators, so it doesn't matter how many "iterations" you chain together as they're iterated over (and each step applied) only when the iterable is consumed_

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -52,41 +64,17 @@ class Books implements BooksInterface {
...
-						.then((res) => {
-							debug.log(ctx, "books_stream:initial_query:result")(res);
-							return resolve(res);
-						})
-						.catch(debug.log(ctx, "books_stream:initial_query:error"));
-				})
-			);
+			const initialState = from(this.get(isbns));
 
+			// only use latest change to query allDocs
```
ikusteu, line 69: I don't see the point of this comment (probably left-over from previous implementation)

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -52,41 +64,17 @@ class Books implements BooksInterface {
...
-									return resolve(res);
-								})
-								.catch(debug.log(ctx, "books_stream:change_query:error"));
-						})
-					)
-				)
+				switchMap(() => from(this.get(isbns)))
 			);
 
+			// call plugin here
```
ikusteu, line 75: I don't see the point of this comment (probably left-over from previous implementation)

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -37,10 +38,21 @@ class Books implements BooksInterface {
...
+			.allDocs<BookEntry>({ keys: isbns.map((isbn) => `books/${isbn}`), include_docs: true })
+			// The rows are returned in the same order as the supplied keys array.
+			// The row for a nonexistent document will just contain an "error" property with the value "not_found".
+			.then((docs) => unwrapDocs(docs));
+
+		const isbnsToFetch = wrapIter(isbns)
+			.zip(books)
+			.filter(([, book]) => book === undefined)
+			.map(([isbn]) => isbn)
+			.map((isbn) => pluginManager.getBook(isbn));
```
ikusteu, line 51: Let's move this last line to Promise.all (after the array has been materialised) - not that this is wrong, but it would be cleaner

ikusteu (changes requested):
Implementation is totally on point, but make sure to clean up (comments). Additionally, I've left a small comment regarding readability.

pkg/db/src/__tests__/tests.ts
```diff
@@ -1132,8 +1134,10 @@ export const booksInterface: TestFunction = async (db) => {
 		price: 39.86
 	};
 
-	const booksInterface = db.books();
+	const pluginManagerInstance = new PluginManager();
+	const booksInterface = db.books(pluginManagerInstance);
 
+	pluginManagerInstance.disable();
```
ikusteu, line 1140: This is totally unnecessary in tests and is something I've suggested as a (click through) debug measure. In tests, you can simply set the transactions into the note, without fetching book data.

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -48,8 +50,12 @@ class Books implements BooksInterface {
...
 			.map(([isbn]) => isbn);
 
-		Promise.all([...isbnsToFetch].map((isbn) => pluginManager.getBook(isbn))).then((res) => this.upsert(res));
-
+		const results = await Promise.all([...isbnsToFetch].map((isbn) => this.#pluginManagerInstance.getBook(isbn)));
+		// in case of disab;ed plugin, the results would be null
+		const filteredResults = results.filter((result): result is BookEntry => Boolean(result));
+		if (filteredResults.length) {
+			this.upsert(filteredResults);
+		}
```
ikusteu, line 58: This can all be done in a single chain (just a style preference). It **shouldn't** be `await`ed (as we've discussed). Might as well feed the filtered array into `this.upsert` (It's a noop if array is empty).

ikusteu: Additionally, upsert still uses `Promise.all`, I thought you've changed that to `bulkDocs` as agreed.

ikusteu (changes requested):

ikusteu:
I've left a couple of comments/fixes which should be applied. I now realise I'd forgotten to mention the "debugging" results:

There definitely are too many calls to `.upsert` (as the `.upsert` method still uses `Promise.all` instead of `(PouchDB).bulkDocs`). 
Once you've sorted that out, you will have noticed that there are still three calls to `.upsert` each time unknown book is streamed through the entries pipeline. Sadly, there's no convenient way around that and we'll just have to live with it (which we most certainly can as it'll provide us with book documents starting with rev `3_(...)` rather than `1_(...)` but it will stop at 3). 
The reason this happens is that when you update the transactions, three streams get a value transmitted on them (10 entries to show, stats of all entries, stock by isbn). Thinking about it...the third one shouldn't get updated so I'll look into it and will also see if I can merge the first two into one stream (since they both come from stream of entries). However, if I don't manage to figure it out in an hour, I will just drop it and live with the fact that book documents will be updated three times when created using the book fetching logic. If I don't manage to do it, we can add additional blocks/bookkeeping to the fetching logic to only perform one request per isbn at a time (thus blocking three requests and subsequent updates) until the first one is processed (either stored in the db, or resolved to not found)

apps/web-client/src/lib/stores/inventory/__tests__/table_content.test.ts
```diff
@@ -129,4 +130,65 @@ describe("tableContentStore", () => {
...
+					title: `Book with ISBN 12345`,
+					authors: "Some Author",
+					warehouseName: "not-found",
+					price: 10,
+					quantity: 1,
+					warehouseId: "v1/jazz",
+					availableWarehouses: new Map()
+				}
+			]);
+		});
```
ikusteu, line 192: What does this test?

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -1,95 +1,102 @@
...
-									resolve();
-								});
-							});
-						});
-				});
-			})
-		);
+		const bookIds = bookEntries.map((b) => `books/${b.isbn}`);
+
+		const docs = await this.#db._pouch.allDocs({ keys: bookIds, include_docs: true }).then(({ rows }) => {
```
ikusteu, line 21: No need for `include_docs: true` - we're just pulling the `_rev`s

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -1,95 +1,102 @@
...
+				// Find the corresponding doc in result
+				const match = rows.find((row) => row.id === `books/${b.isbn}`);
+				// If a match is found, return the merged object
+				if (match && match.doc) {
+					return {
+						...match.doc,
+						_rev: match.value.rev,
+						...b
+					};
+				}
```
ikusteu, line 34: You can zip this...the revs will be there in the same order as the requested keys with `undefined` where the `_rev` couldn't be found

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -1,95 +1,102 @@
...
+				// If no match is found, return the original book with _id added
+
+				return {
+					...b,
+					_id: `books/${b.isbn}`
+				};
+			});
+			return mergedArray;
+		});
+		return new Promise<void>((resolve) => this.#db._pouch.bulkDocs(docs).then(() => resolve()));
```
ikusteu, line 44: `return this.#db._pouch.bulkDocs(docs)` would suffice

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -1,95 +1,102 @@
...
+		const isbnsToFetch = wrapIter(isbns)
+			.zip(books)
+			.filter(([, book]) => book === undefined)
+			.map(([isbn]) => isbn);
+
+		Promise.all([...isbnsToFetch].map((isbn) => this.#pluginManagerInstance.getBook(isbn))).then((results) => {
+			// in case of disabled plugin, the results would be null
+			const filteredResults = results.filter((result): result is BookEntry => Boolean(result));
+
+			this.upsert(filteredResults);
```
ikusteu, line 63: This whole block can be abbreviated to `.then(this.upsert)` (the book fetcher should do the filtering internally)

ikusteu, line 63: This whole block can be abbreviated to `.then(this.upsert)` (the book fetcher should do the filtering internally)

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -1,95 +1,102 @@
...
+		});
+
+		return books;
 	}
 
 	stream(ctx: debug.DebugCtx, isbns: string[]) {
+		// this is the function that will fire whenever there's data is emitted
 		return new Observable<(BookEntry | undefined)[]>((subscriber) => {
+			// listens to changes in docs with these isbns
```
ikusteu, line 72: I'd like to think this is self-explanatory

pkg/db/src/implementations/version-1.2/books.ts
```diff
@@ -1,95 +1,102 @@
...
 	}
 }
 
-export const newBooksInterface = (db: DatabaseInterface): BooksInterface => {
-	return new Books(db);
+export const newBooksInterface = (db: DatabaseInterface, pluginManagerInstance?: PluginManager): BooksInterface => {
+	return new Books(db, pluginManagerInstance);
 };
+
+// given these functions, could you explain why there's a conflict error
```
ikusteu, line 102: ?

pkg/db/src/implementations/version-1.2/books.ts
fadwamahmoud: doing that produces an error of Type '(Response | Error)[]' is not assignable to type 'void'.ts(2322)

pkg/db/src/implementations/version-1.2/books.ts
fadwamahmoud: was just a self reminder when I was trying to debug

pkg/db/src/implementations/version-1.2/books.ts
fadwamahmoud: probably leftover from chatgpt convo

ikusteu:
Closing this as the updates are either in master, or awaiting review in different PRs

Issue #317, opened 2023-08-04
Book prices can be fractional


silviot:
The book edit form currently only allows integer prices. 10,50€ is a valid price

Issue #269, opened 2023-07-05
Build `publisherList` from the current database


silviot:
Currently we hardcode a list of publishers in our code to drive the autocompletion in the create/edit book form. We should instead populate it from the publishers in the current database. It will start off as an empty list.

ikusteu:
I would build on top of this and say we need to enable adding publishers first (as current select element doesn't only autofill, but doesn't typing something that's not in the list)

Tasks I see here:
- [ ] Update publisher selection field to serve as autofill, not an exclusive selection
- [ ] Create db logic (design document or a query stream) used to retrieve the available publishers
- [ ] Update book form to accept publishers as a prop + load publishers from the db and pass them to the book form (from the page component, when opening the form)

ikusteu:
Fixed in #328

PR #326, opened 2023-08-10
Fix/book form

### LINKED ISSUE #286: Book form bugs/desired updates
The book form is buggy at best, with some additional (fine grained) functionality I'd like added. Current behaviour
- ✅ when the book form is submitted it adds the transaction with quantity `1`
- ❌ when the book added already exists (loads data for the isbn provided), no quantity is incremented (even if it's the first transaction of that isbn in a particular note)
- ✅ when book is edited using `Edit` button, the quantity is not incremented

_I've noted all three cases as, with the current implementation, fixing 2nd point would mess up 3rd point (we already had a round of fixes doing the opposite)_

The desired functionality would be the following:
- when the book form is opened (by clicking `Create` button), and filled out (for not-yet-exsiting isbn), it should create a book entry and add the transaction with quantity of `1` to the note
- when the book for is opened (by clicking `Create` button) and the isbn exists it should behave as it does (prompt the user to load/edit the existing data) and still add quantity of `1` to the given note
- when the book is opened for editing (by clicking `Edit` button on a transaction), it should not increment the quantity (merely save the updated book data)

ikusteu:
Fixes #286 

I've tried implementing the behaviour specified in the ticket, but after some time spent trying to implement it without leaving room for possible errors/desired behaviour, I've come to a conclusion that the easiest and most error resistant solution would be to remove `Create` functionality altogether:
- removed the create mode in the book form
- removed the create button in the ScanInput component
- now the add book flow is the following:
  1. Use the input field to add a transaction (isbn, quantity: 1)
  2. The transaction row can be used to edit the quantity as well as click `Edit` button to edit the book data, whilst stock view (without supporting the direct quantity update) also supports `Edit` button and updating the book data
- isbn field is disabled in the book form as each time we open the form, it will be in edit mode - some book data (and required isbn) will be provided

The reasoning for removing the `Create` button is:
- it will probably seldom be used
- create can overwrite the existing data in the book store
- we can fetch the book data in the background on each isbn change, but I feel it's an overkill and would be quite awkward to work with the form that way

@delasous has raised some concerns in terms of supporting a functionality for a private person, not having access to isbn, still being able to add books to their db, but I think this opens a huge can of worms, in which case we would need to redesign the app flows a bit and it would be a bigger effort, including multiple updates other than the book form. Lastly in case we decide to go down that way, we could think about it then and update the functionality of the form.

The fixes applied:
- the book form has nothing to do with the quantity
- all actions are moved outside the form (updating the book data and so)
- the form updates the fields when `book` prop is updated (even though that probably won't happen with current use case)
- some weird validations have been removed, allowing for tests to work (required field validations are still there)
- tests (unit and e2e) are updated to test the behaviour
- these updates now allow for updates and tests of `publisherList` element (with respect to #269) as well as `price` element (with respect to #317)

@silviot we'd like your input. I would like to merge this asap as it does provide some fixes and deterministic, tested behaviour, but I will not merge it until I have everyone's approval on it (@fadwamahmoud).


delasous (approved):
LGTM

Issue #238, opened 2023-05-24
Feature: enhance single remote config / ux


delasous:
- There are a few options for replication: `to`, `from`, `bi-directional "sync"`, but to get started I don’t think we should handle all of these, and I think we should only handle a single “default” replication config. Meaning:
    1. Bi-directional sync
    2. Live sync with either `retry = false` OR `retry = true`
        1. Note: I was initially leaning towards default being retry = false, but by the time I got to the end of these notes, I was open to the alternative as I think it can make for a richer reporting of a remote’s “status”. 
            1. To do so: we would have to update so that the network error thrown in `.on("error")` is used to update the status, rather than cancelling the replication handler
    - This means that:
        - for “one time” replication ⇒ a user will have to let live run until it is complete, then cancel it.
        - instead of “auto retry” ⇒ a user needs to explicitly re-start replication from settings screen
- Config / UX:
    - A user should enter a new remote URL and save it. Once it is saved, replication should start.
    - The URL should not be open for editing once it is set. Editing the URL means interrupting (stopping) the replication and starting it again on save
    - Replication events / status of the remote connection should be communicated to the user:
        - Via:
            1. Toasts that notify the user of important changes as they happen
            2. Badges / controls next to remote settings to notify the user of the current state of the connection and actions they can perform to affect it.
        - Status should stick close to Pouch api:
            1. “Synced” - Live replication is UTD, waiting for more changes…
                1. `.on("pause", (err) => {})` err will indicate UTD if `live | retry: true`, or if “one time” it will be notified in `.on("complete", (info) => {}`
                2. `.on("active", () => {})` will fire after “pause”, either after new changes have happened, or user is back online. Could be used to reset to “In progress”
            2. “In progress” - Currently doing its thing…
                1. `.on("change", (info) => {})` info indicates progress
                2. This can be upgraded to progress bar as an enhancement. Progress bar could also be held in suspended toast
            3. “Error” - There was a problem…
                1. From docs: “If `retry` is `false`, this will also fire when the user goes offline or another network error occurs (so you can handle retries yourself, if you want)” ⇒ It should be cancelled and re-started.
                2. If we are defaulting replication settings to `retry: false` , we could also add “Offline” - User is offline, waiting to reconnect…
                    1. `.on("pause", (err) => {})` err indicates disconnected 
    - Couch replication UI also communicates “Start Time” - is this useful to communicate here?
    - Actions - a user should be able to:
        - Cancel replication
        - Retry replication.
[...]

PR #334, opened 2023-08-16
Feature/remote ux

delasous:


ikusteu:
One note for everyone looking at this, if you have a remote config in your local storage (`librocco:couchdb-config`), from the previous implementation (currently in `main`) you should remove it from local storage in order for the current implementation to use it (slightly different structure).
Otherwise you get an error looking something like `local._pouch[direction] is not callable`

apps/web-client/src/routes/settings/+page.svelte
```diff
@@ -1,60 +1,52 @@
...
-					</div>
-				</div>
-			</form>
+						<div slot="info" class="flex flex-col gap-y-2 pt-2">
+							{#if $status.info}
+								<p class="text-xs font-medium uppercase leading-4 text-gray-500">{$status.info}</p>
+							{:else}
+								<ProgressBar value={$progress.progress !== -1 ? $progress.progress : undefined} />
+								<p class="text-xs font-medium uppercase leading-4 text-gray-500">
+									{$progress.docsWritten}{$progress.docsPending ? ` / ${$progress.docsPending}` : ""} Documents synced
```
ikusteu, line 41: Did you mean `docsWritten/(docsWritten + docsPending)` or was this in intentional?

apps/web-client/src/lib/stores/replication.ts
```diff
@@ -0,0 +1,182 @@
+import { readable, writable, derived, get } from "svelte/store";
+
+import type { DatabaseInterface } from "@librocco/db";
+
+import type PouchDB from "pouchdb";
+
+const BATCH_SIZE = 5;
```
ikusteu, line 7: Removing this and using the default batch size sped up things considerably for me. 

I understand you were playing with this to allow for usage in between writes. This is just an observation (also considering the fact that you had planned to halt the app usage until initial sync is done).

ikusteu: However...it just came to mind: If we plan to use multiple live syncs (or one-timers), we will probably want to find a fine line in-between in order to service both syncs.

ikusteu:
I've (intentionally) not started the docker container before starting the app, which, as is expected, showed an error. 
However, after starting the containers, I've expected the connection to retry and connect, which it didn't do. After refresh it worked.

ikusteu:
After clicking "cancel and edit" it would be more convenient if the previous value is there for edit (and the user doesn't have to type in the whole thing)

ikusteu:
I've tried to cancel and edit and restart the connection. I got an error describing that the connection can't be established. After manually deleting `librocco:couchdb-config` and restarting, it works again (when set by the initial, onMount action).

Now, I might be wrong and missing something, but both configs seem the same to me.

**Manual:**
Screenshot 2023-08-21 at 23 34 58

**After restart** (set programmatically on mount):
Screenshot 2023-08-21 at 23 34 11

apps/web-client/src/routes/settings/+page.svelte
delasous: Unintentional. Have updated

apps/web-client/src/lib/stores/replication.ts
delasous: I have removed this for now. But on my computer it does make the app very laggy while replication is happening with our large test couchdb

delasous:
> I've (intentionally) not started the docker container before starting the app, which, as is expected, showed an error. However, after starting the containers, I've expected the connection to retry and connect, which it didn't do. After refresh it worked.

> After clicking "cancel and edit" it would be more convenient if the previous value is there for edit (and the user doesn't have to type in the whole thing)

> I've tried to cancel and edit and restart the connection. I got an error describing that the connection can't be established. After manually deleting librocco:couchdb-config and restarting, it works again (when set by the initial, onMount action)...

The above issues should be resolved

apps/web-client/src/lib/toasts.ts
```diff
@@ -34,3 +44,24 @@ export const warehouseToastMessages = (warehouseName) => ({
 	warehouseCreated: `${warehouseName} created`,
 	bookDataUpdated: (isbn: string) => `Updated book data for '${isbn}'`
 });
+
+// Aliging with BadgeColor enum here as it this color+message combo is used in the RemoteDb Description List
+export const replicationStatusMessages = {
+	INIT: { color: BadgeColor.Success, message: "Connecting to remote database" },
+	"ACTIVE:REPLICATING": { color: BadgeColor.Success, message: "Syncing with database" },
+	"ACTIVE:INDEXING": { color: BadgeColor.Success, message: "Building indexes" },
```
silviot, line 52: Pet peeve of mine: I higly prefer the spelling "indices".

Me:
> Is the plural of index indexes or indices?
ChatGPT:
> Both "indexes" and "indices" are accepted plurals of "index." "Indexes" is more common in general usage, while "indices" is often used in mathematical and technical contexts.

silviot:
I think the sync status indication should be in the top bar. Ideally it should be in the top left logo: it should indicate progress with some coloring and show more details on hover/click. For now I believe putting the sync info on the top right would be an improvement. Let's discuss this!

I also experienced the jankyness. I could mitigate it closing the window and opening it again. I wonder if it would make sense to "work every other second" to make the app feel usable while it's syncing. It would also be less stressing on the system it runs on.

delasous:
Fixes #238

silviot (approved):

PR #341, opened 2023-08-22
Feature/plugin work setup

### LINKED ISSUE #340: Set up the `book-data-extension` plugin package
Create and scaffold a new package in `pkg/plugins/book-data-extension`:
- should have access to the window when testing (so setup similar to that of `pkg/ui` can be used)
- won't be using Svelte (nor Tailwind, PostCSS, etc...), so all of the related plugins should be removed (eslint svelte and tailwind, prettier, etc.)

ikusteu:
Fixes #340


fadwamahmoud (approved):

ikusteu:
Merging this in as other work depends on it and the failing e2e shouldn't be affected with this patch - different bug

PR #346, opened 2023-08-24
Fix/navigation UI

### LINKED ISSUE #265: Some details around navigation
I thought this would be a quick fix, but am opening a ticket as it will require some playing around and picking the right behaviour (I think the ticket itself would be a small amount of work nonetheless, but more than a 5min fix and move on):

- [x] When we navigate to a note and then refresh, the current note (in the side bar nav) is not highlighted anymore, my reasoning is that (even though to the eye it looks the same), the trailing slash is added to route `id` param - this can easily be remedied by adding a reactive variable `$: currentId = $page.params.id.replace(/\/$/, "")` and use that one for comparison
- [x] The same, but a bit heavier in terms of how we would achieve this: It would be cool if, in `/inbound` view, the warehouse which the note belongs to could be expanded after refresh
- [x] When warehouse view is refreshed (regardless of the current warehouse), we get redirected to `/0-all` - is this the desired behaviour
- [x] Make the default warehouse `v1/0-all` (both `v1/0-all` and `0-all` will work, but versioning the default ensures we can show the default warehouse as current warehouse in side bar nav (as mentioned above) as the warehouse `id` in the side bar nav (used to match the warehouse from `id` route param) is already versioned

~~### Edit: Additinal work~~
~~- [ ] Redirects don't work at all when the app is built - all redirect to default stock view on reset~~ -> see #347

ikusteu:
Fixes #265

### DIFF
--- a/apps/e2e/helpers/sidebar.ts
+++ b/apps/e2e/helpers/sidebar.ts
@@ -43,7 +43,7 @@ function getSideLinkGroup(sidebar: Locator, name: string): SideLinkGroupInterfac
 		return container.getByRole("link", { name: label, exact: true });
 	};
 
-	const { open } = useExpandButton(container);
+	const { open, getExpandButton } = useExpandButton(container);
 
 	const createNote = async (opts?: WaitForOpts) => {
 		await open(opts);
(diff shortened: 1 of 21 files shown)


ikusteu:
Merging this in as it's been open for awhile, all the checks are passing and it doesn't introduce any breaking changes

PR #348, opened 2023-08-28
Implement plugin

### LINKED ISSUE #342: Implement the book data extension plugin
As mentioned in the discussion we need to create the book fetching plugin which implements the bookFetcher interface.
The fetchBooks method is pretty much the same as the one in the librocco-extension content script except it will accept an isbns array.

What the fetchBooks method does:
 - post message to the window
 - waits for response
 - filters out empty results
 - returns an array of book results
 
 ### Tasks:

- [x] create `book-fetcher` plugin inside the plugin package
- [x] write tests (in the DOM) that send a message and expect a response of `bookEntries` (doesn't have to be equal length of isbns provided)
- [x] enhance functionality by providing a timeout for responses 
- [x] Add a check/flag that the extension is actually registered in the browser by sending a ping from the client app to the extension as soon as it's registered, if that flag is false then an empty array is returned from `fetchBooks`

fadwamahmoud:
Fixes #342


plugins/book-data-extension/src/plugin/__tests__/listeners.test.ts
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+describe("book fetching listeners", () => {
+	test("listenForExtension resolves to true when it receives response to ping", async () => {
+		vi.spyOn(helpers, "addEventListener").mockImplementation(mockAddEventListenerExtension);
+
+		helpers.postMessage(`BOOK_FETCHER_REQ:replyFromExtension:`);
+
+		const ww = await listenForExtension("BOOK_FETCHER_REQ:replyFromExtension:", 1000);
```
ikusteu, line 51: What does `ww` stand for?

plugins/book-data-extension/src/plugin/__tests__/listeners.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+		expect(ww).toEqual(true);
+	});
+	test("listenForBooks resolves to books when it receives response from extension", async () => {
```
ikusteu, line 55: Can you please use empty rows in a sensible manner

plugins/book-data-extension/src/plugin/index.ts
```diff
@@ -1,2 +1,49 @@
...
-	"This is a dummy string so that the 'build' script doesn't fail (warn = fail in CI) due to empty bundle. Remove when there's something to build.";
+import { BookFetcherPlugin as BookFetcherPluginType, BookEntry } from "@librocco/db";
+import { postMessage, addEventListener, removeEventListener } from "./window-helpers";
+import { listenForBooks } from "./listeners";
+
+export const BookFetcherPlugin = (): BookFetcherPluginType => {
+	const fetchBookData = async (isbns: string[]): Promise<BookEntry[]> => {
+		// Check if the extension is registered if not resolve to []
+
+		postMessage(`BOOK_FETCHER_REQ:checkForExtension:${isbns}`);
```
ikusteu, line 9: It makes sense to make the comment closer to the line it's referring to than the line before it, doesn't it?

ikusteu, line 9: 1. Can we use a consistent casing/format in messages?
2. Can we use shorter messages? 
3. Why send isbns when checking that the extension is available?

In this case `BOOK_FETCHER:PING` would suffice

plugins/book-data-extension/src/plugin/index.ts
```diff
@@ -1,2 +1,49 @@
...
+				if (!resolved) {
+					clearTimeout(promiseTimer);
+					resolve(false);
+				}
+			};
+			const promiseTimer = setTimeout(promiseHandler, 1000);
+
+			const handler = (event: MessageEvent) => {
+				if (event.source !== window) return;
+				if (event.data.message && event.data.message === `BOOK_FETCHER_REQ:replyFromExtension:`) {
```
ikusteu, line 26: Wouldn't it make more sense to make this `BOOK_FETCHER:RES` (all intention is conveyed in a much shorter and consistently formatted string)

ikusteu (changes requested):
Try this:

- Create a function to handle timeouts:
```ts
// In utils.ts

// starts timeout
// runs callback
// if successful clears timeout and resolves to appropriate value
// if timed out, resolves to undefined
function runWithTimeout<R>(cb: () => R, timeout: number): Promise<R | undefined>
  }
  ```
- create a ping function:
```ts
// Runs with timeout (as short one of, like 200ms - we're pinging a local service worker in the same browser:
// - sends 'BOOK_EXTENSION:PING'
// - listens for 'BOOK_EXTENSION:PONG'
// - on success -> resolves to true
// - timeout -> resolves to false
function pingExtension(): Promise<boolean>
```
- create an internal `fetchBook` function:
```ts
// Fetches one book by isbn
// Runs with timeout (a little longer than ping as it makes a 3rd party request, but not as long as 1000ms)
// Posts a message `BOOK_FETCHER:REQ:{isbn}`
// Listens to response message `BOOK_FETCHER:RES` (+ book data)
// Resolves to book data or undefined
function fetchBook(isbn: string): Promise<BookEntry | undefined>
```
- finally, bring all the pieces together
```ts
// Accepts isbns
// Checks if extension is present
// Does Promise.all for all of the isbns
// After results come back, filters out undefined (leaving us with clean 'BookEntry[]' - possibly empty)
// Resolves
function fetchBooks(isbns: string): Promise<BookEntry[]>
```

plugins/book-data-extension/src/plugin/index.ts
```diff
@@ -1,2 +1,28 @@
-export const dunnystring =
-	"This is a dummy string so that the 'build' script doesn't fail (warn = fail in CI) due to empty bundle. Remove when there's something to build.";
+import { BookFetcherPlugin as BookFetcherPluginType, BookEntry } from "@librocco/db";
+import { postMessage } from "./window-helpers";
+import { listenForBook, listenForExtension } from "./listeners";
+
+export const BookFetcherPlugin = (): BookFetcherPluginType => {
```
ikusteu, line 5: Maybe name this `createBookDataExtensionPlugin`. It's not a convention to name functions in PascalCase, the suggested name is more indicative and the name clashing (which you've solved with aliasing here) would also be present when consuming the plugin in the app.

plugins/book-data-extension/src/plugin/index.ts
```diff
@@ -1,2 +1,28 @@
...
+	const fetchBookData = async (isbns: string[]): Promise<BookEntry[]> => {
+		// Check if the extension is registered if not resolve to []
+		postMessage(`BOOK_FETCHER:PING`);
+
+		const extensionAvailable = await listenForExtension(`BOOK_FETCHER:PONG`, 500);
+
+		if (!extensionAvailable) return new Promise((resolve) => resolve([]));
```
ikusteu, line 12: Inside `async` function, you can just return `[]` and it will be of type `Promise<[]>`

plugins/book-data-extension/src/plugin/index.ts
```diff
@@ -1,2 +1,28 @@
...
+
+		return unfilteredBookData.filter((bookData): bookData is BookEntry => bookData !== undefined);
+	};
+
+	const fetchBook = (isbn: string) => {
+		// Post message to the extension
+		postMessage(`BOOK_FETCHER:REQ:${isbn}`);
+
+		// Listen for a response and resolve to [] after timeout
+		return listenForBook(`BOOK_FETCHER:RES`, 800);
```
ikusteu, line 24: Shouldn't this return `BookEntry | undefined`, not `never[]`

plugins/book-data-extension/src/plugin/listeners.ts
```diff
@@ -0,0 +1,33 @@
+import { BookEntry } from "@librocco/db";
+import { addEventListener } from "./window-helpers";
+export function listenForExtension(message: string, timeout: number): Promise<boolean> {
+	return new Promise((resolve) => {
+		const promiseTimer = setTimeout(() => resolve(false), timeout);
```
ikusteu, line 5: listener should be unregistered here, in case of timeout, as well

plugins/book-data-extension/src/plugin/index.ts
ikusteu: A ok...from looking at the implementation, it is implemented correctly, only the comment is a bit misleading

plugins/book-data-extension/src/plugin/listeners.ts
```diff
@@ -0,0 +1,33 @@
...
+				resolve(true);
+			}
+		};
+		addEventListener("message", handler);
+	});
+}
+
+export function listenForBook(message: string, timeout: number): Promise<BookEntry | undefined> {
+	return new Promise((resolve) => {
+		const promiseTimer = setTimeout(() => resolve(undefined), timeout);
```
ikusteu, line 21: listener should be unregistered here, in case of timeout, as well

ikusteu (changes requested):
Some small stuff (in comments) + I see two missing test cases:
1 . When we request multiple books and one of those resolves to `undefined`, it should return a list of only the books which were found
2. When we request multiple books and one times out, same as above :point_up:

ikusteu (approved):

PR #350, opened 2023-09-01
Feature/receipts interface 1

ikusteu:
In sync with [this PR in the receipt printer server repo](https://github.com/librocco/receipt-printer/pull/2)

Creates a receipts interface, used to print the receipt for the note (and will probably be used to retrieve the receipts in the future). Current implementation (first pass):
- constructs a print job for receipt printing
- stores the print job in the db, under `v1/print_queue/<printer_id>/<job_id>` - the `printer_id` is hardcoded to `"printer-1"` and jobs are timestamped unique strings (so they appear in order, id generated using `timestampId` - the same helper we use for note/warehouse ids)
- adds a `"Print receipt"` button to note view to explicitly initialise the print job (in the future this will probably be removed, receipts printed automatically, or explicitly from receipts UI, TBD)

### DIFF
--- a/apps/web-client/src/routes/inventory/inbound/[...id]/+page.svelte
+++ b/apps/web-client/src/routes/inventory/inbound/[...id]/+page.svelte
@@ -24,7 +24,9 @@
 		type RemoveTransactionsDetail,
 		ProgressBar,
 		BookDetailForm,
-		Slideover
+		Slideover,
+		Button,
+		ButtonColor
 	} from "@librocco/ui";
 	import type { BookEntry, NavMap } from "@librocco/db";
 
@@ -203,14 +205,17 @@
 							</div>
 						{/if}
 					</div>
-					<SelectMenu
-						id="note-state-picker"
-						class="w-[138px]"
-						options={noteStates}
-						bind:value={$state}
-						disabled={[...Object.values(NoteTempState), NoteState.Committed].includes($state)}
-						align="right"
-					/>
+					<div class="flex gap-x-4">
+						<Button color={ButtonColor.White} on:click={() => note?.printReceipt()}>Print receipt</Button>
+						<SelectMenu
+							id="note-state-picker"
+							class="w-[138px]"
+							options={noteStates}
+							bind:value={$state}
+							disabled={[...Object.values(NoteTempState), NoteState.Committed].includes($state)}
+							align="right"
+						/>
+					</div>
 				</div>
 				<ScanInput onAdd={handleAddTransaction} />
 			{/if}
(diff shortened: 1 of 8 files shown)


fadwamahmoud (approved):

silviot:
I tried to run this but I got:

```
proxy.js?v=7d601a0a:15 [HMR][Svelte] Unrecoverable HMR error in <Root>: next update will trigger a full reload
logError @ proxy.js?v=7d601a0a:15
10:59:45.695 pouchdb.js?v=7d601a0a:9901 Uncaught (in promise) TypeError: Cannot create property 'PouchConstructor' on string 'http://admin:admin@127.0.0.1:5000/dev'
    at Function.sync (pouchdb.js?v=7d601a0a:9901:25)
    at PouchDB2.sync (pouchdb.js?v=7d601a0a:10088:29)
    at Gn.sync (index.es.js:2103:30)
    at +layout.svelte:591:3
    at run (chunk-5HU7U5RE.js?v=7d601a0a:19:10)
    at Array.map (<anonymous>)
    at chunk-5HU7U5RE.js?v=7d601a0a:1806:52
    at flush (chunk-5HU7U5RE.js?v=7d601a0a:1092:9)
    at init (chunk-5HU7U5RE.js?v=7d601a0a:1888:5)
    at new Root (root.svelte:996:3)
```
using an incognito window.

ikusteu:
@silviot I wasn't able to reproduce this. I know it's stupid, but did you try running:
- `rush update && rush build` 
- or, if it persists `rm -rf ./**/node_modules && rush update --purge && rush build`
- or, I sometimes needed to do this (very rarely) `rm -rf ./**/.svelte-kit && rush update && rush build` (technically, you would run `rush postinstall`, but it's ran after (as part of the install process) `rush update`

This PR includes recent bumps to PouchDB, Rush and PNPM (which are all in master) and this seems to be an issue with PouchDB

silviot:
@ikusteu I definitely forgot to run `rush build`. I now did, and it's replicating. I think we should bundle two databases with the couchdb docker image: a `small` one and the current `dev` one.

ikusteu:
@silviot ok…that doewn’t sound right. You had a PouchDB issue. PouchDB is (or should be) externalised when bundling the db package and, when running web-client ‘start’ wec-client’s PouchDB should be used (for which only rush update should have sufficed).

This let’s me think we’re not externalising PouchDB in the db package after all…I’ll investigate

silviot:
@ikusteu In case it helps FYI the branch I was switching from (and I do remember I ran `rush build` there) was `feature/remote-ux`.

PR #366, opened 2023-12-11
Feature/web usb printer

ikusteu:
Includes #365 

- Creates a receipt printer package (a copy of node-escpos refactored to work in browser environment).
- Updates db interface to expose the receipt data (instead of printing)
- Wires in the printing functionality in the web client app


fadwamahmoud:
I'm getting this error when I try and print and I checked the site permissions and usb devices was set to `Ask`
image

PR #376, opened 2023-12-15
Wire in the scanning functionality on inbound/outbound note views:

ikusteu:
* Copy + update the scan action from 'ScanInput' component (update to allow 'Enter' submit)
* Use the action on "naked" input elements in new inbound/outbound views (without wrapping into form for easy and consistent styling)

Closes T-283


delasous (approved):
Love this

PR #377, opened 2023-12-15
Fix/small fixes 6

ikusteu:
Closes T-271
Closes T-284
Closes T-286
Closes T-295

@delasous what are your thoughts on the entity list row (and button) classes. I'm not sure how I feel about them: 
- I think button classes are the way to go (using naked buttons and component classes)
- entity list - we partly lose a single source of truth (we still have shared classes), but we gain versatility (no need to worry about updated at badge not being displayed in warehouse list + we add warehouse discount)


delasous (approved):

delasous:
With Buttons I'm totally onboard. Chiefly because we often need to style buttons and links in the same way and this saves us doing imperfect workarounds with `svelte:element` and generic props.

EntityList I'm less sure about. The "updated at" and "warehouse discount" content could be managed as a slot. But my real hesitation is around introducing and managing too many component classes. I think we need a better criteria for defining when the shift from full component to raw html + classes should be made.

I had some thoughts on how we could use Uno CSS to write and organise component class layers (which would still work with all of the tailwind class definitions). I will outline this and some thoughts on the criteria in a discussion shortly...

delasous:
I think this should still go in as is

PR #378, opened 2023-12-20
Feature/table rehaul

delasous:
Closes T-280

Rebuilds table layout: should be usable down to ~400px (collapsing page sidebar at `md` would be final touch to improve this); and manages any interactive cell components as slots.

There's more to be done here to reconcile these updates with existing code. For now the focus was on adding the new working tables to `/proto` views


ikusteu (approved):
All in all LGTM for initial pass, some small things I've found:
- I would prefer we move towards ditching the UI package altogether (and rebuild when and if we need to share components between the inventory and ordering apps), but I figure you've used the ui package for histoire ??
- The left padding on tables seems a bit tight
- I got this when hovering over the buttons on note table views (both in and out):
Screenshot 2023-12-20 at 13 37 08

If you hover over and back (left) the hover box disappears, but if you move from button to button (row by row) they're just left like this

delasous:
- Yeah I've set this up in the UI package for the time being because of Histoire. We should mirror the setup in the app package and copy over the surviving components
- I've been as conservative with padding as possible. I spent more than a few hours tweaking this and don't have an appetite to touch it anymore. I am open to any adjustments that you feel improve usability
- I have removed the odd behaviour with the actions row hanging open. Tried something with open on `mouseenter` but it was hard to get close on `mouseleave` to work properly. Better to just keep it simple

PR #379, opened 2023-12-22
Feat/confirm action dialogs

delasous:
Closes T-262
Closes T-261

Includes warehouse form

### DIFF
--- a/apps/web-client/src/lib/forms/WarehouseForm.svelte
+++ b/apps/web-client/src/lib/forms/WarehouseForm.svelte
@@ -1,0 +1,53 @@
+<script lang="ts">
+	import type { ZodValidation, SuperValidated } from "sveltekit-superforms";
+	import type { SuperForm } from "sveltekit-superforms/client";
+	import { superForm, superValidateSync, numberProxy } from "sveltekit-superforms/client";
+
+	import { warehouseSchema, type WarehouseFormData } from "$lib/forms/schemas";
+
+	import { Input } from "$lib/components/FormControls";
+	import { Percent } from "lucide-svelte";
+
+	type Form = SuperForm<ZodValidation<typeof warehouseSchema>, unknown>;
+
+	export let data: WarehouseFormData;
+	export let options: Form["options"];
+	/**
+	 * Handle click of "X" icon button
+	 */
+	export let onCancel: (e: Event) => void = () => {};
+
+	const form = superForm(superValidateSync(data, warehouseSchema), options);
+
+	const { form: formStore, constraints, enhance } = form;
+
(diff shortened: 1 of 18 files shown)


apps/web-client/src/lib/forms/WarehouseForm.svelte
```diff
@@ -0,0 +1,53 @@
...
+	import { Percent } from "lucide-svelte";
+
+	type Form = SuperForm<ZodValidation<typeof warehouseSchema>, unknown>;
+
+	export let data: WarehouseFormData;
+	export let options: Form["options"];
+	/**
+	 * Handle click of "X" icon button
+	 */
+	export let onCancel: (e: Event) => void = () => {};
```
ikusteu, line 18: Why not dispatch `cancel` event on the component?

apps/web-client/src/lib/forms/WarehouseForm.svelte
```diff
@@ -0,0 +1,53 @@
...
+
+	const discountProxy = numberProxy(formStore, "discount", { emptyIfZero: false, empty: "undefined" });
+</script>
+
+<form class="flex max-w-lg flex-col gap-y-6" aria-label="Edit warehouse details" use:enhance method="POST">
+	<div class="flex flex-col justify-between gap-y-6 p-6">
+		<input type="hidden" name="id" value={$formStore.id} />
+		<div class="basis-full">
+			<Input bind:value={$formStore.name} name="name" label="Name" placeholder="Dish " {...$constraints.name} />
+		</div>
```
ikusteu, line 32: This is cool if `bind:value` works, but I'd really prefere the naked `<input>` element with style classes, rather than a full on component (if possible ofc)

ikusteu: If I remember correctly, the two-way bind was a problem with multiple choice checkboxes, so...since we don't have those here, not really necessary (just a preference)

ikusteu (approved):
Left a couple of soft comments. All in all LGTM, nice to see it in action.

delasous:
I think it's worth chatting about and agreeing on a standard approach for the scenario's you've highlighted above. For now, some quick comments:
- For the cancel handler vs custom event: they both seem like good options. I've sided with the handler because typing custom events seemed more verbose, but I think they've improved automatic detection of this...
- For the form control / input bindings I have mixed feelings about both of our approaches - I worked through them extensively while rebuilding a big form in Edible's builder recently and have my complaints on either side. I think there's a third path which involves leaning more into Superform's interface. 

I will outline and we can discuss these things along with the CSS approach that I mentioned a couple weeks back

PR #381, opened 2023-12-28
Chore/update node 20

ikusteu:
Closes T-303
@delasous pls. check the inline deps thing - I've removed it to make Vitest work, but I don't really know what it does and if removing it is a correct way to set things up


delasous (approved):

delasous:
From my basic understanding of `inline deps`, I don't think we need to worry about removing it

PR #387, opened 2024-01-30
Fix/scan action

### LINKED ISSUE #294: Book form submission by pressing "Enter" doesn't work on webkit browsers
This might be a fluke in the tests, but Playwright tests fail on webkit browsers due to book form not closing (nor being submitted apparently) on "Enter" press.
We've fixed this by always submitting the form using the 'Save' button click, but it would be good to make sure it works in both cases (keyboard and click submission)

delasous:
Closes #294 
Closes T-307
Closes T-170

- Wraps the scan input in a form element to handle form submission and form field reset.
- Updates scan action to increase buffer length threshold and to return early if the field is already focussed
- Adds a few tests for the action
- Fixes a bug on Outbound form where EditBookForm modal wouldn't stay open

I have branched this from #385


silviot (approved):
The problem is now fixed!
I didn't look into the code.

PR #390, opened 2024-02-02
Feat/order dialog

delasous:
Closes T-328

I simplified the design of the order confirmation dialog to bring it in line with the existing dialog. This PR just makes a few small tweaks to the existing component setup/use and adds a story.


pkg/shared/package.json
```diff
@@ -4,9 +4,14 @@
...
-			"require": "./dist/index.js",
-			"default": "./dist.index.es.js"
+			"import": {
+				"default": "./dist/index.es.js",
+				"types": "./dist/index.d.ts"
+			}, 
+			"require": {
+				"default": "./dist/index.js",
+				"types": "./dist/index.d.ts"
+			}
```
ikusteu, line 14: 👌

ikusteu (dismissed):
LGTM

ikusteu (changes requested):
Committing an inbound note, I get this in the dialog (same happens for an outbound note):
Screenshot 2024-02-06 at 10 15 43

Editing the warehouse: 
- clicking 'Save' works as expected
- clicking 'Cancel' acts the same way as 'Save' (updates are applied and changes stored to the db)

delasous:
@ikusteu I fixed the title and description issue, but I'm not seeing the "Cancel submit" behaviour that you're describing. Can you confirm if it's still there?

PR #391, opened 2024-02-02
Enhancement: set pouchDB batch size to 2000

laranotreallycroft:
Batch size is increased from the default 100 to 2000 which speeds up the sync to 1min~, further batch size increase does not have a significant effect

### DIFF
--- a/apps/web-client/src/lib/components/DescriptionLists/RemoteDbData.story.svelte
+++ b/apps/web-client/src/lib/components/DescriptionLists/RemoteDbData.story.svelte
@@ -12,7 +12,8 @@
 		url: "admin:admin@127.0.0.1:5000/dev",
 		direction: "sync" as const,
 		live: true,
-		retry: true
+		retry: true,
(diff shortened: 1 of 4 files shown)


ikusteu (changes requested):
I wrote some small comments, but more importantly, I ran the replication with both default batch size and `batch_size: 2000` and go these results:
- default batch size: 287928ms (roughly 4.8mins)
- `batch_size: 2000`: 316025ms (roughly 5.3mins)

Which leads to one of the following conclusions:
- depending on the architecture of the system, different setup will yield different benefits (speed-up)
- batch size is completely unrelated to the performance of the replication

Can you:
- rerun the experiment, but make sure you've cleared the db from the previous replication
- investigate what exactly the `batch_size` does and how it affects the performance (that might sort out the difference in behaviour across different system architectures)

silviot:
There has been good progress on investigating the pouchdb bug that slows down sync. The approach in this PR can be put to the side for now.

PR #392, opened 2024-02-03
Customer order db interface

fadwamahmoud:
Link T-311


pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -0,0 +1,168 @@
...
+	// 	}
+
+	#db: OrdersDatabaseInterface;
+
+	#initialized = new BehaviorSubject(false);
+	#exists = false;
+
+	draft = true;
+
+	orderId = "";
```
ikusteu, line 34: I don't think we need `orderId` - there's already an `_id` (being a couch document)

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -0,0 +1,168 @@
...
+	// {
+	// 	"id": string, // use 'timestampedId' (or 'timestampId', not sure, it's in the 'db' package, under utils I think
+	// 	"draft": boolean,
+	// 	"displayName": string, // "name" -> "displayName" to keep consistent with the rest of the app
+	// 	"email": string,
+	// 	"books": [ 
+	// 	  { isbn: string }, // No status for now we can add that later when dealing with supplier orders
+	// 	],
+	// 	"deposit"?: number // I think the deposit should be calculated on display without persisting and then added when the order is placed
+	// 	}
```
ikusteu, line 25: Let's remove all excess comments. I understand you've used them as helper (which makes total sense), but I think the code that ends up in `main` should be clean.

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -0,0 +1,168 @@
...
+	_id: VersionedString;
+	docType = DocType.CustomerOrder;
+	_rev?: string;
+	_deleted?: boolean;
+
+
+	entries: VolumeStock[] = [];
+	committed = false;
+	displayName = "";
+	updatedAt: string | null = null;
```
ikusteu, line 59: This seems to be a leftover from when copying the (note) code from the inventory db.

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -0,0 +1,168 @@
...
+		this.#db = db;
+
+		// Store the id internally:
+		// - if id is a single segment id, prepend the warehouse id and version the string
+		// - if id is a full id, assign it as is
+		this._id = !id
+			? versionId(`${uniqueTimestamp()}`)
+			: isVersioned(id, "v1") // If id is versioned, it's a full id, assign it as is
+				? id
+				: versionId(`${id}`);
```
ikusteu, line 71: This can be replaced with `versionId(id)` - as it's idempotent in terms: It won't prepend the already versioned `id`.
Example
```ts
versionId("order-1") // returns "v1/order-1"
versionId("v1/order-1") // returns "v1/order-1"
```

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -0,0 +1,168 @@
...
+		this.updateField("books", data.books);
+		this.updateField("deposit", data.deposit);
+		this.updateField("docType", data.docType);
+		this.updateField("updatedAt", data.updatedAt);
+		this.updateField("email", data.email);
+
+		this.#exists = true;
+
+		return this;
+	}
```
ikusteu, line 151: Can we make it so that the class definition follows this structure:
- declaration of private properties (properties prepended with `#`)
- declaration of data (non-private properties)
- constructor (if any)
- private methods
- public methods

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -0,0 +1,168 @@
...
+
+
+	// get(): Promise<CustomerOrderInterface | undefined> {}
+
+
+
+
+	// async commit(ctx: debug.DebugCtx, options?: { force: boolean }): Promise<CustomerOrderInterface> {}
```
ikusteu, line 159: We don't need a `commit` method on an order

pkg/db/src/implementations/version-1.2/orders-db.ts
```diff
@@ -61,6 +63,11 @@ class Database implements OrdersDatabaseInterface {
 		// We're caching the books interface to avoid creating multiple instances
 		return this.#booksInterface ?? (this.#booksInterface = newBooksInterface(this));
 	}
+
+	customerOrder(): CustomerOrderInterface {
+		return this.#customerOrderInterface ?? (this.#customerOrderInterface = newCustomerOrder(this))
+	}
```
ikusteu, line 69: Say I want to do this:

```ts
const db = new Database(...)

// get the first order
const order1 = db.customerOrder()
const order2 = ??? // How would I go about this ?
```

pkg/db/src/implementations/version-1.2/customer-orders.ts
ikusteu: Also, keep in mind that, during the development, the two dbs should be ran in isolation (orders and inventory), because: If we were to write to the same db using both interfaces, we would get interference:
- warehouse id (path): `v1/<warehouse-id>`
- customer order id (path): `v1/<order-id>`

This is fine during development, later we'll namespace the different sections, and I've opened up a ticket for that (just pointing this out to be kept in mind)

pkg/db/src/implementations/version-1.2/customer-orders.ts
ikusteu: On second thought, we might want to use it to put the order out of "draft" state. Alternatives would be `.place()`, `.placeOrder()`, `.order()`.
Whichever you prefer (including the `.commit()`)

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -62,6 +68,28 @@ class CustomerOrder implements CustomerOrderInterface {
...
 
+	private update(ctx: debug.DebugCtx, data: Partial<CustomerOrderInterface>) {
+		return runAfterCondition(async () => {
+			debug.log(ctx, "customerOrder:update")({ data });
+
+			// Committed notes cannot be updated.
+			if (!this.draft) {
+				debug.log(ctx, "customerOrder:update:customerOrder_not_draft")(this);
+				return this;
+			}
```
ikusteu, line 79: Shouldn't there be an error?
If the notes are handled in the same way as this, then ok...no error. Let's keep things consistent and, if need be, update notes as well as orders (in case we want to throw an error).
However, if notes do throw an error in this case, yeah...let's keep things consistent.

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -62,6 +68,28 @@ class CustomerOrder implements CustomerOrderInterface {
...
+
+			const updatedData = { ...this, ...data, updatedAt: new Date().toISOString() };
+			debug.log(ctx, "customerOrder:updating")({ updatedData });
+			const { rev } = await this.#db._pouch.put<CustomerOrderData>(updatedData);
+			debug.log(ctx, "customerOrder:updated")({ updatedData, rev });
+
+			// We're waiting for the first value from stream (updating local instance) before resolving the
+			// update promise.
+			await firstValueFrom(this.#updateStream);
```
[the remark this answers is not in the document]
ikusteu: But you might be onto something - I think I might have a bug that could be fixed with this approach

pkg/db/src/implementations/version-1.2/customer-orders.ts
```diff
@@ -115,11 +143,89 @@ class CustomerOrder implements CustomerOrderInterface {
...
+	 */
+	updateBookStatus(ctx: debug.DebugCtx, isbn: string, status: string): Promise<CustomerOrderInterface> {
+		const index = this.books.findIndex((book) => book.isbn === isbn);
+		if (index === -1 || this.books[index].status === status) Promise.resolve(this);
+
+		const updatedBooks = [...this.books];
+		updatedBooks[index] = { ...updatedBooks[index], status };
+
+		return this.update(ctx, { books: updatedBooks });
+	}
```
ikusteu, line 187: This is on the right track, however...let's support updating of multiple books. That way:
- we **can** update the entire batch - even if it's a subset of books ordered, it will still probably be multiple books a lot of the time
- if we need to update only one book - just to [book]

pkg/db/src/implementations/version-1.2/customer-orders.ts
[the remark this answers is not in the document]
fadwamahmoud: this is exactly how you implemented it in `notes.update()` I'm not onto something :joy:

fadwamahmoud:
Fixed comments. Typecheck is still failing but Chris has fixed this in #390. @ikusteu

pkg/db/src/types/orders.ts
```diff
@@ -1,7 +1,40 @@
...
+}
+
+export interface CustomerOrderBook { isbn: string; status: string };
+/**
+ * Standardized data that should be present in any customer order
+ * (different implementations might differ, but should extend this structure)
+ */
+export type CustomerOrderData<A extends Record<string, any> = {}> = CouchDocument<
+	{
+		updatedAt: string | null;
```
delasous, line 33: I came here to check how this compared against my mock type for order table data - just a note that I have also included `name`

PR #396, opened 2024-02-09
Feat/order table

delasous:
Closes T-327

- Adds tables for customer and supplier orders
- I decided against using the headless table libraries that I was experimenting with as they quickly get convoluted with their strategies for programmatic cell rendering. Ultimately, our tables don't require any complex features (column re-arranging, row sorting etc) so I have tried instead to focus on building the table variants up from cleaner component/style primitives.


delasous:
Failing typecheck is a problem with shared package type exports. It is resolved in #390

delasous:
Closes T-349
Closes T-350

Rebuilds InventoryTable's with shared shared base components and styles.
Adds separate Stock and Inbound tables for separate layout, and (coming soon) row/cell tooltips

ikusteu (approved):
👌

PR #397, opened 2024-02-09
Enhance/text editable

delasous:
Closes T-275

- Updates structure and styles to:
    - bring it inline with table editable fields
    - allow underlying text element type to be changed out and styled
- Reintroduces into in/out note view to allow editing of note name


ikusteu (approved):
Left a small hair-splitting comment.

As far as the functionality is concerned, LGTM

PR #398, opened 2024-02-15
Feature/supplier interface

fadwamahmoud:
Link T-314


pkg/db/src/types/orders.ts
```diff
@@ -9,10 +9,55 @@ export interface NewOrdersDatabase {
...
+
+/**
+ * Methods interface
+ */
+export interface SupplierOrderProto<A extends Record<string, any> = {}> {
+	create: () => Promise<SupplierOrderInterface<A>>;
+}
+
+/** Data structure */
+export type SupplierOrderData<A extends Record<string, any> = {}> = CouchDocument<
```
ikusteu, line 46: Let's rename these to supplier batch: `OrderBatchInterface`, `OrderBatchProto`, etc...and drop the confusion between order batch/supplier order (both one in the same)

pkg/db/src/types/orders.ts
```diff
@@ -9,10 +9,55 @@ export interface NewOrdersDatabase {
...
+ */
+
+export type SupplierOrderInterface<A extends Record<string, any> = {}> = SupplierOrderProto<A> & SupplierOrderData<A>;
+
+/**
+ * Methods interface
+ */
+export interface SupplierOrderProto<A extends Record<string, any> = {}> {
+	create: () => Promise<SupplierOrderInterface<A>>;
+}
```
ikusteu, line 43: The proto should contain methods:
- create
- delete
- commit*
- csv - this should be available only on committed batches
- addItems - akin to `addTransactions` in the note interface
- stream:
  - displayName
  - items
  - state
  - updatedAt
 

*Let's use `commit` instead of `order` as the latter might imply:
- order interface (one that doesn't exist, but would be consistent with our api, something like `db.supplier("supplier-1").batch("order-batch-1").order("order-1")`)
- `order` - as in `sort`
- `order` - the actual use case (place the order)
Using `commit` removes ambiguity and keeps consistent with the rest of the db interface

pkg/db/src/types/orders.ts
```diff
@@ -9,10 +9,55 @@ export interface NewOrdersDatabase {
...
+
+/** Data structure */
+export type SupplierOrderData<A extends Record<string, any> = {}> = CouchDocument<
+	{
+		updatedAt: string | null;
+		books: OrderBook[];
+		status: string;
```
ikusteu, line 50: Let's use `state` rather than `status` to keep consistent with the inventory api

pkg/db/src/types/orders.ts
```diff
@@ -9,10 +9,55 @@ export interface NewOrdersDatabase {
...
+		csv: string;
+		date: string;
```
ikusteu, line 52: `date` should probably be optional (assigned only when the batch was committed - order placed with the supplier)

ikusteu (changes requested):

PR #403, opened 2024-02-20
Fix/slideover navigation

ikusteu:
Does the same thing as #401  only in a more organised fashion: 
- also removes the '/inventory/(warehouses|inbound)' shared layout
- doesn't duplicate the code, but rather moves the "layout" code into an `InventoryManagementPage`

Depends on #402 (passing e2e tests)


apps/web-client/src/routes/inventory/inbound/+page.svelte
```diff
@@ -64,6 +65,7 @@
 	let dialogContent: DialogContent;
 </script>
 
+<<<<<<< Updated upstream
```
silviot, line 68: Merge conflict leftover

ikusteu: Yess

delasous (approved):

PR #404, opened 2024-02-20
Fix book data fetching not working

fadwamahmoud:
The bug was caused by 
- isbn not passed in array in BookForm 
- fetchBookData not called on scan

Fixed both and added a tsconfig.build.ts for a declaration file build step


delasous:
I fixed the linting error. But noticed that the plugins tests are failing? Also, are they being run in CI?

plugins/book-data-extension/src/plugin/index.ts
```diff
@@ -10,6 +10,8 @@ export const createBookDataExtensionPlugin = (): BookFetcherPlugin => {
 		const extensionAvailable = await listenForExtension(`BOOK_FETCHER:PONG`, 500);
 
 		if (!extensionAvailable) return [];
+		console.log("Extension Available");
+		localStorage.setItem("extensionAvailable", "true");
```
silviot, line 14: I don't think localstorage is conceptually the right spot to store this info. I think a svelte store would be a better fit.

Also, the check for the extension presence should happen on page load IMO, not at the first attempt to invoke `fetchBookData`.

silviot (changes requested):
I think the logic that detects whether the extension is installed or not needs some changes.

I was not able to see this working. I probably have problems with my extension. If I run `window.postMessage({ message: 'BOOK_FETCHER:REQ:aaa' }, "*")` in the console I don't see any action in the extension Network tab, as I'd expect.

plugins/book-data-extension/src/plugin/index.ts
fadwamahmoud: @silviot Yeah, I agree that checking for the extension should happen on page load because the way I implemented it would display a disabled button when the app loads for the first time (before any call to fetchBookData).

My reasoning behind using localStorage is 
a) I couldn't use a svelte store because the plugin package does not use svelte. I could use a different way to manage the state, this way was just the most straight forward.
b) Despite localStorage not being reactive by design, the fetchBookData call listens for the extension availability (and updates the localStorage) on each call so the value would never go stale and it would be constantly synced.

silviot: @fadwamahmoud Can we use a store-like object that can then be imported from the rest of the app?

Something like this [ChatGPT produced this code and I haven't thoroughly checked it]:

```typescript
class BooleanStore {
  private _value: boolean;
  private _subscribers: Array<(value: boolean) => void> = [];

  constructor(initialValue: boolean) {
    this._value = initialValue;
  }

  subscribe(run: (value: boolean) => void): () => void {
    this._subscribers.push(run);
    run(this._value); // Run the subscriber immediately with the current value

    return () => {
      this._subscribers = this._subscribers.filter((sub) => sub !== run);
    };
  }

  set(value: boolean): void {
    this._value = value;
    this._subscribers.forEach((sub) => sub(value));
  }

  update(updater: (currentValue: boolean) => boolean): void {
    this.set(updater(this._value));
  }
}

// Usage:
// const store = new BooleanStore(true);
// store.subscribe(value => console.log(value));
// store.set(false);
// store.update(value => !value);
```

Since there's no load funciton here, the check (and storing the result of the check into the store like object) could then happen "at import time" (or "at module level") in python parlance.

"At the top level of the module"
"Outside of any component or function"
"In the global scope of the module"

silviot (approved):

PR #410, opened 2024-02-26
Feat/outbound note reconciliation

ikusteu:
Closes T-358

When the user tries to commit the outbound note the check is performed:
- if there's at least one transaction without warehouse specified, the 'no-warehouse-selected' modal is displayed
- if all warehouses are specified, but some transactions are our of stock - an out-of-stock dialog is displayed (prompting the user for reconciliation and commit)
- if user confirms the out-of-stock (reconciliation) dialog, an inbound note with `reconciliationNote: true` is created in the background (balancing out the out-of-stock transactions of the outbound note in question)

The db tests and e2e tests have been updated to test / expect this behaviour.

This is flow is rather hard to communicate (to technical people), and it's even harder to communicate to your average cash-register operator, so I'm fairly certain we will need some UX updates communicating the changes made, and changes to be made **if the dialog is confirmed**

### DIFF
--- a/apps/e2e/integration/note-transactions/outbound_note_transactions.spec.ts
+++ b/apps/e2e/integration/note-transactions/outbound_note_transactions.spec.ts
@@ -5,6 +5,7 @@ import { baseURL } from "@/constants";
 import { getDashboard, getDbHandle } from "@/helpers";
 
 import { book1 } from "../data";
+import { compareEntries } from "@/helpers/utils";
 
 test.beforeEach(async ({ page }) => {
 	// Load the app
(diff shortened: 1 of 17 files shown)


ikusteu:
@silviot I've applied the requested updates. I've made some opinionated decisions (as I saw the solution to be simplest one):
- when displaying a spinning arrows icon next to warehouses that will require note reconciliation, only the book being available inside the warehouse, or not, is taken into account and not the selected quantity vs. available quantity per warehouses:
  - my reasoning is:
    1. It's simpler
    2. The spinning icon will always be there (or not), regardless of quantity changes (otherwise the behaviour might be a bit confusing)
  - the tradeoff is - the row might require reconciliation, in case the required quantity is greater than one available in the warehouse, but that won't be denoted inside the warehouse picker, **however**, such case will be apparent by virtue of the row being painted red (signalling out-of-stick - required reconciliation)
- when the book is available in multiple warehouses, **only** if no warehouse is selected, will the row be painted yellow, reasoning - once the warehouse is selected we no longer indicate that some action is required

silviot (approved):

PR #412, opened 2024-03-16
Feat/Remove discount calculation from table entries pipeline

laranotreallycroft:
Closes T-353

### DIFF
--- a/apps/e2e/helpers/table.ts
+++ b/apps/e2e/helpers/table.ts
@@ -204,7 +204,7 @@ const fieldConstructorLookup: {
 	isbn: stringFieldConstructor("isbn"),
 	title: stringFieldConstructor("title"),
 	authors: stringFieldConstructor("authors"),
-	price: stringFieldConstructor("price", (x) => (x === "N/A" ? x : `€${x}`)),
+	price: stringFieldConstructor("price", (x) => (x === "N/A" ? x : `€${Number.parseFloat(x).toFixed(2)}`)),
 	quantity: quantityFieldCostructor,
 	publisher: stringFieldConstructor("publisher"),
 	outOfPrint: outOfPrintFieldConstructor,
--- a/apps/e2e/helpers/table.ts
+++ b/apps/e2e/helpers/table.ts
@@ -148,6 +148,15 @@ const stringFieldConstructor =
 			expect(row.locator(`[data-property="${name}"]`)).toHaveText(transformDisplay(want.toString()), { timeout: assertionTimeout, ...opts })
 	});
 
+const priceFieldConstructor: FieldConstructor<"price"> = (row) => ({
+	assert: (want: string | number, opts) =>
+		typeof want === "string"
+			? expect(row.locator(`[data-property="price"]`)).toHaveText(want, { timeout: assertionTimeout, ...opts })
+			: expect(row.locator(`[data-property="price"]`)).toHaveText(`€${want.toFixed(2)}`, {
+					timeout: assertionTimeout,
+					...opts
+			  })
+});
 const quantityFieldCostructor: FieldConstructor<"quantity"> = (row, view) => ({
 	assert: (want, opts) =>
 		view === "warehouse"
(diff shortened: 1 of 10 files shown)


apps/web-client/src/lib/stores/proto/table_content.ts
```diff
@@ -70,6 +62,8 @@ const mergeBookData = (stock: Iterable<VolumeStockClient>) => (bookData: Iterabl
 		.map(([s, b = {} as BookEntry]) => ({ ...s, ...b }))
 		.array();
 
+//function left if needed for future use
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
ikusteu, line 66: Yeah, let's not do that: we want to keep linting and TS as strict as possible (unless it's really blocking, annoying and would need more engagement to fix, then we're willing to provide at given moment).
I this case, however, If we ever need the fn, we can either check git history, or, merely, re-create it.

ikusteu (changes requested):
Small comment about undead functions

ikusteu (changes requested):
There seems to be a problem with the way warehouse discount is calculated. This isn't your code's fault, but could you, please, check and fix that.

Faulty behaviour:
Screenshot 2024-03-16 at 10 15 13

Warehouse discount value is subtracted from the book price, and not used to calculate the value with applied discount (subtracting 0.2, instead of multiplying by (1-0.2 in this case).

ikusteu:
I get a weird lag:
1. the inventory is displayed without warehouse discount applied
2. (after some time) the UI is updated with the applied discount

I would like for the applied discount UI to be displayed as initial paint (this could be a different ticket, just noting this down)

apps/web-client/src/lib/components/Tables/InventoryTables/BookPriceCell.svelte
```diff
@@ -9,12 +9,12 @@
...
+		<span>€{((price * (100 - warehouseDiscount)) / 100).toFixed(2)}</span>
 		<span class="sr-only">Original price:</span>
-		<span class="text-gray-400 line-through">(€{price})</span>
+		<span class="text-gray-400 line-through">(€{price.toFixed(2)})</span>
 		<span class="sr-only">Percentage discount:</span>
 		<span class="text-gray-400">-{warehouseDiscount}%</span>
 	</div>
 {:else}
-	{price ? `€${price}` : "N/A"}
+	{price ? `€${price.toFixed(2)}` : "N/A"}
```
ikusteu, line 19: Can you refactor this so that:
- if there's only price, the price is shown in a span element with `data-property="full-price"`
- if the discount is also present, the same logic is used:
  - full price is a span with `data-property="full-price"`
  - discount is a span with `data-property="applied-discount"`
  - discounted price is a span with `data-property="discounted-price"`
  
  That way we could also update the price element in e2e api for a more fine grained control (and avoid the full string match)

apps/e2e/helpers/table.ts
```diff
@@ -148,6 +148,15 @@ const stringFieldConstructor =
...
 
+const priceFieldConstructor: FieldConstructor<"price"> = (row) => ({
+	assert: (want: string | number, opts) =>
+		typeof want === "string"
+			? expect(row.locator(`[data-property="price"]`)).toHaveText(want, { timeout: assertionTimeout, ...opts })
+			: expect(row.locator(`[data-property="price"]`)).toHaveText(`€${want.toFixed(2)}`, {
+					timeout: assertionTimeout,
+					...opts
+			  })
+});
```
ikusteu, line 159: Continuing from the comment in the price display element(s), let's have the price field asserter ☝️ accept a price that is either:
- string - used to only match the price (regardless of price being discounted
- object with shape:
  - price - string
  - discount (optional) - number
  - discounted price (optional) - string

That way we have a more fine grained control over the assertion:
- if string - only the full price is asserted (regardless of discount and discounted price
- if object - more strict approach:
  - every field is asserted exactly as specified
  - if optional fields are present - they're expected to not be there

ikusteu (changes requested):
Nice job on the tests - achieved more/less the same result that I would have expected had I written up a ticket (sorry about not doing so, btw.), but I've found some potential tweaks that would improve the testing/asserting to be more robust, but also more concise in calls (avoiding the full price string)

ikusteu (approved):
LGTM

PR #415, opened 2024-03-24
Feat/scan on off

ikusteu:
Closes T-360

### DIFF
--- a/apps/e2e/helpers/scanField.ts
+++ b/apps/e2e/helpers/scanField.ts
@@ -1,3 +1,4 @@
+import { expect } from "@playwright/test";
 import { DashboardNode, ScanFieldInterface } from "./types";
 
 import { selector, testIdSelector } from "./utils";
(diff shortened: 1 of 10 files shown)


apps/e2e/helpers/scanField.ts
```diff
@@ -11,5 +12,32 @@ export function getScanField(parent: DashboardNode): ScanFieldInterface {
...
+	const dashboard = parent.dashboard;
+	const button = dashboard()
+		.content()
+		.locator(selector(testIdSelector("scan-autofocus-toggle")));
+
+	const isOn = () => button.getAttribute("data-is-on").then((value) => value === "true");
+
+	const toggleOn = async () => {
+		// Noop if the autofocus is already on
+		if (await isOn()) return;
```
silviot, line 30: isOn is not async

silviot: It is indeed async: `button.getAttribute("data-is-on").then(...)` is a promise.

silviot (approved):

PR #416, opened 2024-03-25
Feat/If warehouse doesn't exist, don't add it to available warehouses

laranotreallycroft:
Closes T-366

### DIFF
--- a/pkg/db/src/implementations/version-1.2/stock.ts
+++ b/pkg/db/src/implementations/version-1.2/stock.ts
@@ -40,8 +40,7 @@ class Stock implements StockInterface {
 			since: "now",
 			live: true,
 			// We don't care about changes to non-committed notes (as they don't affect the stock)
-			// Need to watch for warehouse deletion to show available warehouses correctly in outbound table
-			filter: (doc) => doc.committed || (doc?.docType === "warehouse" && doc?._deleted === true)
+			filter: (doc) => doc.committed
 		});
 	}
 
--- a/pkg/db/src/implementations/version-1.2/stock.ts
+++ b/pkg/db/src/implementations/version-1.2/stock.ts
@@ -40,7 +40,8 @@ class Stock implements StockInterface {
 			since: "now",
 			live: true,
 			// We don't care about changes to non-committed notes (as they don't affect the stock)
-			filter: (doc) => doc.committed
+			// Need to watch for warehouse deletion to show available warehouses correctly in outbound table
+			filter: (doc) => doc.committed || (doc?.docType === "warehouse" && doc?._deleted === true)
 		});
 	}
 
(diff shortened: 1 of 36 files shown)


pkg/db/src/implementations/version-1.2/utils.ts
```diff
@@ -68,10 +68,12 @@ export const addAvailableWarehouses = (
...
-			map(stock.isbn(e.isbn), ([[, warehouseId], { quantity = 0 }]) => [
-				warehouseId,
-				{ displayName: warehouses.get(warehouseId)?.displayName || "not-found", quantity }
-			])
+			wrapIter(stock.isbn(e.isbn))
+				.filter(([[, warehouseId]]) => warehouses.get(warehouseId) != null)
+				.map(([[, warehouseId], { quantity = 0 }]) => [
+					warehouseId,
+					{ displayName: warehouses.get(warehouseId)?.displayName || "not-found", quantity }
+				])
```
ikusteu, line 76: On the plus side, this is a quick and easy solution, making the app work as we'd expect it to. Additionally, it displays a good understanding of the generators library we're using for iterable operations.

However, this is a patch-me-up solution where we're "treating the symptoms, not the cause". We're not removing the warehouse stock from existence, but rather adding more code to block the data that we no-longer want there from being shown up and, as such, leads to unnecessary entropy (both in code and in the data).

Can we move this fix upstream (so that the stock passed here doesn't include the non-existing warehouses to begin with)

ikusteu (changes requested):

pkg/db/src/implementations/version-1.2/stock.ts
```diff
@@ -51,11 +52,20 @@ class Stock implements StockInterface {
...
 		const mapGenerator = wrapIter(queryRes.rows)
 			.map(({ doc }) => doc as Doc)
 			.filter((doc): doc is NoteData => doc?.docType === "note")
 			.filter(({ committed, entries }) => Boolean(committed && entries?.length))
-			.flatMap(({ entries, noteType }) => wrapIter(entries).map((entry) => ({ ...entry, noteType })));
+			.flatMap(({ entries, noteType }) =>
+				wrapIter(entries)
+					.filter((entry) => warehouses.some((warehouse) => warehouse._id === entry.warehouseId))
+					.map((entry) => ({ ...entry, noteType }))
+			);
```
ikusteu, line 68: I have a hunch this will bugs

ikusteu: Ok, scratch that ☝️ it will work, but let's solve the problem at the root:

Can we move the fix even further upstream: have the `queryRes` no include entries belonging to non-existing warehouses?

ikusteu (changes requested):

ikusteu (approved):

PR #417, opened 2024-04-05
Feat/custom outbound item

ikusteu:
This is really a WIP, but it should provide the full functionality.

TODOs / issues:
- Separate into more sensible commits
- Review the work and, potentially reduce the type drilling (`VolumeStock<"book">`, etc)
- Review (unit) tests and verify that the previous behaviour shouldn't break and the new one is thoroughly tested
- Extend e2e tests to test the new behaviour
- Fix book form (neither book form, nor custom item form seem to be working)
- Fix weird flashing when removing an item (happens for both book rows and custom items) - low priority: potentially a separate ticket


silviot:
I think a UX low hanging fruit for this would be to have it immediatly open the edit form of the newly added custom item after clicking the `Custom Item` button.

It can go in as is, but that single change would make a big difference.

silviot (approved):

PR #431, opened 2024-05-05
book data fetching: minor fixups:

ikusteu:
* remove all direct calls to bookdata-extension plugin (use db.plugin("book-fetcher") as single source of truth)
* when scanning a book:
	* check if book data exists first
	* if book data exists (however incomplete), don't fetch additional data (this can be done explicitly using book form)
	* fetch book data from 3rd party ONLY if no data present

The updates above :point_up: fix the problem of book data being overwritten (including the price) when adding new price as book data would **always** get fetched (regardless of there being some book data in the db)


ikusteu:
YOLO

PR #433, opened 2024-05-07
Feat: compose BookRow to use tooltip only in stock/search view

delasous:
Removes Warehouse Name tooltip from warehouse table view


ikusteu:
The test is failing. Normally I would remove the failing test as I don't consider the functionality supported anymore, but it's a bit ambiguous:
- in warehouse view, the vertical dots button (and additional row actions) is not visible
- in the warehouse +page.svelte, the book form and the row actions are still there (in the code)

Both having and not having the additional actions makes sense, but can you, please, see which route you're going with and remove the unnecessary bits of the other one.

delasous:
e2e caught an error - I did not intend to remove the row actions from the warehouse view. I misjudged the slot composition. It should be fixed now.

I have also adjusted tooltip options so that it won't linger when content is hovered. This was causing multiple tips to appear at once if we hovered a both the next row and the content of the current row

ikusteu:
Perfect, thanks for this PR: I wanted to report some peculiar behaviour with those tooltips, glad someone remembered to address this.

ikusteu (approved):

PR #444, opened 2024-05-21
Include category field in book interface and tables

fadwamahmoud:


apps/web-client/src/lib/components/Tables/InventoryTables/OutboundTable.svelte
```diff
@@ -54,7 +55,18 @@
...
+					rowIx,
+					isbn,
+					authors = "N/A",
+					quantity,
+					price,
+					year = "N/A",
+					title = "N/A",
+					publisher = "",
+					warehouseDiscount,
+					category = ""
```
delasous, line 68: The custom item row is out of sync with the dynamic table. We need to add an additional `<td>` there to account for Category col space on big screen. 

Additionally there's a bug from before: these empty `td`s should use the same responsive classes as a normal row.

delasous (approved):
LGTM. Only one comment on OutboundTable custom item row

apps/web-client/src/lib/components/Tables/InventoryTables/StockBookRow.svelte
```diff
@@ -51,6 +50,9 @@
 	<td data-property="editedBy" class="show-col-xl table-cell-max">
 		{editedBy}
 	</td>
+	<td data-property="category" class="show-col-xl table-cell-max">
```
silviot, line 53: I think lines 53-55 (the new category cell) should be moved after the BookOutPrintCell.
This is the current stock view:
image

PR #445, opened 2024-05-21
Extend createDisplayEntriesStore with a second generic type

fadwamahmoud:


delasous:
After a bit of investigation, it looks like the types aren't reliably defined in `proto/table_content.ts` and are really a result of the `WarehouseDisplayStores` interface in `proto/index.ts`. Sorry for wasting your time on this, but I'd suggest we squash this PR as it doesn't do much for the return type of `createDisplayEntriesStore`

PR #446, opened 2024-05-22
Feat/book transaction history

ikusteu:
- splits history into subviews
- creates transaction history page
- creates past note list page
- uses the URL to navigate between dates on daily summary/past note views
- default view is daily summary/<today>

(pushed to this branch by accident - should have been multiple PRs, but here we are)

### DIFF
--- a/apps/web-client/src/lib/components/HistoryPage.svelte
+++ b/apps/web-client/src/lib/components/HistoryPage.svelte
@@ -22,13 +22,13 @@
 			label: "By ISBN",
 			href: appPath("history/isbn"),
 			linkto: historyView("history/isbn")
+		},
+		{
+			icon: Book,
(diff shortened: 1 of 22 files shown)


silviot:
Can we include the time in the "Notes by date" list?
It makes sense to show them _instead_ of the date as things currently are: all notes in one page are from the same day.
Also, can we fit more notes in a single desktop screen?
Currently I can see four.
It would make sense to pack more info on a single line, and have them wrap if needed.

silviot (approved):
I left a few comments for possible improvements.

The work can go in as is though: it improves the status quo and does not introduce any problem AFAIC.

silviot:
image
ChatGPT has to say this...
I verified the new functionality works as expected by manual testing.

Issue #456, opened 2024-05-27
An invalid CouchDB URL makes the app crash


silviot:
If you make a mistake when entering the remote couchdb URL you'll have a hard time putting it back.

The settings page should always be accessible, even if the CouchDB URL is invalid/points nowhere.

PR #471, opened 2024-06-01
Handle invalid couchdb url without crashing the app

fadwamahmoud:
#456
This handles url crashing in two ways
1. When url is incorrect 
    - Redirect to settings page
2. When connection is lost at any point 
    - A store is used for checking the url connectivity 
    - That check happens when views are switched 
    - Using the same notification toast that displays the book extension status
    
Enhancements for next iteration:

- Use an error boundary
- Display reason when user is redirected: Unauthorized or url is invalid


silviot (approved):
Works as expected when tested in a browser.

PR #473, opened 2024-06-01
Fix/history isbn search fixes

### LINKED ISSUE #468: Full text seems to be a bit broken
I got weird results when employing the full text search functionality:
1. Book data not shown (even though it exists):
I've added a book by scanning a physical copy and fetching the data using book fetcher plugin. I can verify the data exists as, when I go to book history page (by explicitly typing in the isbn), I can see the title, authors and other data. However, **when searching, the book is just not shown**. This happens regardless of refreshes, rebuilds and such - the book data for the particular book exists for some time now.

2. Some book data is duplicated:
This is the data for a book existing in the production dump. The book in question is titled "THE SELFISH GIANT", with both entries having the same isbn and the rest of the data. My suspect is same isbns, with one including a leading/trailing whitespace, but this should be investigated.

### Screenshots

This shows data for "The Prince" by Machiavelli obviously existing in my db:

Screenshot 2024-05-31 at 09 57 44

The following shows both the data for "The Prince" not being found (point 1), as well the data for "THE SELFISH GIANT-THE HAPPY PRINCE" (by Oscar Wilde) being displayed twice:

Screenshot 2024-05-31 at 09 57 53

### LINKED ISSUE #469: Extend full text search to also allow for books with no data (only isbn)
Currently if there's a book in stock (or history), isbn-only, with no corresponding book data, the book can't be found by merely typing in the isbn (as fts uses books from the book data section of the db to build an index).
Thinking about it now, this really an edge case - as at least a price will be needed in order to use the book (in inbound/outbound, etc.).

I propose a simple fix: create a new (empty) book entry every time a new isbn is added - this way, the index will be build on all instances going in/out. (however, to make sure it **really** works, we should sort out the #468 )

ikusteu:
Fixes #469 
Fixes #468 

In order to display existing isbns, even for entries for which there's no data, I've made it so that each time `note.addVolumes` is called, if there's no entry for a book, one will be created with empty data `{isbn: string}`.

I know @silviot is generally against automatically adding data, but this merely creates an entry (only the first time the isbn is encountered).

### DIFF
--- a/apps/e2e/integration/note-transactions/outbound_note_transactions.spec.ts
+++ b/apps/e2e/integration/note-transactions/outbound_note_transactions.spec.ts
@@ -199,13 +199,14 @@ test("should delete the transaction from the note when when selected for deletio
 test("transaction should default to the only warehouse the given book is available in if there is only one", async ({ page }) => {
 	// Setup: Add the book to the warehouse (through an inbound note)
 	const dbHandle = await getDbHandle(page);
-	await dbHandle.evaluate(async (db) => {
-		db.warehouse("wh-1")
+	await dbHandle.evaluate((db) =>
+		db
+			.warehouse("wh-1")
 			.setName({}, "Warehouse 1")
 			.then((wh) => wh.note().create())
 			.then((n) => n.addVolumes({ isbn: "1234567890", quantity: 1 }))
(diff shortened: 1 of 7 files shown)


silviot (approved):
I clicked around and it looks OK

PR #478, opened 2024-06-04
Enhance/data loading

delasous:
- Uses `ReplaySubject`s instead of `BehaviourSubject`s for warehouse, in-note and out-note list streams to "cache" values. This removes the need for default values required by `BehaviourSubject`s which were making it difficult to tell between empty and loading states, but still ensures view switching is snappy post the initial load.
- Makes a slight adjustment to the warehouse list - stock stream merging to defer the calculation of the total book count.
- Adds a loading skeleton "..." to "total book" count in the warehouse list view to account for this delayed data loading

**Todo**
- Write up my understanding of the slice of db interface & rx streams that I broke down to get to these solutions
- Apply the same patterns in the table views


delasous:
Not sure how my current work could have affected the failing db test. Very happy to fix it, but feel like I'm missing some context around the db test setup / data loader. @ikusteu could you advise?

pkg/db/src/implementations/version-1.2/inventory-db.ts
```diff
@@ -63,24 +63,39 @@ class Database implements InventoryDatabaseInterface {
...
+								? of(warehouses)
+								: of(
+										mapIter(
+											warehouses,
+											([id, warehouse]) => [id, { ...warehouse, totalBooks: (s as StockMap).warehouse(id).size }] as const
+										)
+								  )
+						),
+						// Multi-cast the potentially long-running mergeMap to prevent redundant execution fo each subscriber
+						share()
```
fadwamahmoud, line 89: Could caching the last emission using shareReplay make it potentially faster?

fadwamahmoud (approved):

apps/web-client/src/routes/outbound/+page.svelte
```diff
@@ -33,8 +33,9 @@
 
 	let initialized = false;
 	onMount(() => {
-		if (status) firstValueFrom(outNoteListStream).then(() => (initialized = true));
-		else {
+		if (status) {
+			firstValueFrom(outNoteListStream).then(() => (initialized = true));
```
ikusteu, line 37: Ok, I'd like to go back on my suggestion for using the variant `{ ready: false, notes: [] } | { ready: true, notes: NavEntry[]` (or whichever the type for note list), as I see now that `firstValueFrom` is used on a stream, not on the `readable` from said stream - so, yeah, it makes sense to address that at the DB interface level.

My suggestion would still work, I guess, but it also make sense to untangle the streams a bit (if that's what you're going with)

apps/web-client/src/routes/inventory/warehouses/+page.svelte
```diff
@@ -111,9 +109,14 @@
...
 
 							<div class="flex flex-col gap-2 sm:flex-row">
-								<div class="flex w-32 items-center gap-x-1">
-									<Library class="text-gray-700" size={20} />
-									<span class="entity-list-text-sm text-gray-500">{totalBooks} books</span>
+								<div class="entity-list-text-sm flex w-32 items-end items-center gap-x-1 text-gray-500">
+									<Library class="text-gray-700" size={18} />
+									{#if totalBooks === -1}
+										<PlaceholderDots />
+									{:else}
```
ikusteu, line 116: Love it 👍

ikusteu (approved):
LGTM, I condone the updates, I don't think this should break anything*, I didn't notice any loss to snappiness of view changing.

_*having said that: the db tests **are** failing, so something **did** break. I'm looking into it (as requested) - the db tests breaking doesn't necessarily mean the functionality can't be updated to work with the latest updates._

PR #479, opened 2024-06-04
Fix/invalid couchdb

fadwamahmoud:
#456 Enhancements

- Use an error boundary: 
    - the +error.svelte page catches all error thrown inside routes, so it doesn't catch errors thrown inside layout.ts or errors thrown inside the handle hook
    - it's displayed when app goes offline while a couchdb url is provided
- Display reason when user is redirected:
    - username and password are invalid
    - url format is invalid


silviot (commented):
[Screencast from 05-06-24 14:03:57.webm](https://github.com/librocco/librocco/assets/396425/d651ea8e-bb28-4aa5-9e08-9339e4a4d694)

It works, but is not super consistent: if I click to o away from the settings page the URL briefly flashes the route I clicked, but then goes back to the settings URL (but the page shown is a different one).

I suspect these issues are symptoms of an underlying problem.

silviot (approved):

PR #490, opened 2024-06-07
Enhancement/open library

fadwamahmoud:


silviot:
Ideally we'd like (I think) all plugins to be tried at once.
So all requests (google books and open library) will be fired at once.
When the first one returns book data, we should immediately store them.
This way it will be as fast as the fastest of the two.

Ideally when the second requests comes in we should merge the two results to get the best of both. But I believe it's not easy at all, and probably a job for an LLM.

ikusteu:
> Ideally we'd like (I think) all plugins to be tried at once. So all requests (google books and open library) will be fired at once. When the first one returns book data, we should immediately store them. This way it will be as fast as the fastest of the two.
> 
> Ideally when the second requests comes in we should merge the two results to get the best of both. But I believe it's not easy at all, and probably a job for an LLM.

See my comment in #491

ikusteu:
Addressed in #491

PR #491, opened 2024-06-09
book fetcher plugin controller:

ikusteu:
* add a root book fetcher plugin as a controller and register all instances against it
* when fetching book data the controller will perform fetch on all registeres instances and merge the results
* when checking for availability (isAvailableStream) - the plugin controller will emit `true` if at least one instance is available

Fixes #477

PR #492, opened 2024-06-10
fetch book data each time a new (unknown) book is added to the note)

ikusteu:
@silviot 

I've tried this with the scanner and some books I have on the side. It worked for me.

_Note: this is a quick fix, should be rolled back and fixed in a more thought-through manner_


silviot:
Thanks! I'll use this code for now. I don't think we want to merge the PR, since it makes the tests red.

ikusteu:
> Thanks! I'll use this code for now. I don't think we want to merge the PR, since it makes the tests red.

If you want to use the code from #491 see #493

PR #501, opened 2024-06-25
Fix/settings flash

fadwamahmoud:
This PR eliminates the temporary page blinking before redirecting to the settings page ~~well, for the most part~~ . My intention was to achieve that by getting rid of the redundant url fetch requests but they were not the reason for the delay. The culprit was that the db was being created but then couldn't be accessed and would throw errors that went uncaught. 
There are two scenarios which can cause the url to crash. 
1. Invalid url: ~~This is a bit problematic still. It takes too long on the first load to redirect to the settings page and it throws in the console but all future navigation is lag-free.~~
2. Invalid credentials: ~~This one works the best. It redirects instantly.~~

Edit: After clicking through and manually testing the above on the deployed preview, it works fine.


silviot:
I tried clicking around and found an issue that should probably be tackled.

https://github.com/librocco/librocco/assets/396425/d9d30e91-a7ca-4edd-8a94-e28308654214

PR #510, opened 2024-07-26
Feat/dbms agnostic public interface

ikusteu:
Done this as a step towards SQLite implementation, but it can be considered a self-contained unit of work.

- refactored all of the public db interface to exclude any DBMS or implementation specific behaviour (note: replication is still PouchDB - only)
- the public api has docs use the (single segment) `id` instead of multi-segment (full path), pouch/couch specific `_id`


delasous (approved):
Looks like a series of very precise changes across a lot of files 👌

PR #524, opened 2024-10-06
Enable autocomplete for custom item titles in custom item form:

### LINKED ISSUE #442: Pre-written choices in custom item form
The custom item form will (almost) always be filled with the same values (a handful).

It would be nice if users could pick from a list.

Can we leverage the browser auto completion features for this? If not can we have a setting (local storage or db) with a list of common items that will be selectable when filling in the form?

ikusteu:
* use 'librocco-custom-item-title' as field name (in lieu of 'title' to prevent autocompletion saved for other, generic 'title' elements)
* use string proxy to proxy the input as 'title' in svelte superforms (more straightforward handling of submitted form)
* enable autocompletion for aforementioned field

Fixes #442

### DIFF
--- a/apps/web-client/src/lib/forms/CustomItemForm.svelte
+++ b/apps/web-client/src/lib/forms/CustomItemForm.svelte
@@ -7,7 +7,7 @@
 
 <script lang="ts">
 	import type { ZodValidation } from "sveltekit-superforms";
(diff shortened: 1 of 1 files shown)


delasous:
Should I review this? It's been kicking around a while

ikusteu:
Stale...probably not worth keeping around (will revisit when we get to it)

PR #540, opened 2024-10-08
order management: spec:

ikusteu:
* create initial types spec
* keep the old types spec (as to not wreak havoc on the repo -- should be removed as when we start with the implementation)


ikusteu:
Here as a reference, I don't see a point in having this in `main`

PR #554, opened 2024-10-18
Wire in customer orders and order lines

fadwamahmoud:
~Reactivity is buggy but it's temporary~
~reloading causes an error (same crsqlite wasm build error) but if you click on the customers route button it should load~
Edit quantities needs to be implemented
Delete row: what's the delete logic?
Book data needs to be displayed ie year, title, etc.
SupplierIds need to be provided on adding a new book line


apps/web-client/src/lib/db/orders/db.ts
```diff
@@ -18,7 +18,7 @@ export async function getDB(dbname: string) {
 }
 
 export async function initializeDB(db: DB) {
-	await db.exec(`CREATE TABLE customer (
+	await db.exec(`CREATE TABLE IF NOT EXISTS customer (
```
silviot: I don't think `IF NOT EXISTS` is a good idea. Creating the table should be part of the db initialization. There are a couple of possible approaches:

* the db initialization is idempotent, and always run
* the db initialization might error out when performed on an already initialized db, and should be run only on a db that needs it.

I think the latter is what we want here: it will prevent db initialization from slowing down page load times when the db is already initialized.

I agree there are two `CREATE TABLE` and they both should be aligned, but I think the alignment should happen in the other direction, i.e. removing `IF NOT EXISTS` from the `customer_order_lines` creation statement.

ikusteu: Fair point (didn't know the `IF NOT EXISTS` pays a performance toll)

silviot: To clarify: I didn't mean to imply that `IF NOT EXISTS` pays a performance toll. My point was that between the two approaches:

* issue 10/20/30 - tens of queries every time the db object is created in the code. They will have an effect the first time (initialize the db) and no effect subsequent times
* issue one query every time the db object is created in the code. This query will determine whether the db needs initializing. If it does, the tens of queries needed for initialization will be sent to the db.

I prefer the latter, since I suspect executing a query and waiting for the result is in itself a performance toll, regardless of the query complexity/content.

PR #560, opened 2024-10-28
Enhancement/customer order UI

fadwamahmoud:
This PR:
1. Improves customer details UI from a basic form to TextEditable components
2. Achieves basic reactivity by separating stores:
    a. updating customer order details
    b. adding and removing order lines from customer orders
    c. creating customer orders
    d. updating order line quantity
3. Displays book meta data in order line table 
4. Implements updateRowQuantity function for order lines

### DIFF
--- a/apps/web-client/src/lib/components/TextEditable/NumberEditable.svelte
+++ b/apps/web-client/src/lib/components/TextEditable/NumberEditable.svelte
@@ -1,0 +1,109 @@
+<script lang="ts">
+	import { tick } from "svelte";
+	import { PencilLine } from "lucide-svelte";
+	import { clickOutside } from "$lib/actions";
+
+	export let name: string;
+	export let id: string = name;
+
+	/**
+	 * This is the exposed 'value' of the text/input element. This value accepts updates from the parent component, while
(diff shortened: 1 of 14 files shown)


apps/web-client/src/lib/components/TextEditable/NumberEditable.svelte
delasous, line 1: I think it would be better to pass the input as a slot in order to switch between text and number. The classes defind on the input can be passed as a slot prop and applied when the slot content is passed

delasous (approved):

silviot (changes requested):
I tried [the deployed app](https://test.libroc.co/enhancement/customerOrder_UI/customers/) and I was not able to edit customer details.

I also think the PR needs a description briefly mentioning what changes are in the PR and why they improve on the status quo.

fadwamahmoud:
> I tried [the deployed app](https://test.libroc.co/enhancement/customerOrder_UI/customers/) and I was not able to edit customer details.
> 
> I also think the PR needs a description briefly mentioning what changes are in the PR and why they improve on the status quo.

Fixed it. The input field values were not propagating back to the customer store

silviot (approved):

PR #568, opened 2024-11-08
Create markCustomerOrderAsReceived

fadwamahmoud:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new function to update the timestamp for when a customer order is marked as received.
	- Added functionality for managing supplier orders, enhancing order processing capabilities.

- **Bug Fixes**
	- Updated test cases to verify the accuracy of order status updates and timestamps. 

- **Documentation**
	- Updated import statements to reflect new dependencies related to supplier order management.

### DIFF
--- a/apps/web-client/src/lib/db/orders/__tests__/orders.test.ts
+++ b/apps/web-client/src/lib/db/orders/__tests__/orders.test.ts
@@ -16,7 +16,7 @@ import { createSupplierOrder, getPossibleSupplerOrderLines } from "../suppliers"
 import { createCustomerOrders, getRandomDb, getRandomDbs, syncDBs } from "./lib";
 
 describe("Db creation tests", () => {
-	it.skip("should allow initializing a database", async () => {
+	it("should allow initializing a database", async () => {
 		const randomTestRunId = Math.floor(Math.random() * 100000000);
 		const db = await getDB("init-db-test" + randomTestRunId);
 		await expect(getAllCustomers(db)).rejects.toThrow();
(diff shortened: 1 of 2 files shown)


apps/web-client/src/lib/db/orders/customers.ts
```diff
@@ -81,3 +81,23 @@ export const updateOrderLineQuantity = async (db: DB, bookId: number, quantity:
 	const params = [quantity, bookId];
 	await db.exec(sql, params);
 };
+
+export const markCustomerOrderAsReceived = async (db: DB, supplierOrderLines: SupplierOrderLine[]) => {
+	//query customer order lines with isbn and order them by placed
+	for (const line of supplierOrderLines) {
+		await db.execO<DBCustomerOrderLine>(
```
silviot, line 88: When we're not interested in the result we should use `db.exec` instead of `db.execA` and `db.execO` in my opinion. Also, coderabbit has a good point.

apps/web-client/src/lib/db/orders/__tests__/orders.test.ts
```diff
@@ -120,3 +127,37 @@ describe("Customer order tests", () => {
...
+		const possibleOrderLines = await getPossibleSupplerOrderLines(db);
+
+		expect(possibleOrderLines).toStrictEqual([
+			{ supplier_id: 1, isbn: "1", quantity: 1 },
+			{ supplier_id: 1, isbn: "2", quantity: 1 },
+			{ supplier_id: 2, isbn: "3", quantity: 2 }
+		]);
+		// creating a supplier order should update the timestamp of placed
+		const newOrders = await createSupplierOrder(db, possibleOrderLines);
+		// expect(newOrders).toStrictEqual([]);
```
silviot, line 148: Why is this here, but commented out? It should probably either stay or go, but it really needs to make up its mind.

apps/web-client/src/lib/db/orders/__tests__/orders.test.ts
```diff
@@ -120,3 +127,37 @@
...
+
+		// takes in supplier orders
+		// finds customerorderlines for those supplier orders
+		// updates the timestamp of received
+
+		await markCustomerOrderAsReceived(db, [...newOrders[0].lines, ...newOrders[1].lines]);
+		const books = await getCustomerBooks(db, 1);
+		expect(books[1].received).toBeInstanceOf(Date);
+		expect(books[2].received).toBeInstanceOf(Date);
+	});
```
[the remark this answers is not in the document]
silviot: It's true that comprehensive tests do have their place in the world. But in this case we're not 100% sure the defined interface will be final. So we prefer to write only the tests that exercize the basic functionality.

apps/web-client/src/lib/db/orders/customers.ts
```diff
@@ -81,3 +81,23 @@ export const updateOrderLineQuantity = async (db: DB, bookId: number, quantity:
...
+    FROM customer_order_lines
+    WHERE isbn = ?
+      AND placed IS NOT NULL
+    ORDER BY placed ASC
+    LIMIT 1
+);`,
+			[line.isbn]
+		);
+	}
+};
```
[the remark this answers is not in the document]
silviot: Good bot!

apps/web-client/src/lib/db/orders/__tests__/orders.test.ts
```diff
@@ -120,3 +127,37 @@ describe("Customer order tests", () => {
...
+	it("can update the timestamp of when a customer order is placed (to supplier)", async () => {
+		const possibleOrderLines = await getPossibleSupplerOrderLines(db);
+
+		expect(possibleOrderLines).toStrictEqual([
+			{ supplier_id: 1, isbn: "1", quantity: 1 },
+			{ supplier_id: 1, isbn: "2", quantity: 1 },
+			{ supplier_id: 2, isbn: "3", quantity: 2 }
+		]);
+		// creating a supplier order should update the timestamp of placed
+		const newOrders = await createSupplierOrder(db, possibleOrderLines);
```
silviot, line 147: I'm torn between suggesting to hardcode 
```
[
			{ supplier_id: 1, isbn: "1", quantity: 1 },
			{ supplier_id: 1, isbn: "2", quantity: 1 },
			{ supplier_id: 2, isbn: "3", quantity: 2 }
]
```

here or use `getPossibleSupplerOrderLines(db)` as a parameter. I do think that the assertion nedds to go though (as I stated in the comment above).

PR #572, opened 2024-11-09
change the way app version is passed to the web-client app:

### LINKED ISSUE #571: Error loading package.json in `web-client/src/lib/constants.ts` in dev mode
Running a dev server produces the following error:
```cli
Pre-transform error: Failed to load url /package.json (resolved id: (...)/librocco/apps/web-client/package.json) in (...)/librocco/apps/web-client/src/lib/constants.ts. Does the file exist?
SvelteKitError: Not found: /preview/package.json
```

This is the code in question:
```ts
import pkg from "../../package.json"
```

This should, in theory, work, but it doesn't for me. Idk if others are experiencing the same issue...anyways, I will move the GIT_SHA version and other related constants to the vite config to make sure the package.json import happens in node environment (pre bundling)

ikusteu:
* parse package.json in vite config and define the env variable (VITE_PKG_VERSION)
* rename VITE_VERSION (passed to dev/build process) to VITE_GIT_SHA (to avoid clashing and it communicates the git SHA anyway)

Fixes #571 

@silviot care to take a look?


delasous (approved):
Very similar to something I had to do in Edible. Confirm that it fixes it in dev and works in production

PR #574, opened 2024-11-12
Reconciliation order

fadwamahmoud:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new test suite for reconciliation order management.
	- Added functionality to create, retrieve, and update reconciliation orders.
	- New `reconciliation_order` table created in the database for managing reconciliation orders.
	- Exported `multiplyString` function for broader accessibility.
	- Enhanced `getSupplierOrder` function to include unique identifier in return object.

- **Bug Fixes**
	- Improved error handling in `upsertSupplier` function for supplier validation.

- **Documentation**
	- Updated type definitions to include optional `id` for `SupplierOrder` and added new `reconciliationOrder` type.

### DIFF
--- a/apps/web-client/src/lib/db/orders/reconciliation.ts
+++ b/apps/web-client/src/lib/db/orders/reconciliation.ts
@@ -1,4 +1,4 @@
-import { markCustomerOrderAsReceived, multiplyString } from "./customers";
+import { multiplyString } from "./customers";
 import type { DB, ReconciliationOrder } from "./types";
 
 export async function getAllReconciliationOrders(db: DB): Promise<ReconciliationOrder[]> {
--- a/apps/web-client/src/lib/db/orders/reconciliation.ts
+++ b/apps/web-client/src/lib/db/orders/reconciliation.ts
@@ -1,3 +1,4 @@
+import { multiplyString } from "./customers";
 import type { DB, ReconciliationOrder } from "./types";
 
 export async function getAllReconciliationOrders(db: DB): Promise<ReconciliationOrder[]> {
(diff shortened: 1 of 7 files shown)


fadwamahmoud:
@coderabbitai full review

apps/web-client/src/lib/db/orders/reconciliation.ts
```diff
@@ -0,0 +1,36 @@
...
+export async function createReconciliationOrder(db: DB, supplierOrderIds: string[]) {
+	if (!supplierOrderIds.length) {
+		throw new Error("Reconciliation order must be based on at least one supplier order");
+	}
+
+	const supplierOrderIdsJoined = supplierOrderIds.join(", ");
+
+	const tt = await db.execA(`INSERT INTO reconciliation_order (supplier_order_ids) VALUES (?) RETURNING id;`, [supplierOrderIdsJoined]);
+	return tt[0][0];
+}
```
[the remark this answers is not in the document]
fadwamahmoud: Making sure supplierOrder Ids exist makes more sense on finalizing the order

apps/web-client/src/lib/db/orders/reconciliation.ts
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * The way I see it while the isbns are being scanned they'll be cached
```
silviot, line 32: What's the improvement from debouncing ISBN additions to a reconciliation order? I can see the added complexity, but it's not clear to me what other advantages are there.

apps/web-client/src/lib/db/orders/reconciliation.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export async function createReconciliationOrder(db: DB, supplierOrderIds: string[]) {
+	if (!supplierOrderIds.length) {
+		throw new Error("Reconciliation order must be based on at least one supplier order");
+	}
+
+	const supplierOrderIdsJoined = supplierOrderIds.join(", ");
+
+	const tt = await db.execA(`INSERT INTO reconciliation_order (supplier_order_ids) VALUES (?) RETURNING id;`, [supplierOrderIdsJoined]);
```
silviot, line 28: If we use this shape (have the `supplier_order_ids` be a comma separated string) we won't be able to join the supplier orders with the reconciliation data. I need to think some more about this. Will get back to it!

silviot: After some pondering (and LLM chatting) I believe JSON support in sqlite is advanced enough that this layout/shape/architecture should not be a problem at all. We should change `supplier_order_ids` to be a full JSON array, and add a constraint for it to contain valid JSON, like `CHECK (json_valid(supplier_order_ids))`. Maybe also a check to make sure the JSON content is an array of at least length 1 (using `json_array_length`).

@coderabbitai can you confirm/critique my take on `supplier_order_ids`?

silviot: Interesting. Given what I learned here I can improve the layout of the list of supplier orders we maintain for each customer order line: it could use the same strategy.

PR #579, opened 2024-11-23
Add navigation to mockups

fadwamahmoud:


delasous (changes requested):
I understand the route you've taken, but I don't think it makes sense to blend the themes like this around the Page component.

=> I would suggest:
- removing the Page component from all of the `/orders` routes pages. 
- add a `+layout.svelte` to the root of `/orders` which copies the nav over from the Page component verbatim. Keep the nav in the Page component so it continues to work in the rest of the app and links to these new pages. Allow the existing routes to sit in the slot, without any of the Page headers above.

The nav on the layout.svelte can remain in the old styles, but at least on the next pass we can continue to update the app UI without clashing.

apps/web-client/src/routes/orders/customers/+page.svelte
```diff
@@ -0,0 +1,157 @@
...
+
+<Page view={view("orders/customers/id")} loaded={initialized}>
+	<svelte:fragment slot="topbar" let:iconProps let:inputProps>
+		<header class="navbar mb-4 bg-neutral">
+			<input type="checkbox" value="forest" class="theme-controller toggle" />
+		</header>
+	</svelte:fragment>
+
+	<svelte:fragment slot="main">
+		<main class="h-screen">
```
delasous, line 54: If a parent div sets h-screen the children should generally only set h-auto or h-full, otherwise it can remain unresponsive to the parent when the screen resizes

apps/web-client/src/routes/orders/+layout.svelte
```diff
@@ -0,0 +1,219 @@
...
+	 * Handle create note is an `on:click` handler used to create a new outbound note
+	 * _(and navigate to the newly created note page)_.
+	 */
+	const handleCreateNote = async () => {
+		const note = await getDB().db?.warehouse().note().create();
+		await goto(appPath("outbound", note.id));
+	};
+</script>
+
+<div id={testId("page-container")} data-view={view} data-loaded={loaded} class="flex h-screen w-screen overflow-hidden">
```
delasous, line 105: As far as I can tell, the props/attributes here are helpful for e2e testing. The new routes don't have any tests right now, so just leave this out

delasous (approved):

apps/web-client/src/routes/orders/+layout.svelte
```diff
@@ -0,0 +1,219 @@
...
+		},
+		{
+			label: tNav.supplier_orders(),
+			href: appPath("supplier_orders"),
+			icon: Truck,
+			linkto: "orders/suppliers"
+		}
+	];
+
+	export let view: WebClientView;
```
delasous, line 85: The layout will not receive these props

apps/web-client/src/routes/orders/+layout.svelte
```diff
@@ -0,0 +1,219 @@
...
+					{/each}
+				</ul>
+			</nav>
+		</div>
+	</div>
+	<!-- Sidenav end -->
+
+	<!-- Main content -->
+	<div id="content" data-loaded={loaded} class="relative flex h-screen w-full flex-col overflow-hidden bg-gray-50">
+		<!-- Main section -->
```
delasous, line 151: h-screen should be removed here. The parent div already sets it. Use h-full instead

apps/web-client/src/routes/orders/+layout.svelte
```diff
@@ -0,0 +1,219 @@
...
+	const handleCreateNote = async () => {
+		const note = await getDB().db?.warehouse().note().create();
+		await goto(appPath("outbound", note.id));
+	};
+</script>
+
+<div data-view={view} data-loaded={loaded} class="flex h-screen w-screen overflow-hidden">
+	<!-- Sidenav -->
+	<div class="flex">
+		<div class="hidden h-screen bg-gray-800 sm:inline-block">
```
delasous, line 108: CSS like `bg-gray-` and `text-gray-` are messing with the daisy-ui theme. Remove them

delasous:
Should I merge this? Is there a commit with the resolved comments?

fadwamahmoud:
> Should I merge this? Is there a commit with the resolved comments?

No, sorry. Forgot to push...Will merge it once I do

PR #585, opened 2024-11-30
Feat/cr sync dev

ikusteu:
- set up cr-server (in the `web-client` package) for dev purposes
- add a sync worker
- start sync worker (and start sync on `/customers/...` route group)
- fix reactivity for `/customers` route
- works on my machine :joy:
- no, really, the sync seems to be working (for customers routes) across tabs and browsers


ikusteu:
Merging this as I need it to proceed (to avoid conflicts)

PR #596, opened 2024-12-12
Wire in supplier order view and place order

fadwamahmoud:
Wired in supplier order view to 
  - load supplier order data from db
  - place order

Added getPossibleOrderLinesForSupplier function
  - Allows fetching order lines for a specific supplier
  - Similar to existing function but with supplier-specific filtering

### DIFF
--- a/apps/web-client/src/lib/db/orders/suppliers.ts
+++ b/apps/web-client/src/lib/db/orders/suppliers.ts
@@ -62,6 +62,22 @@ export async function getPossibleSupplerOrderLines(db: DB): Promise<SupplierOrde
 	);
 	return result;
 }
+
+export async function getPossibleOrderLinesForSupplier(db: DB, supplierId: number): Promise<SupplierOrderLine[]> {
+	// We need to build a query that will yield all books we can order, grouped by supplier
+	const result = await db.execO<SupplierOrderLine>(
+		`SELECT supplier_id, supplier.name as supplier_name, book.isbn, SUM(quantity) as quantity
+      FROM supplier
+        JOIN supplier_publisher ON supplier.id = supplier_publisher.supplier_id
+        JOIN book ON supplier_publisher.publisher = book.publisher
+        JOIN customer_order_lines ON book.isbn = customer_order_lines.isbn
+      WHERE quantity > 0 AND placed is NULL AND supplier_id = ?
+      GROUP BY supplier_id, book.isbn
+      ORDER BY book.isbn ASC;`,
+		[supplierId]
+	);
+	return result;
+}
 export async function getPossibleSupplerOrderInfos(db: DB): Promise<(SupplierOrderInfo & { supplier_name: string })[]> {
 	const result = await db.execO<SupplierOrderInfo & { supplier_name: string }>(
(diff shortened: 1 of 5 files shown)


apps/web-client/src/routes/orders/suppliers/[id]/+page.ts
```diff
@@ -1 +1,19 @@
...
+	const isbns = lines.map((book) => book.isbn);
+	const bookData = (await ordersDb.execO(`SELECT * FROM book WHERE isbn IN (${isbns.join(", ")})`)) as BookEntry[];
+
+	const bookDataMap = new Map<string, BookEntry>();
+	bookData.forEach((book) => {
+		bookDataMap.set(book.isbn, book);
+	});
+
+	const supplierOrderLinesWithData = lines.map((line) => ({ ...line, ...bookDataMap.get(line.isbn) }));
+	return { lines, books: supplierOrderLinesWithData, bookDataMap };
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/[id]/+page.ts:17]: In the view `data.books` is accessed like a map at one point `data.books[isbn]`. I was a little confused if "books" meant book data or order_line data. I would use `orderLines` or `orderLinesWithBookData` to distinguish it from the idea of a book entity

apps/web-client/src/routes/orders/suppliers/[id]/+page.svelte
```diff
@@ -90,28 +17,32 @@
 		lastUpdated: new Date()
 	};
 
-	let selectedBooks: string[] = [];
+	let selectedBooks: string[] = data?.books.map((book) => book.isbn) ?? [];
+
 	$: totalAmount = selectedBooks.reduce((acc, isbn) => {
-		const book = books.find((b) => b.isbn === isbn);
+		const book = data?.books[isbn];
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/[id]/+page.svelte:23]: Accessing the `books`/`orderLine` array with "isbn: string" won't give us the result we need here? If we return book data in the query then we can just access `price` directly

delasous (changes requested):
Could you please a description summarising the intent of the pr

apps/web-client/src/lib/db/orders/suppliers.ts
```diff
@@ -62,6 +62,22 @@ export async function getPossibleSupplerOrderLines(db: DB): Promise<SupplierOrde
...
+
+export async function getPossibleOrderLinesForSupplier(db: DB, supplierId: number): Promise<SupplierOrderLine[]> {
+	// We need to build a query that will yield all books we can order, grouped by supplier
+	const result = await db.execO<SupplierOrderLine>(
+		`SELECT supplier_id, supplier.name as supplier_name, book.isbn, SUM(quantity) as quantity
+      FROM supplier
+        JOIN supplier_publisher ON supplier.id = supplier_publisher.supplier_id
+        JOIN book ON supplier_publisher.publisher = book.publisher
+        JOIN customer_order_lines ON book.isbn = customer_order_lines.isbn
+      WHERE quantity > 0 AND placed is NULL AND supplier_id = ?
```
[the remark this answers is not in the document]
delasous: We agreed that supplier_order_lines should have a quantity but it should not be editable. The price for a row will need to be calculated as `SUM(supplier_order_line.quantity * book.price), 0) AS total_price`

PR #640, opened 2024-12-16
Db refactor/setup

ikusteu:
This is a setup for the [db refactor work](https://github.com/librocco/librocco/discussions/603).

Renames the schema `orders.sql` -> `init.sql` (as it will be used for both sides of the DB -- single init)

Moves the `db/orders` -> `db/cr-sqlite` (in `web-client`) as the same folder will be used to house interactivity for both DBs.


ikusteu:
_note: looking into failing CI, after that it's ready to go in (will probably YOLO it as the rest of the work depends on it)_

PR #652, opened 2024-12-26
Feature/wire reconcile list

### LINKED ISSUE #595: Wiring up order reconciliation: create and list
Reconciliation is the process of checking delivered isbn's against the isbn's in supplier order(s). A `reconciliation_order` is an "in process" reconciliation attempt. It It needs to track:
- the supplier orders that are being reconciled
- the books ("isbn"s) have been delivered
- when it was created
- when it was last updated
- its status (finalised or not)

A user needs to be able to:
1. See a list of all "in progress" reconciliations
2. Edit an ongoing reconciliation. This has 3 steps: add delivered isbn's; compare delivered vs ordered; finalise (updates customer orders)
3. Create a new reconciliation from "ordered" supplier order(s)

Scope:
1. There are existing query-handlers in the web-client `lib/db/reconciliation.ts`, and there are mock views under `routes/suppliers`. Mocked data and functions should be removed from the views, and the query/handlers should be setup to load and modify the tables. JSdoc comments should be added to query-handlers and the file to explain the context and their action.
2. This task will only concern itself with:
    1. creating a new reconciliation by selecting "ordered" supplier_order(s)
    2. listing existing/in-progrress reconciliations
    3. viewing an in-progress reconciliation and adding "delivered" isbn's to it
3. It will ignore: 
    1. comparing & finalising orders
    2. edge cases. A note will be made of any challenging scenario's and they will be considered in separate tasks. One such issue is: when an "ordered" supplier_order has been selected for reconciliation, what should happen to it so that it cannot be selected for an additional process

Implementation / Considerations:
1. New reconciliations are created by selecting multipe "ordered" supplier_orders and clicking "reconcile". Their id's are passed to `createReconciliationOrder` and we redirect to the reconcile view
2. The reconcile view currently sits at `/suppliers/reconcile`. There either needs to be another path segment to include the id, or a query param should be used
3. `reconciliation_order` currently doens't have an `updated_at` column. This should be added, and it should be updated when "delivered isbns" are added. 
4. The `reconcilation_order` table currently has a `customer_order_line_ids` json array. I think this should be updated to `delievered_isbns`. Does the current query-handler `addOrderLinesToReconciliationOrder` push new items to this array, or overwrite it each time? This method will be called each time a row is added to the "delivered" table (scanned or manual isbn input)
5. `getAllReconciliationOrders` should be added to the `suppliers/+layout.ts` to load the list that is shown under the "received" tab. This name should be reconsidered?
6. `getReconciliationOrder` should be called when loading the reconciliation view, and the data passed in place of the mocks. This mocked view currently lists the names of the suppliers alongside the supplier order ids. The query should be updated to return this.

fadwamahmoud:
Fixes  #595 
List and create reconciliation orders wire in


apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -9,6 +9,14 @@
...
+	import type { PageData } from "./$types";
+	import { addOrderLinesToReconciliationOrder } from "$lib/db/cr-sqlite/order-reconciliation";
+	import { page } from "$app/stores";
+	import { onDestroy, onMount } from "svelte";
+	import { invalidate } from "$app/navigation";
+
+	// implement order reactivity/sync
+	export let data: PageData;
 	// Mock data for the comparison view
 	const mockSupplierBooks = [
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:21]: There is still mock data being used here. The UI displays the name & id of each supplier order in this reconciliation. We can either make updates to get this data, or challenge the mocks - maybe we don't need the name, we could just display a list of ids, and make them links so a user can easily navigate to and see the details of included supplier orders. We can make a call in this PR, or write up a small enhancement ticket

apps/web-client/src/lib/db/cr-sqlite/types.ts
```diff
@@ -86,8 +86,8 @@ export type SupplierOrder = {
  */
 export type ReconciliationOrder = {
 	SupplierOrderIds: number[];
-	created: Date;
-	customer_order_line_ids: string[]; // isbns
+	created: number;
```
delasous, line 89: Also add updated_at

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -124,12 +158,14 @@
...
 
 						<div class="flex flex-row items-center justify-between gap-y-2 md:flex-col md:items-start">
-							<h2 class="prose">#{reconciliation.id}</h2>
+							<h2 class="prose">#{data?.reconciliationOrder.id}</h2>
 
 							<span class="badge-accent badge-outline badge badge-md gap-x-2 py-2.5">
 								<span class="sr-only">Last updated</span>
 								<ClockArrowUp size={16} aria-hidden />
-								<time dateTime={reconciliation.lastUpdated.toISOString()}>{reconciliation.lastUpdated.toLocaleDateString()}</time>
+								<!-- <time dateTime={data?.reconciliationOrder.created.toISOString()}
```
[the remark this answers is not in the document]
Comment (delasous) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:166]: Show `updated_at` here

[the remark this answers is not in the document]
Comment (delasous) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:166]: Show `created` as a separate fact

delasous (changes requested):

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { BookEntry } from "@librocco/db";
+import type { PageLoad } from "./$types";
+
+export const load: PageLoad = async ({ parent, params, depends }) => {
+	depends("reconciliationOrder:data");
+
+	const { ordersDb, dbCtx } = await parent();
+	const reconciliationOrder = await getReconciliationOrder(ordersDb, parseInt(params.id));
+
+	const books: string[] = JSON.parse(reconciliationOrder.customer_order_line_ids) || [];
```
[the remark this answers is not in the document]
Comment (delasous) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts:11]: `getReconciliationOrder` should parse the supplier order ids and throw an error if its invalid json. I don't think it should return empty [] as default, as then we aren't reconciling against any orders?

apps/web-client/src/lib/components/supplier-orders/ComparisonTable.svelte
```diff
@@ -14,12 +19,16 @@
...
 
-	export let supplierBooks: SupplierBooks[];
+	export let supplierBooks: ProcessedOrderLine[];
 
-	$: getSupplierSummary = (books: Book[]) => {
+	$: sortedSupplierBooks = sortLinesBySupplier(supplierBooks);
+	// list all books by isbn
+	// // only show one row per isbn
+	//
+	$: getSupplierSummary = (books: (BookEntry & { delivered: boolean })[]) => {
```
delasous, line 31: Remove unused imports & locally defined types

apps/web-client/src/lib/components/supplier-orders/ComparisonTable.svelte
```diff
@@ -14,12 +19,16 @@
 		books: Book[];
 	};
 
-	$: totalDelivered = supplierBooks.reduce((acc, supplier) => acc + supplier.books.filter((b) => b.delivered).length, 0);
-	$: totalOrdered = supplierBooks.reduce((acc, supplier) => acc + supplier.books.length, 0);
+	$: totalDelivered = supplierBooks.length;
```
delasous, line 22: Remove unused imports & locally defined types

apps/web-client/src/lib/db/cr-sqlite/utils.ts
```diff
@@ -0,0 +1,63 @@
+import type { BookEntry } from "@librocco/db";
+
+export type ProcessedOrderLine = ({ supplier_name: string } & BookEntry) & {
+	delivered: boolean;
+	wasOrdered: boolean;
+	orderedQuantity: number;
+	deliveredQuantity: number;
+	remainingQuantity: number;
+};
+export const processOrderDelivery = (
```
delasous, line 10: I think this could be rebuilt slightly:
1. We only use `orderedQuantity` and `deliveredQuantity` in the ComparisonTable. This means we can drop `delivered` and `remainingQuantity`.
2. I would like us to emphasise this list of unmatched/un-ordered books, instead of it just being an unlabelled supplier group in the same table => I think we should return this as a separate list. This means we can won't return unmatched-procssedLines with `supplier_name=""`,  we can drop `wasOrdered`, and the function would return `{ processedLines: ProcessedLine[]; unmatchedBooks: BookEntry[](with quantity) }`
3. The current implementation iterates `scannedBooks` 3 times - it doesn't need to. If we need a map of isbn->quantities, and a map of isbn->bookEntry, we can set them in the same iteration of scannedBooks
4. Instead of iterating `scannedBook`s at the end to figure out the unmatched result, we could do some book-keeping while iterating `placedOrderLines` to figure out what's left.

apps/web-client/src/lib/db/cr-sqlite/utils.ts
```diff
@@ -0,0 +1,63 @@
+import type { BookEntry } from "@librocco/db";
+
+export type ProcessedOrderLine = ({ supplier_name: string } & BookEntry) & {
```
delasous, line 3: Move to shared types file

apps/web-client/src/lib/db/cr-sqlite/utils.ts
delasous, line 1: I think these could be moved to the main `order-reconciliation` file for now. I know they don't query the db, but they are pretty core to the reconcilation fow logic, and in `utils` I feel like they're a little lost.

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -57,20 +65,38 @@
...
+		const { rx } = data.dbCtx;
+
+		const disposer1 = rx.onPoint("reconciliationOrder", BigInt($page.params.id), () => invalidate("reconciliationOrder:data"));
+		const disposer2 = rx.onRange(["reconciliation_order"], () => invalidate("reconciliationOrder:data"));
+		disposer = () => (disposer1(), disposer2());
+	});
 
+	//#endregion reactivity
+
+	onDestroy(async () => {
```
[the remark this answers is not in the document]
Comment (delasous) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:33]: To be more concrete in my feedback here:
1. don't debounce right now
2. insert into reconciliation_order_line table on every scan. See `note.ts > addVolumesToNote` for quantity conflict update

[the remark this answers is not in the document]
Comment (fadwamahmoud) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:33]: will create a separate ticket for including quantity in reconciliation order lines table

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
```diff
@@ -1,64 +1,81 @@
...
 
 	if (!reconOrder[0]) {
 		throw new Error(`Reconciliation order ${id} not found`);
 	}
 
-	await db.exec(
-		`UPDATE reconciliation_order SET customer_order_line_ids = (json_array(${multiplyString("?", isbns.length)})) WHERE id = ?;`,
-		[...isbns, id]
-	);
+	const params = isbns.map((isbn) => [id, isbn]).flat();
```
delasous, line 56: I think we should work with an array of `{isbn, quantity}[]` for the "new lines" parameter. We would need to update this map, remove the fixed `1` quantity below in `VALUES (...)` to reference the mapped quantity, and update `quantity + 1` to reference `excluded.quantity`. This gives us the foundation to take another pass at your debounce logic in the reconcilation view

apps/web-client/src/lib/schemas/index.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+import type { Infer } from "sveltekit-superforms";
+
+export type SettingsSchema = Infer<typeof settingsSchema>;
```
delasous, line 4: Move this back to `../forms/schema.ts`

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts
```diff
@@ -0,0 +1,47 @@
...
+
+	const supplierOrders = placedOrderLines.reduce<{ [supplier_order_id: string]: { supplier_name: string; supplier_id: string } }>(
+		(acc, val) =>
+			acc[val.supplier_order_id]
+				? acc
+				: { ...acc, [val.supplier_order_id]: { supplier_name: val.supplier_name, supplier_id: val.supplier_id } },
+		{}
+	);
+
+	const books = reconciliationOrderLines.reduce(
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts:25]: `getReconciliationOrderLines` should get the book data, similar to how we get supplier order lines

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { BookEntry } from "@librocco/db";
+
+export const load: PageLoad = async ({ parent, params, depends }) => {
+	depends("reconciliationOrder:data");
+
+	const { ordersDb } = await parent();
+	const reconciliationOrder = await getReconciliationOrder(ordersDb, parseInt(params.id));
+	const reconciliationOrderLines = await getReconciliationOrderLines(ordersDb, parseInt(params.id));
+
+	const supplierOrderIds = JSON.parse(reconciliationOrder.supplier_order_ids) || [];
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts:13]: `getReconciliationOrder` should parse this JSON and throw if its invalid. The `[]` default should not be returned

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts
```diff
@@ -0,0 +1,47 @@
...
+
+	const placedOrderLines = await getPlacedSupplierOrderLinesForReconciliation(ordersDb, supplierOrderIds);
+
+	const supplierOrders = placedOrderLines.reduce<{ [supplier_order_id: string]: { supplier_name: string; supplier_id: string } }>(
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts:17]: This map is not necessary, at least from how its consumed in the UI

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -7,95 +7,54 @@
...
-	let books: Array<{
-		isbn: string;
-		title: string;
-		authors: string;
-		price: number;
-		quantity: number;
-	}> = [];
+	$: books = data?.mergedBookData || [];
 
+	let isbns = [];
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:40]: Seems to be unused?

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -7,95 +7,54 @@
...
 	// Mock supplier orders data
-	const selectedOrders = [
-		{ id: 1, supplier: "Academic Books Ltd", books: 5 },
-		{ id: 2, supplier: "Penguin Random House", books: 3 }
-	];
 
-	function handleIsbnSubmit() {
+	$: scanned = data?.placedOrderLines;
+
+	async function handleIsbnSubmit() {
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:46]: Remove global `isbn` variable and pass it as a function param

apps/web-client/src/lib/db/cr-sqlite/utils.ts
delasous: I have moved comments around `processOrderDelivery` updates to a new issue

delasous (changes requested):
I have moved comments concerning order comparison (in ComparisonTable & processOrderDelivery) to new issues. I have added some new comments concerning `reconciliation_order_lines` handlers & the script block of the reconciliation view.

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts
Comment (fadwamahmoud) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts:13]: there is a check in the db for invalid json. will remove the [] though

apps/web-client/src/lib/schemas/index.ts
fadwamahmoud: why? importing from forms causes a loading error on orders/supplier routes

delasous: Because then we have two schema files. It works if its imported via `$lib/forms/schemas` (so the export from `forms/index` can be removed)

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts
Comment (delasous) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts:13]: The call to `JSON.parse` doesn't know what has happened elsewhere. It should still be handled properly. It makes the most sense for it to happen in the handler, where other cases are being handled

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
fadwamahmoud: what about the reconciliation_order_id?

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts
```diff
@@ -0,0 +1,19 @@
...
+import { getPlacedSupplierOrderLinesForReconciliation } from "$lib/db/cr-sqlite/suppliers";
+
+export const load: PageLoad = async ({ parent, params, depends }) => {
+	depends("reconciliationOrder:data");
+
+	const { ordersDb } = await parent();
+	const reconciliationOrder = await getReconciliationOrder(ordersDb, parseInt(params.id));
+	const reconciliationOrderLines = await getReconciliationOrderLines(ordersDb, parseInt(params.id));
+
+	const supplierOrderIds = JSON.parse(reconciliationOrder.supplier_order_ids);
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.ts:12]: Return the parsed result from `getReconciliationOrder`?

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -7,95 +7,55 @@
...
+		validators: zod(scannerSchema),
+		validationMethod: "submit-only"
+	});
+
+	const { form: formStore } = form;
+
+	$: placedOrderLines = data?.placedOrderLines;
+	$: totalDelivered = data?.reconciliationOrderLines.map((book) => book.quantity).reduce((acc, curr) => acc + curr, 0);
+	$: totalOrdered = placedOrderLines.length;
+	$: processedOrderDelivery = processOrderDelivery(data?.reconciliationOrderLines, data?.placedOrderLines);
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:58]: This will re-run ever time we add new lines. This is a downside of having all of this logic packed in one view. A slight improvement would be to move it into the if block where the ComparisonTable is rendered as a `{@const processedOrderDeliver = ...}` declaration

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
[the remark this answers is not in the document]
delasous: Please add some documentation comments

apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
```diff
@@ -84,22 +83,49 @@ describe("Suppliers order creation", () => {
...
 
 		expect(res2).toMatchObject({
-			customer_order_line_ids: null,
 			id: 1,
 			supplier_order_ids: "[1]",
 			finalized: 0
 		});
 
-		await addOrderLinesToReconciliationOrder(db, 1, ["123", "435", "324"]);
+		await addOrderLinesToReconciliationOrder(db, 1, [
```
delasous, line 91: Do we have a case to check that existing order line quantity will be updated on conflict?

delasous:
I have contributed a few small updates:
- fixed the superforms setup of the scanner form in the reconciliation page (this will use the existing component when the inventory app is reskined)
- fixed the settings store setup that was causing inventory pages to crash
- added more documentation on the reconciliation process in the db file (to be built into more comprehensive docs)

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -212,23 +190,27 @@
...
 										</tr>
 									{/each}
 								</tbody>
 							</table>
 						</div>
 					{/if}
 				{:else if currentStep > 1}
-					<ComparisonTable supplierBooks={mockSupplierBooks} />
+					{@const processedOrderDelivery = processOrderDelivery(data?.reconciliationOrderLines, data?.placedOrderLines)}
+					}
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:212]:
```suggestion
					
```

PR #654, opened 2024-12-27
Set up reconciliation finnalization test

fadwamahmoud:
Reconciliation finalization
- [ ]  test setup

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
@@ -195,4 +195,33 @@ describe("Suppliers order creation", () => {
 			);
 		});
 	});
+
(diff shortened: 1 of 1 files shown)


apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
```diff
@@ -195,4 +195,33 @@ describe("Suppliers order creation", () => {
 			);
 		});
 	});
+
+	describe("Reconciliation order finalization", () => {
+		let db: DB;
```
silviot, line 200: The line `let db: DB` and the first line in `beforeEach` (`db = await getRandomDb();`) can be removed: the `describe("Reconciliation order finalization", ...)` directive is nested under `describe("Suppliers order creation", ...)`, so it inherits its `beforeEach` that already takes care of db initialization.

PR #648, opened 2024-12-21
Feature/supplier order view

fadwamahmoud:
Create supplier order details view
Fixes #599

PR #662, opened 2025-01-04
Compare delivered and ordered books

### LINKED ISSUE #653: Wiring up order reconciliation: Comparing and finalizing
Comparing Orders                                                                                               

  -  comparison between delivered ISBNs and ordered ISBNs from supplier orders               
  -  The comparison should identify:                                                                             
  -  Which ordered books have been delivered                                                                  
  - Which delivered books weren't in the original orders                                                     
  -  Any discrepancies in quantities 

       
Finalizing:
  - finalizeReconciliationOrder

fadwamahmoud:
WIP: Reconciliation order comparison against ordered lines
Fixes #653

### DIFF
--- a/apps/e2e/integration/reconciliation_order.spec.ts
+++ b/apps/e2e/integration/reconciliation_order.spec.ts
@@ -1,6 +1,6 @@
 import { test, expect } from "@playwright/test";
 
-import { baseURL } from "../constants";
+import { baseURL } from "./constants";
 test.beforeEach(async ({ page }) => {
 	await page.goto(baseURL);
 	page.getByLabel("Main navigation");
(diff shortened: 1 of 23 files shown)


delasous (commented):
It is difficult to evaluate what should be reviewed here. It is based off of work in #652 which I have left comments on, and is being merged back into 652, adding new things but also fixing (some) parts of the work that needs to be done there. It also seems to include code from #648, which needed work. 

I think part of the problem is that it trying to build new logic (the reconciliation comparison) and wire it in at the same time. I think we should sit down together and sketch out the entire reconciliation picture based off of what is being built up in each of these PRs, then isolate small bits of work. 

My feeling is that this PR should only include work in `/cr-sqlite/utils.ts` and its unit tests, and that it should be directed at the main branch. The rest can be isolated into its own units.

Issue #666, opened 2025-01-09
Configure hash routing


delasous:
Sveltekit now allows us to [configure hash routing](https://svelte.dev/docs/kit/configuration#router). This should fix #580? This should be easy to implement but we'll probably need to upgrade Sveltekit first, which may mean updating Svelte, which may mean... We should assess knock-on dependency effects before committing too much time to this, but also schedule a little upgrade party (as well as familiarise ourselves with svelte 5) soon.

ikusteu:
Mentining this here as well: #580 as it might be fixed with hash (client side) routing
@chiruzzimarco if that's the case, please mention the issue in PR so as to close it.

PR #673, opened 2025-01-14
Implement hash routing

chiruzzimarco:
This PR will upgrade svelte, svelte kit and related dependencies in order to implement hash based routing.

Related to issue #666 .

### DIFF
--- a/apps/e2e/constants.ts
+++ b/apps/e2e/constants.ts
@@ -22,6 +22,31 @@ export function getPort(): Promise<number> {
 
 const port = await getPort();
 export const baseURL = `http://localhost:${port}/preview/`;
+export const HASHES = {
+	debug: "#/debug/",
+	// Inventory mgmt
+	stock: "#/stock/",
+	books: "#/books/",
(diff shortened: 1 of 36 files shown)


chiruzzimarco:
I had to add the [`skipLibCheck`](https://www.typescriptlang.org/tsconfig/#skipLibCheck) option to the typescript options in order to avoid the following exception during `rush typecheck`:

```
/home/runner/work/librocco/librocco/apps/web-client/../../common/temp/node_modules/.pnpm/@types+node@22.10.6/node_modules/@types/node/buffer.d.ts:130:14
Error: Cannot export 'Buffer'. Only local declarations can be exported from a module. 
    export function resolveObjectURL(id: string): Blob | undefined;
    export { Buffer };
  ...2069 lines omitted...
```

This issue may be due to the presence of pouchdb [as outlined in this issue](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/62134). Removing pouchdb as a dependency turned out to be quite a challenge at this stage of development as many tests and parts of the codebase will break as a result.

PR #676, opened 2025-01-17
Fix/db timestamps

### LINKED ISSUE #647: Flaky customer sync test
The customer orders sync test was failing before and it was fixed. However, now it's flaky. See this screenshot:

Screenshot 2024-12-20 at 12 39 15

What I read from this is:
- customer order entries in different DBs are matched against each other
- their `updatedAt`s don't **always** match up
- sometimes they do match up as: 
  - the db timestamps on every second
  - sometimes the update and sync happen on the same second - test is fine
  - sometimes the sync happens just at the start of the next second - resulting in a mismatched timestamp
- **my assumption is this:** sync also updates the table, triggering `updatedAt` - if I'm mistaken, this whole line of reasoning is more/less worthless (don't have the time to get deep into it atm, so whoever picks this up should investigate)

If my assumption (and line of reasoning) is correct, the possible solutions are:
- do we want the sync to trigger `updatedAt` update - if not, block it
- if we do want the sync to trigger `updatedAt` - make sure the tests ignore the `updatedAt` field when comparing synced DBs

ikusteu:
Fixes #647 

Removes all DB timestamp triggers and explicitly sets the timestamp (with ms-precision) for both orders and inventory sides of the DB

Includes #650

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
@@ -103,7 +103,7 @@ describe("Customer order tests", () => {
 describe("Customer order tests", () => {
 	let db1: DB, db2: DB;
 	beforeEach(async () => ([db1, db2] = await getRandomDbs()));
-	it("Should sync customer creation", async () => {
+	it.only("Should sync customer creation", async () => {
 		// We create one customer in db1 and a different one in db2
 		let db1Customers: Customer[], db2Customers: Customer[];
 		await upsertCustomer(db1, { fullname: "John Doe", id: 1, email: "john@example.com", deposit: 13.2 });
@@ -111,9 +111,13 @@ describe("Customer order tests", () => {
 		[db1Customers, db2Customers] = await Promise.all([getAllCustomers(db1), getAllCustomers(db2)]);
 		expect(db1Customers.length).toBe(1);
 		expect(db2Customers.length).toBe(1);
+
+		await new Promise((resolve) => setTimeout(resolve, 1000));
+
 		await syncDBs(db1, db2);
 		expect((await getAllCustomers(db2)).length).toBe(2);
 		await syncDBs(db2, db1);
+
 		expect((await getAllCustomers(db1)).length).toBe(2);
 		[db1Customers, db2Customers] = await Promise.all([getAllCustomers(db1), getAllCustomers(db2)]);
 		expect(db1Customers).toMatchObject(db2Customers);
--- a/apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
@@ -114,6 +114,8 @@ describe("Customer order tests", () => {
 		expect(db1Customers.length).toBe(1);
 		expect(db2Customers.length).toBe(1);
 
+		// This tests for a regression we had: we want to ensure that the sync won't update the `updated_at` field
+		// - it should be the same as for the original entry
 		await new Promise((resolve) => setTimeout(resolve, 1000));
 
 		await syncDBs(db1, db2);
--- a/apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
@@ -102,8 +102,10 @@ describe("Customer order tests", () => {
 
 describe("Customer order tests", () => {
 	let db1: DB, db2: DB;
+
 	beforeEach(async () => ([db1, db2] = await getRandomDbs()));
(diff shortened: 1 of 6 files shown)


apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
```diff
@@ -111,13 +113,18 @@ describe("Customer order tests", () => {
 		[db1Customers, db2Customers] = await Promise.all([getAllCustomers(db1), getAllCustomers(db2)]);
 		expect(db1Customers.length).toBe(1);
 		expect(db2Customers.length).toBe(1);
+
+		await new Promise((resolve) => setTimeout(resolve, 1000));
+
```
silviot, line 118: For me the test passes even without the wait.
I think we should never wait a constant amount of time in tests, but rather wait for a specific thing to come true.
```suggestion
```

apps/web-client/src/lib/db/cr-sqlite/customers.ts
```diff
@@ -31,27 +31,32 @@
...
            deposit = COALESCE(?, deposit);`,
 		[
 			customer.id,
 			customer.fullname ?? null,
 			customer.email ?? null,
 			customer.deposit ?? null,
+			timestamp,
 			customer.fullname ?? null,
 			customer.email ?? null,
 			customer.deposit ?? null
```
silviot, line 62: In case of update we want `updated_at` to get a new value.

```suggestion
           updated_at = ?,
           deposit = COALESCE(?, deposit);`,
		[
			customer.id,
			customer.fullname ?? null,
			customer.email ?? null,
			customer.deposit ?? null,
			timestamp,
			customer.fullname ?? null,
			customer.email ?? null,
			customer.deposit ?? null
			timestamp
```

apps/web-client/src/lib/db/cr-sqlite/note.ts
```diff
@@ -292,15 +297,19 @@ export async function addVolumesToNote(db: DB, noteId: number, volume: VolumeSto
...
 			quantity = book_transaction.quantity + excluded.quantity,
-			updated_at = (strftime('%s', 'now') * 1000)
+			updated_at = excluded.updated_at
 	`;
 
-	await db.exec(insertOrUpdateTxnQuery, [isbn, quantity, warehouseId, noteId]);
+	await db.tx(async (txDb) => {
+		const timestamp = Date.now();
+		await txDb.exec(insertOrUpdateTxnStmt, [isbn, quantity, warehouseId, noteId, timestamp]);
+		await txDb.exec(`UPDATE note SET updated_at = ? WHERE id = ?`, [timestamp, noteId]);
```
silviot, line 311: I suggest we don't use `excluded`, but repeat the `timestamp` on line 310 to make sure `updated_at` gets a new value on update.

apps/web-client/src/lib/db/cr-sqlite/note.ts
```diff
@@ -361,19 +370,27 @@ export async function updateNoteTxn(
 		return;
 	}
 
+	// We're deleting the existing txn and placing the new one instead of it
+	// in case of mere update, we're keeping the original `updated_at`
```
silviot, line 374: I think we should have `updated_at` get a new value on update. Or change its name to `created_at`, if we really want this behavior.

apps/web-client/src/lib/db/cr-sqlite/note.ts
```diff
@@ -397,25 +417,37 @@ export async function upsertNoteCustomItem(db: DB, noteId: number, payload: { id
 
 	const { id, title, price } = payload;
 
+	const timestamp = Date.now();
+
+	// In case of conflict (update) we're keeping the initial 'updated_at' (to keep consistent display order)
```
silviot, line 422: Let's discuss this.

silviot (changes requested):

apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
ikusteu: @chiruzzimarco did this to ensure the test won't fail even if we don't sync immediately. I've added a comment explaining this.

apps/web-client/src/lib/db/cr-sqlite/note.ts
ikusteu: I'm perfectly ok with this, but can you elaborate on the shortcomings of using `excluded.updated_at`?

apps/web-client/src/lib/db/cr-sqlite/note.ts
ikusteu: Regarding the `updated_at` / `last_bubbled_at`, I say: let's merge this, as #677 depends on it, and, I would much rather have it in (with working UI, accurate tests and all) before I can be sure I didn't skip / mess something up in the process. I've opened an issue: #690

silviot (approved):

PR #684, opened 2025-01-22
Improve processOrderDelivery logic

### LINKED ISSUE #678: Improve "delivered vs ordered" comparison
I think the current implementation of "delivered vs ordered" comparison logic could be improved.

**Generally:**
-  The two functions that handle this process are written in `/db/cr-sqlite/utils.ts`. I think these could be moved to the main order-reconciliation file for now. I know they don't query the db, but they are pretty core to the reconcilation fow logic, and in utils I feel like they're a little lost.
- The types defined in the file should be moved to shared db types `/db/cr-sqlite/types`

**Rebuilding `processOrderDelivery`**
There are two reasons for this: 
- line 39 iterates scanned books... while iterating scanned books
- the returned result should reflect what we want to show in the UI. There are two main things to make the user aware of here: unmatched isbns (books that were never ordered), and over & under-ordered isbns (especially extra quantities). I would like us to emphasise the former, potentially outside of the comparison table, definitely at the top of the view. The latter can be shown in the comparison table - over-orders can potentially be highlighted by a change in the row/cell colour. This issue should just concern itself with updates to the returned type from `processOrderDelivery` - updates to the UI will be handled in a separate issue.

I propose the following changes:
1. We only use `orderedQuantity` and `deliveredQuantity` in the ComparisonTable. This means we can drop `delivered` and `remainingQuantity` from the result
2. Return unmatched books as a separate list, instead of it just being an unlabelled supplier group in the same table. This means we won't return unmatched-processedLines with `supplier_name=""`,  we can drop `wasOrdered`, and the function would return `{ processedLines: ProcessedLine[]; unmatchedBooks: BookEntry[](with quantity) }`
3. Restructure the function logic. Some considerations:
     1. The current implementation iterates `scannedBooks` 3 times - it doesn't need to. If we need a map of isbn->quantities, and a map of isbn->bookEntry, we can set them in the same iteration of scannedBooks
    2. Instead of iterating `scannedBook`s at the end to figure out the unmatched result, we could do some book-keeping while iterating `placedOrderLines` to figure out what's left.
    3. It might make more sense to flip it, so that we iterate scannedBooks and check them against a map of orderedBooks & quantity (supplierOrderLines)? Again, there are two main responsibilites: unmatched books, and delivered quantity => if scannedBooks doesnt find a corresponding orderedEntry, it gets pushed to unmatched array; if its in ordered map, we update deliveredQuantity...
4. Update tests

fadwamahmoud:
Fixes #678


apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
```diff
@@ -202,3 +203,61 @@ export async function finalizeReconciliationOrder(db: DB, id: number) {
...
+): { processedLines: ProcessedOrderLine[]; unmatchedBooks: (BookEntry & { quantity: number })[] } => {
+	const scannedLinesMap = new Map<string, BookEntry & { quantity: number }>(
+		scannedBooks.map((book) => {
+			return [book.isbn, book];
+		})
+	);
+	const result = { processedLines: [], unmatchedBooks: [] };
+
+	for (const placedOrderLine of placedOrderLines) {
+		if (scannedLinesMap.has(placedOrderLine.isbn)) {
```
delasous, line 234: This is looking good. However, I had a click through and noticed one case which isn't handled as I would have expected: an "under delivered" quantity of 0 where there is an ordered quantity (In the mock data you have set up isbn 123456789 appears in the "unmatched books" table). I think in this case the row should be reported in its supplier order group as Ordered Quantity = X, Delivered Quantity = 0. To handle this, it might be better to start by iterating scannedOrderLines & check against a placedOrderLines map?

apps/web-client/src/lib/components/supplier-orders/ComparisonTable.svelte
```diff
@@ -1,14 +1,20 @@
...
+		unmatchedBooks: []
+	};
+
+	$: sortedSupplierBooks = sortLinesBySupplier(supplierBooks.processedLines);
 
-	$: sortedSupplierBooks = sortLinesBySupplier(supplierBooks);
-	// list all books by isbn
-	// // only show one row per isbn
-	//
+	onMount(() => {
```
delasous, line 15: Remove

apps/web-client/src/lib/components/supplier-orders/ComparisonTable.svelte
```diff
@@ -1,14 +1,20 @@
 <script lang="ts">
-	import { sortLinesBySupplier, type ProcessedOrderLine } from "$lib/db/cr-sqlite/utils";
+	import { sortLinesBySupplier } from "$lib/db/cr-sqlite/order-reconciliation";
+	import type { ProcessedOrderLine, SupplierPlacedOrderLine } from "$lib/db/cr-sqlite/types";
 
 	import type { BookEntry } from "@librocco/db";
+	import { onMount } from "svelte";
 
-	export let supplierBooks: ProcessedOrderLine[];
+	export let supplierBooks: { processedLines: ProcessedOrderLine[]; unmatchedBooks: (BookEntry & { quantity: number })[] } = {
```
delasous, line 8: Might be better to rename it `reconciledBooks` or  `processedOrderLines` - something to relate it to the processing in the reconciliation flow

apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
```diff
@@ -253,3 +257,179 @@ describe("Reconciliation order creation", () => {
...
+					price: 10,
+					quantity: 2,
+					title: "Book 1"
+				}
+			],
+			unmatchedBooks: []
+		});
+	});
+
+	it("should handle partial delivery", () => {
```
delasous, line 302: Add test case for the under-delivered 0 quantity case I mentioned below

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -167,10 +163,16 @@
 				</nav>
 
 				{#if currentStep === 1}
-					<form class="flex w-full gap-2" use:enhance method="POST">
+					<form
+						class="flex w-full gap-2"
+						on:submit|preventDefault={() => {
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:168]: This `on:submit` event handler is exposed by svelte, it is not connected to superforms. If we are going to use superforms, we should use it as I have set it up (diff on line 48 -> 57) otherwise we aren't gettting any benefits of the validation that occurs in their [event chain](https://superforms.rocks/concepts/events). The input should already be cleared on every successful submission - if its not behaving, there are other ways to fix it

delasous (changes requested):

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
fadwamahmoud: I could also add the placedOrderLine to the processedLines array (instead of the scannedBook to keep things consistent) and in the else block include a deliveredQuantity of 0

delasous: Not sure I follow... the else block would need to decide if its pushing to processLines or unmatchedBooks?

fadwamahmoud: take a look at my latest commit

delasous: Ok, makes sense. Only other two things I'd say:
- add some space. Lines 235 - 237 could use it, otherwise its hard to read the `scannedLinesMap.delete()` action, which is pretty important
- line 250:  `result.unmatchedBooks` doesn't need to be spread.

apps/web-client/src/lib/components/supplier-orders/ComparisonTable.svelte
delasous: ?

apps/web-client/src/lib/components/supplier-orders/ComparisonTable.svelte
delasous: ?

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
Comment (delasous) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:168]: ?

delasous (commented):
I've left a "?" on 3 changes that have been resolved with no discussion or changes?

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
Comment (delasous) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:168]: If you want to keep the input focussed after submit, see what I have done in `apps/web-client/src/routes/orders/customers/[id]/+page.svelte` of my `fix/customer-order-line` PR

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -50,10 +50,18 @@
 
 	const form = superForm(defaults(zod(scannerSchema)), {
 		validators: zod(scannerSchema),
-		validationMethod: "submit-only"
+		validationMethod: "submit-only",
+		onUpdated: async ({ form: { data, valid } }) => {
```
Comment (delasous) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:54]: Check out the change I made in my `fix/customer-order-lines` PR. Not essential, but it does improve the experience

delasous (approved):

PR #688, opened 2025-01-23
Fix/customer order lines

### LINKED ISSUE #593: Update customer order line states: Received & Delivered
Customer order lines currently transition through the following states:
- Draft: Has created timestamp (inital state)
- Placed: Has placed timestamp (order sent to supplier)
- Received: Has received timestamp (books arrived from supplier)
- Collected: Has collected timestamp (customer picked up books)

I propose that we update these so that the initial state is "Received", and the current received state becomes "Delivered".

My rational for this, is that "Draft" carries conotations of a working, uncommited state, but from the perspective of customer ordering nothing needs to be finalised - the line is already added to the pool received orders for that supplier.

### LINKED ISSUE #598: Remove the `customer_supplier_order` table
We don't need to keep track what customer order generated what supplier order. We can figure out what customer order to fullfill during the reconciliation.

See [the figjam diagram](https://www.figma.com/board/zhm2SSX7f74qP4gHLUXI7K/Librocco-Next-Steps?node-id=516-484&t=TvqIxUj6n7Zqciew-4) and/or the [source of the diagram on dbdocs](https://dbdocs.io/silviot/librocco).

### LINKED ISSUE #681: Book data is not loaded for customer order lines
- [x] The `example_data` does not  connect `book.isbn`s to `customer_order_lines`.isbn
- [x] Even when this is updated, book data is not populated per row => investigate why...?
  -  [x] The `customers/[id]` route sets up a `booksMap`, this is never returned and the `booksEntry` array that is returned is accessed like this map. This setup depends on a loose book data query in the load function that should be removed in favour of a join through `getCustomerBooks`
  - [x] `getCustomerBooks` works with the table that should be removed in #598

delasous:
- Updates `getCustomerOrderLine` handler to return joined book data. This is only needs to be a subset of book properties that are shown customer orders table. Removes logic in customer order load and view that handled this.
- Removes the Quantity col from the customer order table
- Tidys up "adding customer lines": the page scanner form setup & handler + db handler param type
- Removes mock customer orders route `/c`

Fixes #681
Fixes #598 
Fixes #593


apps/web-client/src/routes/orders/customers/+layout.ts
```diff
@@ -15,6 +15,8 @@ export const load: LayoutLoad = async ({ depends }) => {
 	const customers = await getAllCustomers(db);
 	const customerOrderLines = await getAllCustomerOrderLines(db);
 
+	console.log(customerOrderLines);
```
delasous, line 18:
```suggestion
```

apps/web-client/src/lib/db/cr-sqlite/types.ts
```diff
@@ -74,6 +49,7 @@ export type SupplierOrderLine = {
 	title: string;
 	authors: string;
 	publisher: string;
+	0.84;
```
silviot, line 52:
```suggestion
```

apps/web-client/src/lib/db/cr-sqlite/__tests__/customer-orders.test.ts
```diff
@@ -69,33 +85,33 @@ describe("Customer order tests", () => {
...
 
 	it("can remove books from a customer order", async () => {
 		await upsertCustomer(db, { fullname: "John Doe", id: 1 });
 
-		await addBooksToCustomer(db, 1, [{ isbn: "9780000000000" }, { isbn: "9780000000000" }]);
-		let books = await getCustomerBooks(db, 1);
+		await addBooksToCustomer(db, 1, ["9780000000000", "9780000000000"]);
+		let books = await getCustomerOrderLines(db, 1);
 		expect(books.length).toBe(2);
 		removeBooksFromCustomer(db, 1, [books[0].id]);
```
silviot, line 113:
```suggestion
		await removeBooksFromCustomer(db, 1, [books[0].id]);
```

silviot (approved):

PR #689, opened 2025-01-23
Fix/small orders fixes

### LINKED ISSUE #591: Orders with no books in them should be considered "In Progress", not "completed"
Currently a newly created customer order is shown as "complete". It has no pending book still to order, so it kind of makes sense... but should be considered a sepcial case.

### LINKED ISSUE #651: Get book data at the same time as the ISBN
There are a couple of queries in

* [apps/web-client/src/routes/c/[id]/+page.ts](https://github.com/librocco/librocco/blob/f2c530bdee648034aa66811e33ce96e4c37d3a44/apps/web-client/src/routes/c/%5Bid%5D/%2Bpage.ts#L24):24
* [apps/web-client/src/routes/orders/customers/[id]/+page.ts](https://github.com/librocco/librocco/blob/f2c530bdee648034aa66811e33ce96e4c37d3a44/apps/web-client/src/routes/orders/customers/%5Bid%5D/%2Bpage.ts#L24):24

where we first invoke `getCustomerBooks` and then we issue a query to get the book title, author etc. We should instead make `getCustomerBooks` return this data in the first place.

In general we should strive, when possible, to minimize the number of SQL queries needed to render a view.

### LINKED ISSUE #657: Email should not be a mandatory customer field
It should be optional.

ikusteu:
Fixes #591 
Fixes #651 
Fixes #657

There's poor test coverage of these updates (and related functionality)...I don't know if I'd missed something or if it's something we should add for 'down-the-line'

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/customers.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/customers.ts
@@ -79,16 +79,34 @@ export const getAllCustomerOrderLines = async (db: DB): Promise<DBCustomerOrderL
 
 export const getCustomerBooks = async (db: DB, customerId: number): Promise<CustomerOrderLine[]> => {
 	const result = await db.execO<DBCustomerOrderLine>(
-		`SELECT customer_order_lines.id, isbn, customer_id, created, placed, received, collected, GROUP_CONCAT(supplier_order_id) as supplierOrderIds
+		`SELECT
+			customer_order_lines.id,
+			customer_order_lines.isbn,
+			customer_order_lines.customer_id,
+			customer_order_lines.created,
+			customer_order_lines.placed,
+			customer_order_lines.received,
+			customer_order_lines.collected,
+			GROUP_CONCAT(customer_supplier_order.id) as supplierOrderIds,
+			COALESCE(book.title, 'N/A') AS title,
+			COALESCE(book.price, 0) AS price,
+			COALESCE(book.year, 'N/A') AS year,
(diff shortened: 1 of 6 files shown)


apps/web-client/src/lib/forms/schemas.ts
```diff
@@ -66,6 +66,6 @@ export type CustomerOrderSchema = Infer<typeof customerOrderSchema>;
 export const customerOrderSchema = z.object({
 	id: z.number(),
 	fullname: z.string().default(""),
-	email: z.string().email(""),
+	email: z.string().email("").optional(),
```
delasous, line 69:
```suggestion
	email: z.string().email().optional(),
```

apps/web-client/src/lib/db/cr-sqlite/customers.ts
```diff
@@ -79,16 +79,34 @@ export const getAllCustomerOrderLines = async (db: DB): Promise<DBCustomerOrderL
...
+		`SELECT
+			customer_order_lines.id,
+			customer_order_lines.isbn,
+			customer_order_lines.customer_id,
+			customer_order_lines.created,
+			customer_order_lines.placed,
+			customer_order_lines.received,
+			customer_order_lines.collected,
+			GROUP_CONCAT(customer_supplier_order.id) as supplierOrderIds,
+			COALESCE(book.title, 'N/A') AS title,
```
delasous, line 91: Nice

delasous (approved):

PR #692, opened 2025-01-24
Feature/get non reconciling supplier orders

### LINKED ISSUE #675: A supplier order should indicate whether it is part of an in progress reconciliation
When supplier orders are selected for reconciliation, their ids are included in the new `reconciliation_order.supplier_order_ids` json array. The `supplier_order` itself does not indicate if it is part of an ongoing reconciliation. This means we cannot easily exclude it from the list of "ordered" orders, and it could be re-selected for another reconciliation.

### wip
A `supplier_order` should minimally include a timestamp or boolean flag to indicate that it is already being reconciled. When this issue was brought up yesterday our discussion led to the suggestion that it may need two timestamp/flags: one for when the reconciliation started, the other for when it was finalised. 

I now believe that we don't need the latter, as the primary outcome is that they are excluded from the "ordered" list, and the finalisation is noted in the `reconciliation_order.finalized` column. If this holds, then the new supplier order column could be called `received` or `reconciled`. 

Alternatively: if it might be useful for a user to navigate from an order to a reconciliation process, this column could reference the `reconciliation_order.id`?

### implementation
This update will affect:
- The "Ordered" supplier order list should exclude orders that are being reconciled
- The createReconciliationOrder handler should update the relevant supplier orders

fadwamahmoud:
Fixes #675 

Functionality for Non-Reconciling Supplier Orders:
- getNonReconcilingSupplierOrders, was added to the order-reconciliation.ts file. It queries supplier orders that have not been reconciled 

- We use a CTE which extracts supplier_order_ids from a JSON array (supplier_order_ids) in the reconciliation_order table
```
WITH Reconciled AS (
    SELECT CAST(value AS INTEGER) AS supplier_order_id
    FROM reconciliation_order AS ro
    CROSS JOIN json_each(ro.supplier_order_ids)
)
```
1. The json_each function is used to break down the JSON array into individual rows, with value representing each order ID.
2. The CAST(value AS INTEGER) ensures that the extracted values are treated as integers.
3. a CROSS JOIN is an explicit way to produce the Cartesian product of two tables without any joining condition.

-` AND r.supplier_order_id IS NULL`: Excludes supplier orders that are referenced in a reconciliation order.


delasous:
It would be helpful to include a comment about the approach you have taken.

delasous:
This doesn't need to be a comprehensive breakdown of the changes you've made. Just some high level comments along the lines, "Issue X discussed these approaches, I chose this last one, these other things were impacted as a result"

apps/web-client/src/routes/orders/suppliers/+page.svelte
```diff
@@ -79,6 +79,7 @@
 
 			const placed = await getPlacedSupplierOrders(data?.ordersDb);
 			await createReconciliationOrder(data?.ordersDb, [placed[0].id]);
+			await createSupplierOrder(data?.ordersDb, possibleLines);
```
delasous, line 82: What does this do?

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
```diff
@@ -204,6 +204,39 @@ export async function finalizeReconciliationOrder(db: DB, id: number) {
 	});
 }
 
+export async function getNonReconcilingSupplierOrders(db: DB) {
```
delasous, line 207: - This is difficult to read - how about `getUnreconciledSupplierOrders`? 
- Add tests
- Add JSDoc comment

delasous (changes requested):

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
```diff
@@ -204,6 +204,50 @@ export async function finalizeReconciliationOrder(db: DB, id: number) {
...
+/**
+ * Retrieves all supplier orders that have not yet been included in any reconciliation process.
+ * @param db
+ *
+ * @returns {Promise<SupplierPlacedOrderLine[]>} Array of unreconciled supplier orders with:
+ *  - id: The supplier order ID
+ *  - supplier_id: The ID of the supplier
+ *  - supplier_name: The name of the supplier
+ *  - created: Creation timestamp of the order
+ *  - total_book_number: Total quantity of books in the order (0 if no books)
```
delasous, line 216:
```suggestion
```

Removes the description of the `SupplierPlacedOrderLine` properties. This was Silvio's major feedback on my jsdoc PR

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
```diff
@@ -204,6 +204,50 @@ export async function finalizeReconciliationOrder(db: DB, id: number) {
 	});
 }
 
+/**
+ * Retrieves all supplier orders that have not yet been included in any reconciliation process.
```
delasous, line 208:
```suggestion
 * Retrieves all supplier orders that have not been selected for reconciliation in any `reconciliation_order`.
```

delasous (approved):

PR #693, opened 2025-01-24
Feat/orders display id

### LINKED ISSUE #658: Customers should have an "id" field visible to the user. It should be a string
Let's discuss this a bit more before implementing.

The field's default value should be:

* if there are no customers, `1`
* If there are customers, `the current highest (but less than 10k) number` + 1

ikusteu:
Fixes #658


delasous:
I don't understand why this is necessary (#658 generally)? The record `id` was already displayed under "Customer Order" in the left column. It could maybe be indicated a little more prominently. I don't see any advantage to an editable "displayId", just adds more complexity / checks IMO

silviot:
@delasous I asked for this because in my opinion it allows users to edit customer ids with fewer risks of code defects. The use case I have in mind is: "Oh, no! User 334 should really have id 344 - and user 344 should actually be 345 - let's fix this!".

silviot (approved):

PR #695, opened 2025-01-27
Create a table for listing currently reconciling orders

### LINKED ISSUE #694: UI: View for listing currently reconciliating orders
We need a view that lists currently reconciliating orders
It needs to show whether they're finalized

fadwamahmoud:
Fixes #694 
- A new table component for displaying reconciliation orders
- Introduced a new function getReconcilingOrders which queries the database for orders that are not finalized (finalized = 0)
- Added test for retrieving reconciling orders (getReconcilingOrders).
- Added a new SupplierOrderFilterStatus value: "reconciling"
- Added getReconcilingOrders to the load function

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
@@ -47,7 +47,22 @@ ascending
  */
 export async function getAllReconciliationOrders(db: DB): Promise<ReconciliationOrder[]> {
 	const result = await db.execO<ReconciliationOrder>(
-		"SELECT id, supplier_order_ids, finalized, updatedAt, created FROM reconciliation_order ORDER BY id ASC;"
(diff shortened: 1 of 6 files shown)


apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
```diff
@@ -40,7 +40,22 @@ ascending
...
+	);
+	return result;
+}
+/**
+ * Retrieves currently reconciling orders, ordered by ID
+ascending
+ * @param db
+ * @returns ReconciliationOrder array
+ */
+export async function getReconcilingOrders(db: DB): Promise<ReconciliationOrder[]> {
```
delasous, line 54: I feel like this can be merged with the function above, by passing an extra param optional param `finalised: boolean`?  If the param is present, add the where clause in the query with 1 or 0; if its not, don't include it and return all orders...

delasous (changes requested):

PR #696, opened 2025-01-27
Create test for supplier order lines aggregation

### LINKED ISSUE #665: Manage order line quantity across customer, supplier and reconciliation orders
In the current implementation:
- Customer order lines do not have quantity. 
- Supplier order lines do
- Reconciliation order lines (the scanned delivered books) do
- The reconciliation compare functionality that Fadwa is working on in #662 will compare quantites between supplier and reconciliation order lines

Needs to be implemented:
- [ ] the `createSupplierOrder` logic does not aggregagte customer order lines by isbn
- [x] delivered books that are added to the reconciliation order also need to be aggregated. This could either be something that the scan input handler does or if we add `reconcilation_order_lines` to the model (#652), it could be something that a new query-handler `addReconcilationOrderLines` does.

fadwamahmoud:
Fixes #665
I had a misunderstanding that supplierOrderLines were not being aggregated when creating a supplierOrder so I wrote a test for checking that supplierOrderLines are getting aggregated on the same isbn

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/__tests__/supplier-order.test.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/__tests__/supplier-order.test.ts
@@ -12,6 +12,7 @@ import {
 	createSupplierOrder,
 	getPlacedSupplierOrderLines
 } from "../suppliers";
+import { addBooksToCustomer } from "../customers";
 
 describe("Supplier order handlers should", () => {
 	let db: DB;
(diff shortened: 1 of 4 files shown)


apps/web-client/src/lib/db/cr-sqlite/__tests__/supplier-order.test.ts
```diff
@@ -108,8 +109,73 @@ describe("Supplier order handlers should", () => {
 		const newPossibleOrderLines = await getPossibleSupplierOrderLines(db, 1);
 		expect(newPossibleOrderLines.length).toBe(0);
 	});
+	it("Aggregates supplier orders lines quantity when placing an order", async () => {
```
delasous, line 112: This description should align with the behaviour: "... when getting possible order lines"

delasous (changes requested):
Ok, I see. Both of the supplier order page tables need to a quantity column then

delasous (approved):

PR #704, opened 2025-01-30
Create a test for customer order

fadwamahmoud:
-  Create dbHandlers for customer order flow
- Write tests for customer and supplier order flows
- Use fixtures for setting up customer orders                   
- Use dbNamePersisted in orders layout instead of hardcoded DB name which was causing test data to be created in a different context


apps/e2e/integration/customer_order.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+	await page.goto(`${baseURL}orders/customers/`);
+
+	await page.getByRole("link", { name: "Update" }).first().click();
+
+	await page.waitForURL(`${baseURL}orders/customers/1/`);
+	await expect(page.getByText(customer.fullname)).toBeVisible();
+	await expect(page.getByText(customer.email)).toBeVisible();
+
+	const isbnField = page.getByRole("textbox");
+	isbnField.fill("5678");
```
delasous, line 98: Same comments as above when checking existing customer order lines

apps/e2e/integration/customer_order.spec.ts
delasous, line 1: I don't think we should be using `page.getByText` to make assertions against table data. It gives us very little garauntee on the structure of the data. It should be possible to `const table = getByRole("table")` and `table.getByRole("row")` and then make assertions about that rows content...

delasous (changes requested):

delasous:
Add assertion for customer order line deletion when #714 is merged

apps/e2e/integration/customer_order.spec.ts
fadwamahmoud: sorry didn't see this one before I merged. Will resolve it in 709

Issue #691, opened 2025-01-24
Misc order mgmt todo dump


delasous:
This is a list of todos that I've noted quickly over the last couple weeks. Dumping here so they are public. I haven't had time to think these things through too thoroughly - they may or may not need to be done - the last three are more for discussion...

- [x] Getting errors when running Populate DB on /debug: 
    - I noticed then when updating data (customer_order_line -> book.isbn relations), but didn't have time to fix it
    - Needs to look something like this... but this is still showing `null` for all columns of data for "delivered" lines on the compare stage
	```
		INSERT INTO reconciliation_order_lines (reconciliation_order_id, isbn) VALUES
		(1, '9781234567897'),
		(2, '9788804797142');
	```
   - The reocnciliation e2e is still passing, but since this is broken... the assertions need  to be tightened up?
- [x] Any types that reference book data should probably reference the book data type. Also `BookEntry` from the old db package is referenced in this `cr-sqlite/types`. With `getCustomerOrderLines` I have tried to only return the book data that is used in that table (I've been more selective in this pass at design, I don't think every table needs a publisher list, edited by... )
    - [x] I have done this focussing on Supplier Order types. See [`cr-sqlite/types` in #705](https://github.com/librocco/librocco/pull/705/files#diff-4a5815c24f6c45d2bd4d1653a08c6fcf23f66e59dfe4aaae6f4c4ed4f2236b62)
    - [x] Reconciliation order types
    - [x] I tidied up Customer types a little. Could probably be improved a little more in line with above
- [x] `coalesce` book data to default values (like in `getCustomerOrderLines`) in other methods that retrieve/join on book data?
    - [x] I have done this focussing on Supplier Order handlers. See [`cr-sqlite/suppliers.ts` in #705](https://github.com/librocco/librocco/pull/705/files#diff-37cd4bf815e44aad8e2c9e2a27a3613f5f7be16acb13b06f8aee184fa1c53a7e)
    - [x] Reconciliation order types
    - [x] Customer order types?
- [ ] The meta data (left column in the desktop UI) for customer, supplier and reconciliation order views passes db date timestamps to `Date`. In `getCustomerOrderLines` we have a small helper "marshCustomerOrderLineDates" to return Date objects in the handlers. I think this should be standardises across the board
    - [x] In #705: I have continued to return number timestamps, but minimised duplicate `new Date()` constructions

- [x] Extend unit testing:
	- [x] Assess thoroughness of coverage in customer and reconciliation orders
[...]

PR #706, opened 2025-02-03
Fix/order touchups

ikusteu:
A collection of fixups with respect to #691 + some of my finds of bugs and wrinkles, namely:
- attempt sync with the server only if instructed to do so (via env variable - for dev purposes, should be rethought when preparing for production)
- remove '/s/*' route (not used)
- use root layout 'dbCtx' in 'orders' section (replace 'ordersDb') - we have a single source of truth for the DB (well...we had that already, but this cleans up the accessing of said single source of truth)
- make sure all load functions check for 'dbCtx' being defined (browser env) before interacting (with respect to #691)
- fix reconciliation order types (in a sense...should be revisited)
- update customer order creation in reconciliation order view


delasous (approved):

delasous:
I resolved a few conflicts that were introduced when I merged #695 (I resolved similar conflicts there too before merging). I picked up on a few other small fixes as I did so:
- The debug page was using the wrong dbName
- A `customerOrderLIne` no longer has a `supplierOrderIds` property
- There was old mock data sitting in one of the routes

PR #710, opened 2025-02-05
Fix/supplier routes

### LINKED ISSUE #697: Feat: supplier route updates
The current orders routing, particularly the `/suppliers` paths, needs an update, especially to account for the supplier mgmt being introduced in #529/#528. I think the following should be done separately, before 529/528, as it lays the groundwork for some other updates I'd like to make to the supplier order mgmt views (lists/steps):

1. `/suppliers/[id]` currently shows a new supplier order - that is, order lines for that supplier that have been aggregated from customer orders. I think we need `/suppliers/[id]` to be a dashboard for that supplier, it can show the supplier details + publisher mapping, order history, and link to "place a new order". The new "supplier-publisher" config introduced in 529/Figma mocks, should go here.
2. The "create new supplier order" functionality at `/suppliers/[id]` should be moved to `/suppliers/[id]/create-order`, or something along these lines
3. `/suppliers/+page.svelte` currently acts as the central point for managing entities in the order->reconciliation flow. I think this needs be moved to `/suppliers/orders/+page.svelte`, so that the existing route can act as a list of suppliers ([this](https://www.figma.com/design/KkaMHw74ZDvGOv5heaVRiP/Librocco-v2?node-id=2200-11684&t=B0u1D3WR1eCQC4EJ-0)). Additionally, there is an existing route `/suppliers/order/[id]` which should be moved under this pluralized order list view.

delasous:
Fixes #697. 

I have done enough to make space for the supplier list and supplier mgmt (configure publisher mapping) views that need to be added. I have added placeholder pages for these views

There's more to do here to improve the order management flows - but I will write that up separately


silviot (approved):
LGTM. Interesting how no update was needed for tests.

delasous:
Fadwa is writing the e2e that would cover this now

PR #741, opened 2025-02-10
supplier order creation touchups:

### LINKED ISSUE #715: Refactor: review `createSupplierOrder` sql/implementation
WIP: needs detail

Evaluate weaknesses and try to break it. #713 attempted to improve test coverage for the `supplier` handlers, but it focussed mainly on the `get...` querys => test's should also be reviewed.

I don't think it should be possible to create a new `supplier_order` if no `possibleOrderLines` are passed. There are test cases (starting lines 392 and 524 in `supplier-order.test`) that should be updated.

ikusteu:
* throw an error if trying to create an empty supplier order
* restrict supplier order creation to single supplier order at a time
* edge case: throw an error if trying to create a supplier order with lines belonging to a different supplier
* edge case: check for requested quantity (as specified by params) and required quantity (found in customer order lines) - always take a min of the two
* extend tests to test for aforementioned updates

Fixes #715


delasous (approved):
The defensive check seems a little excessive but I guess it can't hurt now that we have it.

Issue #722, opened 2025-02-07
Fix: debug page


delasous:
- [ ] The reset button doesn't fully reset the db (populate -> reset -> populate will throw unqiue constraint errors)
- [ ] Add `customer_order_lines` that have not been placed (to give us mock "possible" supplier orders upfront - currently have to edit manually edit a customer order after population to make this happen)

delasous:
Additionally: I can't view an existing customer order, the app crashes. Is this due to debug data that is out of sync with our models?

PR #746, opened 2025-02-11
Fixes and improvements to the debug page

chiruzzimarco:
Related to issue #722 .

This PR fixes a bug in the debug page which caused a console error when resetting the database. The issue was caused by the `reconciliation_order_lines` table not being properly resetted.

Additionally this implements a database query interface useful for debugging (thank you @silviot for suggesting this).

Example SQL data was moved out of the routes directory into the `lib/__testData__` directory which is probably the place where it belongs.

### DIFF
--- a/apps/web-client/src/routes/debug/+page.svelte
+++ b/apps/web-client/src/routes/debug/+page.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { BookCopy, Library, PackageMinus, Search, Settings, PersonStanding, Book, Truck } from "lucide-svelte";
-	import { Plus, RotateCcw } from "lucide-svelte";
+	import { Plus, RotateCcw, Play } from "lucide-svelte";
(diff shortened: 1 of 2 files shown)


apps/web-client/src/routes/debug/+page.svelte
```diff
@@ -199,29 +228,73 @@
...
+									<div class="mt-4 rounded-lg bg-red-500 p-3 text-white shadow">
+										{errorMessage}
+									</div>
+								{:else if queryResult.length === 0}
+									<p>No results found.</p>
+								{:else}
+									<table class="table">
+										<thead>
+											<tr>
+												{#each Object.keys(queryResult[0]) as column}
```
delasous, line 255: Column names are numbers. I think we need `execO` to get back an object?

delasous (changes requested):
Looks good. Love the little query runner.

silviot (approved):

PR #747, opened 2025-02-11
Collect customer order

### LINKED ISSUE #730: feat: marking orders as collected
Correct me if I'm wrong, but I didn't find any place in the app where a customer order line can be marked as collected. This should probably be a functionality of the `/customers/:id` view

fadwamahmoud:
Fixes #730 
- Add Collect column in customer order table that displays collect button which is only enabled if received timestamp is set
- Collected books display the date of collection in the collect column
- New funciton markOrderAsCollected which sets the collected timestamp
- Create fixture for a collected customer order 
- E2e test

### DIFF
--- a/apps/e2e/helpers/cr-sqlite.ts
+++ b/apps/e2e/helpers/cr-sqlite.ts
@@ -1,5 +1,13 @@
 import type { DB } from "@vlcn.io/crsqlite-wasm";
-import { Customer, Supplier, SupplierOrderLine } from "./types";
+import {
+	Customer,
+	DBCustomerOrderLine,
+	PlacedSupplierOrder,
+	ReconciliationOrder,
+	ReconciliationOrderLine,
+	Supplier,
+	SupplierOrderLine
+} from "./types";
 
 // #region books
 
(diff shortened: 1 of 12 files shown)


apps/web-client/src/lib/db/cr-sqlite/customers.ts
```diff
@@ -291,3 +291,35 @@ export const markCustomerOrderAsReceived = async (db: DB, isbns: string[]): Prom
...
+ * Marks customer order lines as collected when customer order lines are picked up.
+ * For each supplied ISBN, it updates the earliest unfulfilled customer order line
+ * (that has been received but not collected) with the current timestamp as collected date.
+ *
+ * @param {DB} db - The database connection instance
+ * @param {isbns[]} isbns - Array of supplier order line isbns that have been collected
+ * @returns {Promise<void>} A promise that resolves when all relevant customer orders are marked as collected
+ */
+
+export const markCustomerOrderAsCollected = async (db: DB, isbns: string[]): Promise<void> => {
```
silviot, line 305: The `markCustomerOrderAsCollected` curently only takes ISBNs as parameter.

It has no way to know exactly what customer (and thus order line) it should mark as collected.

Either it takes customer id and isbn (but then if a single customer ordered the same book twice we might change "the other row" when users click on the button in the UI, since we have no way to differentiate the two rows) or a `rowid` for a specific `customer_order_line`.

apps/web-client/src/lib/db/cr-sqlite/customers.ts
```diff
@@ -291,3 +291,35 @@ export const markCustomerOrderAsReceived = async (db: DB, isbns: string[]): Prom
...
+		 UPDATE customer_order_lines
+            SET collected = (strftime('%s', 'now') * 1000)
+            WHERE rowid IN (
+                SELECT MIN(rowid)
+                FROM customer_order_lines
+                WHERE isbn IN (${placeholders})
+                    AND placed IS NOT NULL
+                    AND received IS NOT NULL
+                    AND collected IS NULL
+                GROUP BY isbn
```
silviot, line 320: This query is extremely problematic. But there is no query that could do the right thing with just the ISBNs.

When writing the new query please remember to use javascript to get the timestamp and put that one into the db. It will already have millisecond precision, so no need to multiply it by 1000. [See #676](https://github.com/librocco/librocco/pull/676/files).

apps/e2e/helpers/cr-sqlite.ts
```diff
@@ -338,3 +346,107 @@ export async function createReconciliationOrder(db: DB, supplierOrderIds: number
 	);
 	return recondOrder[0].id;
 }
+
+export async function finalizeReconciliationOrder(db: DB, id: number) {
```
silviot, line 350: This is duplicated from [apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts](https://github.com/librocco/librocco/blob/45aed8e601eccd3dc0cabf38225c1d29889e6755/apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts#L176) right? I think we should find a way to avoid the duplication (like a symlink) or at least we should leave a comment here pointing to the original function.

My concern is that in the future we will update one but not the other and will have problems because of that.

apps/e2e/helpers/cr-sqlite.ts
```diff
@@ -338,3 +346,107 @@ export async function createReconciliationOrder(db: DB, supplierOrderIds: number
...
+            	        AND received IS NULL
+            	    GROUP BY isbn
+				);`,
+				customerOrderLines
+			);
+		}
+	});
+}
+
+export async function getPlacedSupplierOrders(db: DB): Promise<PlacedSupplierOrder[]> {
```
silviot, line 400: Same as above, I think we should mention that this is copied over from https://github.com/librocco/librocco/blob/45aed8e601eccd3dc0cabf38225c1d29889e6755/apps/web-client/src/lib/db/cr-sqlite/suppliers.ts#L263

apps/e2e/helpers/cr-sqlite.ts
```diff
@@ -338,3 +346,107 @@ export async function createReconciliationOrder(db: DB, supplierOrderIds: number
...
+		LEFT JOIN supplier_order_line sol ON sol.supplier_order_id = so.id
+		LEFT JOIN book ON sol.isbn = book.isbn
+        WHERE so.created IS NOT NULL
+        GROUP BY so.id, so.supplier_id, s.name, so.created
+        ORDER BY so.created DESC;`
+	);
+	return result;
+}
+
+export async function addOrderLinesToReconciliationOrder(db: DB, params: { id: number; newLines: { isbn: string; quantity: number }[] }) {
```
silviot, line 420: Let's mention this is copied over from https://github.com/librocco/librocco/blob/45aed8e601eccd3dc0cabf38225c1d29889e6755/apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts#L147

silviot (changes requested):
The `markCustomerOrderAsCollected` needs substantial changes.

The functions copied over from `web-client` should be marked as such. Maybe placing them in a separate file, and naming the file `duplicates.ts` would make things easier.

We should never use the `now` from sqlite-land. We should instead use Javascript's `new Date()` to get millisecond precision.

apps/e2e/helpers/cr-sqlite.ts
ikusteu: We already had this discussion: the helpers are quite similar (but not exactly the same) as the ones used in the app. If a discrepancy happens, it will be evident as the person introducing it will:
1. Immediately be notified (by failing e2e)
2. Know where they might have introduced it (which function was edited)

☝️ my points for why the symlink wouldn't work

However, definitely couldn't hurt to add a comment at the top of the file, explaining this.

apps/web-client/src/lib/db/cr-sqlite/customers.ts
ikusteu: Customer order line actually has an `id` property so that seems the most straightforward way to handle this.

apps/e2e/helpers/cr-sqlite.ts
ikusteu: Tbh, I can't see a scenario where we'd want to use this function in e2e test code (the UI should reflect the necessary state -- and used to make assertions)

apps/web-client/src/lib/db/cr-sqlite/customers.ts
```diff
@@ -251,43 +253,73 @@ export const removeBooksFromCustomer = async (db: DB, customerId: number, bookId
...
+    WHERE id IN (${placeholders})
+        AND placed IS NOT NULL
+        AND received IS NULL
+    ORDER BY placed ASC
+    LIMIT 1
+);`,
+			[timestamp, ...rowIds]
+		);
+	});
+};
```
ikusteu, line 291: 2 points here: 
- can we use customer order line `id` here (name the param `ids` as well...`rowIds` is a bit opaque)
- when using `customer_order_lines.id` - it is unique, no need for the subquery, the sql could simply set received where `id` (in table) in `ids` (param):
```ts
const markAsReceived(db: DB, ids: string[]) => {
  // ...
  
  db.exec(`
      UPDATE ... SET ... WHERE id IN (${multiplyString("?", ids.length)})
    `,
    [timestamp, ...ids]
  )
}
```

apps/e2e/helpers/cr-sqlite.ts
```diff
@@ -338,3 +405,119 @@ export async function createReconciliationOrder(db: DB, supplierOrderIds: number
...
+		`SELECT
+			id, customer_id, created, placed, received, collected,
+			col.isbn
+			FROM customer_order_lines col
+		WHERE customer_id = $customerId
+		ORDER BY col.isbn ASC;`,
+		[customerId]
+	);
+	return result;
+};
```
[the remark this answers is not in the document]
fadwamahmoud: This one only gets the statuses, it's not concerned with the book data like the getCustomerOrderLines function in cr-sqlite/customers.ts

apps/e2e/helpers/fixtures.ts
```diff
@@ -38,6 +53,31 @@ export const testOrders = test.extend<OrderTestFixture>({
...
+
+		const placedOrderIds = placedOrders.map((placedOrder) => placedOrder.id);
+		const reconciliationOrderId = await dbHandle.evaluate(createReconciliationOrder, placedOrderIds);
+		await dbHandle.evaluate(addOrderLinesToReconciliationOrder, { id: reconciliationOrderId, newLines: [supplierOrderLine] });
+		await dbHandle.evaluate(finalizeReconciliationOrder, reconciliationOrderId);
+
+		const customerOrderLines = await dbHandle.evaluate(getCustomerOrderLineStatus, customer.id);
+
+		await use(customerOrderLines);
+	},
```
ikusteu, line 80: I would much rather prefer to:
- define static data in the fixture (with ids, states, orders, etc...)
- set up the state with such static data (in DB)
- return the static data

This returns dynamic data - stores something in the DB and returns a query result. This complicates and obfuscates things (in case of debugging a failing test).

I'll merge this as I think we'll need to do a full sweep of fixtures, e2e in general and dashboard helpers, anyway.
Just leaving this here as some feedback.

ikusteu (approved):

PR #752, opened 2025-02-17
Review/customer order tests

### LINKED ISSUE #718: Refactor: review `customer-order` tests
WIP needs detail. 

Do what #713 did for supplier orders to customer orders. I don't think there is as much to do here, but its worth a review;
- Evaluate and improve test coverage where necessary
- Use refs to test data in assertions

There is also a test commented out at the end of the file which needs to be re-introduced?

ikusteu:
Fixes #718 

As the ticket requested, I've reviewed (and greatly extended/reorganised customer order tests).

Furthermore, while testing I've applied some updates to general customer order functionality:

**Replace 'getAllCustomerOrders' + 'getAllCustomerOrderLines' -> 'getCustomerOrderList'**:
- previously, the customer orders list view used the combination of `getAllCustomerOrders` and `getAllCustomerOrderLines`, combining the two in JS code to attach `status` to every order
- I've replaced this so that the `getCustomerOrderList` returns just that: a list of all customer orders, each containing an already calculated `completed` field (if `true` -> `completed`, if `false` -> `in_progress`)
- I've thoroughly tested this, replaced the usage in the UI and removed the old functions

**Add customer order line status field:**
- this is currently done at the query level, it extends the returned structure for a customer order line (doesn't remove any of the old functionality)
- the UI still uses the old functionality and this should be replaced
- the logic for status is an enum:
  - 0 - pending line
  - 1 - placed line
  - 2 - received line
  - 3 - collected line
- I've used an enum for this (a proper `enum`, not the string one) as it's convenient when calculating order's `completed` status as the enum is comparable - when calculating order status, we take the minimum of all associated order lines (the order status will be `3` only if all are collected) and then we check if the value is `3` - completed (anything else and the order is not `completed`)


silviot (approved):
LGTM

PR #758, opened 2025-02-20
fix: remove the annoying check in 'createSupplierOrder'

ikusteu:
* remove the check checking if there are enought customer order lines for a particular order
* remove the test for the check, TODO: simplify all other tests being unnecessarily complex due to this functionality


ikusteu:
Merging this as I have other code depending on it and it's a small one. Whoever is interested, do check the closed PR ex post

PR #761, opened 2025-02-21
Refactor/e2e db helpers sharing

### LINKED ISSUE #749: Devise a strategy to share code between `web-client` and `e2e`
Currently there are duplicate functions (some are not exact duplicates, like `upsertBook`, I believe) defined both in `e2e/helpers/cr-sqlite.ts` and in the `web-client` package:

| Function Name | web-client Location |
| :--- | :--- |
| `upsertBook` | lib/db/cr-sqlite/books.ts |
| `upsertWarehouse` | lib/db/cr-sqlite/warehouse.ts |
| `createInboundNote` | lib/db/cr-sqlite/note.ts |
| `createOutboundNote` | lib/db/cr-sqlite/note.ts |
| `updateNote` | lib/db/cr-sqlite/note.ts |
| `addVolumesToNote` | lib/db/cr-sqlite/note.ts |
| `commitNote` | lib/db/cr-sqlite/note.ts |
| `upsertNoteCustomItem` | lib/db/cr-sqlite/note.ts |
| `multiplyString` | lib/db/cr-sqlite/suppliers.ts<br>lib/db/cr-sqlite/customers.ts |

We should have a single definition of each function. If two functions have the same name, they should also behave the same. So we need to either make them the same or change one name.

ikusteu:
Fixes #749


apps/e2e/helpers/cr-sqlite.ts
```diff
@@ -1,55 +1,69 @@
...
+
+// #region books
+
+// Extend the window object with the db
+declare global {
+	interface Window {
+		_db: DB;
+		db_ready: boolean;
+
+		books: Record<string, any>;
```
silviot, line 19: I think it's good to resist the temptation to better specify this `any` here. I second not having types for the return objects of these functions.

apps/e2e/helpers/cr-sqlite.ts
```diff
@@ -1,55 +1,69 @@
...
+		const w = window as { [key: string]: any };
+
+		// Wait for the db to become initialised
+		await new Promise<void>((res) => {
+			if (w["db_ready"]) {
+				return res();
+			} else {
+				// Creating a separate function, as we want to run the listener only once and then remove it
+				const finalise = () => {
+					window.removeEventListener("db_ready", finalise), res();
```
silviot, line 52: I was about to comment that this `finalize` function does nothing, since it only removes the event listener that just invoked it. But then I noticed the `,res()` promise resolution.

I think it would be more readable in its own line, with a semicolon.

```suggestion
					window.removeEventListener("db_ready", finalise);
					res();
```

...but I fear I might be missing something.

ikusteu: No, you're right - this was a one liner turned into a block (resulting in this retarded hybrid)

silviot (approved):

delasous (approved):
It looks good and it works, so I think it should go in. I've added a comment to the discussion in #749 with a suggestion for improving types

Issue #724, opened 2025-02-07
Feat: edit book details of order table lines


delasous:
WIP needs detail

Book data can be updated from any table on the inventory side. It is useful as a user starting from a fresh db, without the book data plugin. Should this be possible on all order tables, or say, just customer orders? It seems particularly useful to be able to add/edit price and publisher here. And it seems useful to be able to edit it for "possible" supplier orders - but in this case, an edit to publisher could move the line to a different order.

Do we agree with:
1. Customer order lines -> can edit all book data 
2. Possible supplier order lines -> can edit any book data, but not publisher?
3. Placed supplier order lines -> can't edit book data
4. Reconciliation orders -> can't edit book data

The BookForm will need to be updated to daisy ui. 

Not sure this is a priority for MVP, so have added nice-to-have label

PR #773, opened 2025-02-25
Allow editing of books from the customer order detail table

chiruzzimarco:
Related to issue #724 , this PR will allow books in a customer order to be edited directly from the customer order interface.

## Summary by CodeRabbit

- **New Features**
  - Introduced a dynamic book editing form that enables users to update details including ISBN, title, price, year, authors, publisher, edited by, category, and out-of-print status.
  - Enhanced the customer order page with dynamic dialogues for editing entries and a preloaded publisher list to streamline data entry.

- **Improvements**
  - Refined order management interactions with clearer status indicators and improved UI flow to ensure smooth confirmation of actions.


apps/web-client/src/lib/forms/DaisyUIBookForm.svelte
```diff
@@ -0,0 +1,90 @@
...
+	<label class="input-bordered input flex items-center gap-2">
+		Year
+		<input bind:value={$formStore.year} name="year" type="number" {...$constraints.year} />
+	</label>
+	<label class="input-bordered input flex items-center gap-2">
+		Authors
+		<input bind:value={$formStore.authors} name="authors" type="text" {...$constraints.authors} />
+	</label>
+	<label class="form-control w-full">
+		<div class="label">Publisher</div>
```
delasous, line 65: @silviot @ikusteu on the inventory side, an employee should be able to type in a new publisher and save the form with this new option right? Should this be possibe here?

apps/web-client/src/lib/forms/DaisyUIBookForm.svelte
```diff
@@ -0,0 +1,90 @@
...
+	use:enhance
+	method="POST"
+	id={testId("book-form")}
+>
+	<div class="flex justify-between">
+		<label class="input-bordered input flex items-center gap-2">
+			ISBN
+			<input bind:value={$formStore.isbn} name="isbn" type="text" class="grow" placeholder="0000000000" {...$constraints.isbn} disabled />
+		</label>
+		<button disabled={!isExtensionAvailable} type="button" class="btn-secondary btn" on:click={() => onFetch($formStore.isbn, formStore)}>
```
delasous, line 42: This button really needs to provide more feedback about the state of the request. The original form didn't have it either, so I don't think it needs to be tackled here. Just a note that I've opened #778

delasous:
New form looks nice though!

apps/web-client/src/lib/forms/DaisyUIBookForm.svelte
```diff
@@ -0,0 +1,90 @@
...
+		Year
+		<input bind:value={$formStore.year} name="year" type="number" {...$constraints.year} />
+	</label>
+	<label class="input-bordered input flex items-center gap-2">
+		Authors
+		<input bind:value={$formStore.authors} name="authors" type="text" {...$constraints.authors} />
+	</label>
+	<label class="form-control w-full">
+		<div class="label">Publisher</div>
+		<select class="select-bordered select w-full" bind:value={$publisherProxy}>
```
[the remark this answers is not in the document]
chiruzzimarco: @delasous I can't reproduce this, what browser are you using? Can you provide a screenshot?

apps/web-client/src/lib/forms/DaisyUIBookForm.svelte
ikusteu: I think so, yeah... (if I remember correctly, the publisher list is made by a query retrieving all distinct `publisher` values -- no able exists)

chiruzzimarco: @delasous Yes the inventory book form features a custom input component and some meltUI magic to provide a "select2" like functionality, thus we have a regression here.
I spent some time trying to re-implement this with DaisyUI in this contenxt but ultimately opted for a clean/easy select field since DaisyUI doesn't provide any kind of searchable select layout and it was taking me too much time to figure this out.
@ikusteu @delasous I suggest addressing this in a dedicated issue, what do you think?

delasous: Yeah I agree. Just clicking through the old form and it seems to be a little buggy, so we could fix that in the new unit of work too

apps/web-client/src/lib/forms/DaisyUIBookForm.svelte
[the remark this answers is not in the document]
delasous: Screenshot from 2025-03-03 13-15-54

I'm using Firefox. But on closer inspection, it looks like the options are being presented as a single option.

chiruzzimarco:
@CodeRabbit review full

apps/web-client/src/routes/orders/customers/[id]/+page.svelte
```diff
@@ -113,6 +145,31 @@
...
+		 *
+		 * It is still safe to assume that the required properties of BookEntry are there, as the relative form controls are required
+		 */
+		const data = form?.data as BookEntry;
+
+		try {
+			await upsertBook(db, data);
+			bookFormData = null;
+			customerOrderLines = await getCustomerOrderLines(db, Number(customerId));
+			open.set(false);
```
Comment (ikusteu) [inline apps/web-client/src/routes/orders/customers/[id]/+page.svelte:167]: This shouldn't be set explicitly: instead, it should:
- `upsertBook`
- close the form

What should happen after that (as a reactive effect):
- the update to `books` table should trigger the `invalidate` function - invalidating the load function
- the load function runs again (having been invalidated) and retrieves the up-to-date book data

I just checked the reactivity subscriptions and they lack `book` table subscription (see: lines 64-79, `onMount` block)
More on the load function invalidation: https://svelte.dev/tutorial/kit/invalidation

ikusteu (changes requested):
All in all it looks good to me. I left one comment regarding reactive subscriptions, which I think is an important concept in the app's context.

Issue #756, opened 2025-02-18
Bug/Enhancement: Total delivered should display (Number of delivered book instances that were ordered / Ordered)


fadwamahmoud:
In the compare view the Total Delivered currently displays:
  - (Number of total books delivered / Ordered)
  - so if there were 2 unmatched books delivered and 2 originally ordered books would display 2/2 which makes it seem like ordered books were delivered

ikusteu:
I think we should display total quantity of books ordered / total quantity of ordered books reconciled. This is the only logical solution as other reasonings don't hold up in some edge cases:

Say we wanted to mark the filled (supplier order) lines: you'd have total number of lines / total number of lines that were fully delivered:
- supplier order: 
  - isbn: "1234", quantity: 2
  - isbn: "4321", quantity: 2
- reconciliation order (scanned books):
  - isbn: "1234", quantity: 1 🔴 
  - isbn "4321", quantity: 2  ✅ 
  - Total delivered: 1 / 2

In other case: we want to mark all the lines that were, at least, partially filled:
- supplier order: 
  - isbn: "1234", quantity: 2
  - isbn: "4321", quantity: 2
- reconciliation order (scanned books):
  - isbn: "1234", quantity: 1 ✅  
  - isbn "4321", quantity: 2  ✅ 
  - Total delivered: 2 / 2

None of the two seem intuitive as half filled orders shouldn't be considered filled and partially filled orders aren't the same as if the quantity delivered is 0.

To reiterate my suggestion:
- supplier order: 
  - isbn: "1234", quantity: 2
  - isbn: "4321", quantity: 2
- reconciliation order (scanned books):
  - isbn: "1234", quantity: 1 ✅ 🔴  
  - isbn "4321", quantity: 2  ✅ ✅ 
  - Total delivered: 3 / 4

This does make for intuitive sense: of 4 ordered books, 3 were delivered. Furthermore, this goes inline with our reconciliation logic: Outside of the supplier order, it really doesn't matter which line contained which quantity...lines either get filled (if enough books were delivered) or they don't

PR #775, opened 2025-02-26
Feature/total delivered regression test

ikusteu:
This adds a regression test for the case described in the issue the PR is referencing (#756), so the PR doesn't seem to be doing what it's supposed to. 

Furthermore, I have a different angle on the way I think Total delivered should be handler (see the issue comments)


ikusteu:
S**t...I made a mess here, didn't I...I meant to open this against your branch (not main)...but ok, it seems you've picked up off of it.

apps/e2e/integration/reconciliation_order.spec.ts
```diff
@@ -241,7 +232,43 @@ testOrders("should show unmatched deliveries when ordered books do not match sca
...
 
 	await expect(page.getByText("Total delivered:")).toBeVisible();
-	await expect(page.getByText("2 / 2")).toBeVisible();
+	await expect(page.getByText("1 / 3")).toBeVisible();
+});
+
+testOrders("regression: unmatched books shouldn't affect the Total delivered count", async ({ page, placedOrders, books }) => {
+	// NOTE: this is a very dirty (but sintactically legal) way of referencing the fixtures (having them run)
+	// and silencing the unused variable warning
+	placedOrders;
```
delasous, line 241: Use the `depends()` helper

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -68,9 +68,11 @@
 		}
 	});
 
+	$: processedOrderDelivery = processOrderDelivery(data?.reconciliationOrderLines, data?.placedOrderLines);
```
Comment (silviot) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:71]: Would it make sense to optimize it when the user is on the first screen, and only calculate this if we're past the first step?

```suggestion
	$: processedOrderDelivery = currentStep > 1 ? processOrderDelivery(data?.reconciliationOrderLines, data?.placedOrderLines) : [];
```

silviot (approved):
LGTM I suggested a small optimization.

apps/e2e/integration/reconciliation_order.spec.ts
silviot: ChatGPT suggested to add a convention to our linting: variables starting with an underscore are ok if unused. It's a convention that is also present in python.

---

By configuring ESLint to ignore variables prefixed with an underscore, you can indicate intentional non-use:

```typescript
const _placedOrders = placedOrders;
```

To enable this behavior, adjust your `.eslintrc.json` configuration:

```json
{
  "rules": {
    "@typescript-eslint/no-unused-vars": [
      "error",
      {
        "varsIgnorePattern": "^_",
        "argsIgnorePattern": "^_"
      }
    ]
  }
}
```

This setup instructs ESLint to ignore variables and function parameters that start with an underscore.

ikusteu:
I would merge this as I'd done some conflict resolution and might have messed something up in the reconciliation e2e, but I'm working on that anyway, so even if it's wrong, it will be fixed soon...

apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
Comment (ikusteu) [reply apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte]: Premature optimisation is the root of all evil...
I've accepted it and it introduced a rabbit hole-inducing regression...I back pedalled for the purpose of this PR...if this really starts causing problems, I think it should be tackled as separate unit of work

PR #776, opened 2025-02-27
Enhancement/fix reconciliation mistakes

### LINKED ISSUE #760: During reconciliation users should have a way to fix scanning mistakes
Currently users can scan books during the reconciliation process.

But in case they make a mistake they have no way to correct it.

We should let them do either of these two things, in addition to scanning books:
- [ ] Abort the reconciliation altogether (the supplier order(s) should go back to the state before reconciliation started)
- [ ] Reduce quantity of a scanned book (in case they made a mistake and scanned one copy too many, for instance, or the wrong book)

The former is a must-have. The latter is a should-have.

fadwamahmoud:
Fixes #760


apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
```diff
@@ -1321,3 +1323,204 @@ describe("Misc helpers", () => {
...
+				{ isbn: "2", quantity: 1 }
+			]);
+
+			// Delete the reconciliation order
+			await deleteReconciliationOrder(db, reconciliationOrderId);
+
+			// Verify order is deleted
+			expect(await getReconciliationOrder(db, reconciliationOrderId)).toBeUndefined();
+			// Verify lines are deleted
+			await expect(getReconciliationOrderLines(db, reconciliationOrderId)).rejects.toThrow(
```
delasous, line 1350: This just tells us the `reconciliation_order` has been deleted. We should also check the `reconciliation_order_line` table isn't returning anything

apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
```diff
@@ -1321,3 +1323,204 @@ describe("Misc helpers", () => {
...
+
+			// Create first reconciliation order
+			await createReconciliationOrder(db, reconciliationOrderId, [supplierOrderId]);
+			await addOrderLinesToReconciliationOrder(db, reconciliationOrderId, [{ isbn: "1", quantity: 1 }]);
+
+			// Delete the reconciliation order
+			await deleteReconciliationOrder(db, reconciliationOrderId);
+
+			const reconciliationOrder2Id = 1234;
+
```
delasous, line 1381: Check that the supplier order is returned with getPlacedSupplierOrders?

apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
```diff
@@ -1321,3 +1323,204 @@ describe("Misc helpers", () => {
 		expect(result["Supplier 2"]).toHaveLength(1);
 	});
 });
+describe("Reconciliation order deletion", () => {
+	describe("deleteReconciliationOrder", () => {
+		it("should delete a reconciliation order and its associated lines", async () => {
```
delasous, line 1328: Add another reconciliation order and check its not affected?

apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
```diff
@@ -1321,3 +1323,204 @@ describe("Misc helpers", () => {
...
+		it("should throw error when reconciliation order doesn't exist", async () => {
+			const db = await getRandomDb();
+
+			const nonExistentId = 999;
+			await expect(deleteOrderLineFromReconciliationOrder(db, nonExistentId, "1")).rejects.toThrow(
+				`Reconciliation order ${nonExistentId} not found or already finalized`
+			);
+		});
+
+		it("should not throw when deleting non-existent ISBN from valid order", async () => {
```
delasous, line 1482: It shouldn't throw?

delasous (changes requested):
Its good work.

There are two bigger things that I think need to be addressed:
1. It took me a while to figure out how `reconciliation_order_line` quantity was being reduced when `deleteLinesFrom` was not decrementing. Using a negative quantity with `addOrderLines` works nicely because of `on conflict`, but it betrays the function name and highlights something we haven't tested for. These two things need to be updated.
2. There are no controls to delete/reset the whole reconciliation. This could be added as a button in the left column (where customer orders have the pencil to edit), and it should have a confirmation modal step.

apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
fadwamahmoud: should it throw?

apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
```diff
@@ -210,6 +211,38 @@ export class ErrReconciliationOrderFinalized extends Error {
...
+  * - The reconciliation order with the given ID is not found
+  * - The reconciliation order is already finalized
+  * - Database transaction fails
+  */
+export async function deleteReconciliationOrder(db: DB, id: number): Promise<void> {
+	const reconOrder = await db.execO<ReconciliationOrder>("SELECT * FROM reconciliation_order WHERE id = ?;", [id]);
+
+	if (!reconOrder[0]) {
+		throw new Error(`Reconciliation order ${id} not found`);
+	}
```
ikusteu, line 233: Can we use `ErrReconciliationOrderNotFound` or `ErrReconciliationOrderFinalized` (depending on the case)...easier and more straight forward to test, provides for predictable expectations (when error handling) and already used in other places (already exists)

ikusteu (changes requested):
What @delasous said...plus this minor thing: use the "canned" recon order errors instead of a custom string + test for both cases

apps/web-client/src/lib/db/cr-sqlite/__tests__/reconciliation-order.test.ts
delasous: I think something should be returned to the caller. Does the transaction fail if a matching row isn't found?

fadwamahmoud: no

PR #779, opened 2025-02-28
Publish e2e tests and add URL to the PR

silviot:
Currently each push gets built and deployed to test.libroc.co.

This PR adds the upload of the e2e test results, so they can be inspected more easily.

Since this PR is based off the `fix/test-publishing` branch, the deployed code is at https://test.libroc.co/fix/test-publishing/ and test results at https://test.libroc.co/fix/test-publishing/tests/ [ if you see the app instead of the tests results try an incognito window - I know, I know - we need to fix our service worker]


ikusteu:
LGTM, on that regard, we should really rethink the whole helpers API thing...as this would have been much more useful if we were able to receive actual source (test code, not the helpers internals)

PR #781, opened 2025-03-01
e2e test fix

ikusteu:
Fix ambiguity with quantity cell (`2`) in reconciliation order tests 

Wait for 100ms before navigation (to try and make sure the indexeddb txns have passed before navigation)


silviot (approved):
LGTM. I do dislike waiting, but I think failing tests are worse, so for lack of a better solution this works and gets the job done.

PR #782, opened 2025-03-01
Cleanup: stale artifacts

ikusteu:
This lays a ground work for other cleanup related PRs by removing unused artefacts (fewer files depending on refactors and artefacts cleaned up afterwards)

PR #783, opened 2025-03-01
Fix/quick checkout functionality

ikusteu:
Depends on #782 

We have a button, in inventory page, that takes you to quick checkout: new outbound note. This functionality was broken and, while cleaning up the stale db-related artefacts, this poped up. 

This small PR reinstates the functionality passing (and slightly prop drilling) the `createOutboundNote` function to the `Page` component from each inventory view


silviot (approved):
I did not look into the code, but verified it's working TTW (is Through The Web a known enough acronym? It is in the Plone world, so I tend to use it).

I opened #790 to add it to the desktop menu.

PR #787, opened 2025-03-04
Disable reconcile button on reconciled orders in supplier view

### LINKED ISSUE #763: Improve order list in `/orders/suppliers/<supplierId>`
Currently `/orders/suppliers/<supplierId>/` does not detect when a reconciliation has already started. It still shows a (working - it creates a new reconciliation object) reconcile button.
For orders that have a corresponding reconciliation the "reconcile" button should be removed, and in its place there should be a link to the reconciliation screen at `/orders/suppliers/reconcile/<orderId>/`. It makes sense to style it a bit differently from a button since it navigates to a new view without writing to the db, unlike the "reconcile" button.

fadwamahmoud:
Fixes #763

### DIFF
--- a/apps/e2e/integration/supplier_order.spec.ts
+++ b/apps/e2e/integration/supplier_order.spec.ts
@@ -180,7 +180,7 @@ testOrders(
 		});
 
 		// Navigate to supplier's orders view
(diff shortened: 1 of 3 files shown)


silviot (approved):
LGTM

PR #793, opened 2025-03-05
View Reconciliation instead of disabled reconcile button

### LINKED ISSUE #792: Enhancement: suppliers view
instead of disabling reconcile button we need a "view order" button instead

fadwamahmoud:
Fixes #792

### DIFF
--- a/apps/e2e/integration/reconciliation_order.spec.ts
+++ b/apps/e2e/integration/reconciliation_order.spec.ts
@@ -593,7 +593,7 @@ testOrders("should allow supplier orders to be reconciled again after deletion",
 	await page.keyboard.press("Enter");
 
 	// Delete and verify all supplier orders can be reconciled again
-	await page.getByRole("button", { name: "Delete reconciliatoin order" }).click();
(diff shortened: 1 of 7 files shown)


apps/e2e/integration/reconciliation_order.spec.ts
```diff
@@ -593,17 +593,23 @@ testOrders("should allow supplier orders to be reconciled again after deletion",
...
-	await page.getByRole("button", { name: "Delete reconciliatoin order" }).click();
+	await page.getByRole("button", { name: "Delete reconciliation order" }).click();
 	await page.getByRole("button", { name: "Confirm" }).click();
 
 	await expect(page.getByRole("dialog")).toBeHidden();
 
-	// Verify back at supplier orders
-
+	await page.waitForURL("**/orders/suppliers/orders/");
+	await page.waitForTimeout(1000);
```
silviot, line 602: Is this really necessary? Also next line's `page.reload()` is less than ideal (I understand it's not part of this unit of work, but it would be nice to remove it if not 100% needed).

silviot (approved):

PR #806, opened 2025-03-07
Review/reconciliation e2e pt1

ikusteu:
This is a soft refactor of reconciliation e2e tests:
- makes sure we always click on same supplier order(s) when reconciling (independent of the order in which those are listed)
- asserts the lines (when scanning/comparing) in an order-agnostic way (similar to previous point)
- removes `firstOrder`, `secondOrder` and such...(using the particular matcher)

NOTE: As stated: this is a really soft refactor. It doesn't change the testing logic, unless absolutely necessary in order to make the tests not fail. I'm working on a continuation of this: refactoring the tests altogether - kinda like what we did for unit tests - make the tests make more sense, cover potential edge cases and so...

### DIFF
--- a/apps/e2e/helpers/fixtures.ts
+++ b/apps/e2e/helpers/fixtures.ts
@@ -60,20 +60,21 @@ type FixtureSupplierOrder = {
 		id: number;
 		supplier_id: number;
 		supplier_name: string;
+		totalBooks: number;
 	};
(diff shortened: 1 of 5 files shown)


ikusteu:
Merging this in as it doesn't break anything, it's long overdue, and I don't want it to go out of sync with incoming work

PR #808, opened 2025-03-09
Use handlers to populate database

### LINKED ISSUE #722: Fix: debug page

fadwamahmoud:
Fixes #722


chiruzzimarco (commented):
This works perfectly and looks good.

My only concerns are:
* we are adding logic to something that should be as simple as possible
* we are validating known static data, which is validated again when the "real" upsert functions are called
* we still need to update debug data if the app models are changed, which i guess is a necessary evil without relying on some library to generate random fake data from factories out of types. Additionally we may need to maintain handlers code now instead of just updating some raw SQL

If such concerns are considered unimportant and can be dismissed then this LGTM.

apps/web-client/src/routes/debug/+page.svelte
```diff
@@ -89,11 +101,29 @@
...
 		console.log("Populating database");
 		try {
-			await db.db.exec(debugData);
+			await upsertBooks(db, dd.books);
+			await upsertCustomers(db, dd.customers);
+			await upsertSuppliers(db, dd.suppliers);
+			await associatePublishers(db, dd.supplierPublishers);
+			// possible order lines
+			const isbns = dd.customerOrderLines.map((line) => line.isbn);
+			await addBooksToCustomer(db, dd.customers[0].id, isbns);
```
chiruzzimarco, line 114: This is adding all books to the first customer order regardless of the `customer_id` specified on the customer order line

apps/web-client/src/lib/__testData__/debugHandlers.ts
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * @param {DB} db - Database connection
+ * @param {BookData[]} books - Array of book metadata objects
+ * @throws {Error} If any book is missing an ISBN
+ * @returns {Promise<void>}
+ */
+export async function upsertBooks(db: DB, books: BookData[]): Promise<void> {
+	if (!books.length) {
+		return;
+	}
```
ikusteu, line 58: I think this should be done directly in debug page:
- this function is defined here, along with the data
- it is used in only one place, with data defined right next to it
- uses `db.tx` lightly, which could cause problems if the underlaying function (e.g. `upsertBook`) uses `db.tx` as  well -- this isn't the case here, but there are other functions that do
- no need for checks/error aggregation: they will be performed (if legal) in the same way as any other update, in case of an error: it will be reported as any other update

apps/web-client/src/lib/__testData__/debugHandlers.ts
```diff
@@ -0,0 +1,226 @@
...
+
+	return db.tx(async (txDb) => {
+		for (const supplier of suppliers) {
+			await txDb.exec(
+				`INSERT INTO supplier (id, name, email, address)
+VALUES (?, ?, ?, ?)
+ON CONFLICT(id) DO UPDATE SET
+name = COALESCE(?, name),
+email = COALESCE(?, email),
+address = COALESCE(?, address);`,
```
ikusteu, line 103: The missing data (or non-existence of it) should be handled in data setup (definition), rather than within the function.

This function uses raw SQL, this is not much different from using a full SQL file for db seed (which is the functionality we're replacing here)

ikusteu (changes requested):
I wrote some comments on some handlers, but the reasoning applies to all handlers:
- if we're running this functionality (using predefined handlers), we should use them directly, not add additional complexity around them
- if we're seeding the DB using app's handlers, we're trying to future-proof this (as compared to using raw SQL) - in that case, no raw SQL should be used as it defeats the whole purpose

PR #812, opened 2025-03-10
Fix new note button not working

### LINKED ISSUE #796: bug: inventory: 'New note' button in warehouse page view doesn't work
`/inventory/warehouses/:id` - If there's not inventory to show, the page shows a 'New note' button. Clicking this button doesn't do anything. 

I didn't investigate further, but seeing as we have more of the buttons with same behaviour scattered around the app (which do work), I think this should be as easy as hooking up the function to the button.

fadwamahmoud:
Fixes #796


ikusteu (approved):
Seems ok, and works

PR #814, opened 2025-03-11
Create e2e tests for editing note title

### LINKED ISSUE #797: bug: inventory: note page (both inbound and outbound): the text editable doesn't persist name changes
In both inbound and outbound view, the note title is an editable text, which should save the updated title on enter in edit mode. This doesn't happen.

Lead: there's an event dispatched when that happens, which is logged out to the console.

- [x] make this work
- [ ] write a test
- [x] remove the aforementioned console.log

fadwamahmoud:
Fixes #797


ikusteu (approved):
LGTM

PR #834, opened 2025-03-12
Update outbound note warehouse on default warehouse dropdown

### LINKED ISSUE #798: bug: inventory: outbound note view: default warehouse selector has no effect
I outbound note view, there's a dropdown for the default warehouse: that should be persisted, but selecting it does nothing.

NOTE: I think we're not storing this into the DB, but are, rather using a persisted store (saving state into local storage). I might be wrong on this, but this is how I remember it and it makes sense...

fadwamahmoud:
Fixes #798

### DIFF
--- a/apps/e2e/integration/outbound_flow.spec.ts
+++ b/apps/e2e/integration/outbound_flow.spec.ts
@@ -1,4 +1,4 @@
-import { test } from "@playwright/test";
+import { expect, test } from "@playwright/test";
 
 import { baseURL } from "./constants";
 import { assertionTimeout } from "@/constants";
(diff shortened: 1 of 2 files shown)


apps/e2e/integration/outbound_flow.spec.ts
```diff
@@ -255,3 +255,58 @@ test("should display book original price and discounted price as well as the war
 		.assertRows([{ isbn: "1234567890", price: { price: "(€12.00)", discountedPrice: "€10.80", discount: "-10%" } }]);
 });
 // TODO: Test renaming using the editable title
+test("should update default warehouse for outbound note using dropdown", async ({ page }) => {
+	await page.goto(baseURL);
```
ikusteu, line 259: I don't think this tests renaming using editable title

ikusteu (changes requested):

apps/e2e/integration/outbound_flow.spec.ts
ikusteu: Right...this is a TODO, never mind

ikusteu (approved):
LGTM, fixed small conflict, waiting for CI and merging

PR #840, opened 2025-03-13
fix: navigate back to supplier orders page after finalizing order reconciliation:

### LINKED ISSUE #830: bug: orders: commit reconciliation order doesn't navigate
I feel it should navigate back to supplier orders or something...

ikusteu:
* apply the fix + add e2e test

Fixes #830

### DIFF
--- a/apps/e2e/integration/reconciliation_order.spec.ts
+++ b/apps/e2e/integration/reconciliation_order.spec.ts
@@ -1621,6 +1621,14 @@ testOrders("commit: applies delivery updates to customer order lines", async ({ 
 	await dialog.getByRole("button", { name: "Confirm" }).click();
 	await dialog.waitFor({ state: "detached" });
 
+	// Should have navigated back to supplier orders
+	//
+	// These buttons should be enough to conclude we're back at the supplier orders page
+	await page.getByRole("button", { name: "Unordered", exact: true }).waitFor();
+	await page.getByRole("button", { name: "Ordered", exact: true }).waitFor();
+	await page.getByRole("button", { name: "Reconciling", exact: true }).waitFor();
+	await page.getByRole("button", { name: "Completed", exact: true }).waitFor();
+
 	// Navigate to customers page and check customers
 	//
 	// NOTE: At the time of this writing, this is the state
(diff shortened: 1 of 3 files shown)


apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte
```diff
@@ -43,8 +43,6 @@
 		// Unsubscribe on unmount
 		disposer();
 	});
-	//#endregion reactivity
```
Comment (silviot) [inline apps/web-client/src/routes/orders/suppliers/reconcile/[id]/+page.svelte:46]: The corresponding opening tag on line 32 should be removed as well.

apps/e2e/integration/reconciliation_order.spec.ts
```diff
@@ -1706,3 +1706,48 @@ testOrders("commit: applies delivery updates to customer order lines", async ({
...
+	await dialog.waitFor({ state: "detached" });
+
+	// Should have navigated back to supplier orders
+	//
+	// These buttons should be enough to conclude we're back at the supplier orders page
+	await page.getByRole("button", { name: "Unordered", exact: true }).waitFor();
+	await page.getByRole("button", { name: "Ordered", exact: true }).waitFor();
+	await page.getByRole("button", { name: "Reconciling", exact: true }).waitFor();
+	await page.getByRole("button", { name: "Completed", exact: true }).waitFor();
+});
```
silviot, line 1753: This is the actual "meat" of the test: making sure we're redirected after confirming the reconciliation.

I'm thinking about pros and cons of this approach (a single test is responsible for a single piece of behaviour) vs adding assertions to existing tests.

My instinct goes to keeping the number of tests in check, so I'd prefer having this added as part of the existing three tests.

Two of them state `// Should navigate to supplier orders` and then proceed to click a UI link to navigate to a different page. I think replacing that comment with these checks would have the advantage of not introducing an additional test. And the disadvantage that a single test has too many responsibilities.

silviot (changes requested):

apps/e2e/integration/reconciliation_order.spec.ts
ikusteu: Right...so, this was done as part of this work (to provide a test for the functionality). It could have been added to an existing test, but the existing tests were under refactor. Afterwards, the refactor came in and became part of `main` (and, by rebase, this branch).
I've moved this assertino to an existing test

silviot (approved):

PR #841, opened 2025-03-13
Display correct book count in commit message

### LINKED ISSUE #800: bug: inventory: commit message - wrong book count
When committing a note, the commit message reads `<number of books> books will be removed from stock`. 

The message is fine, but the `n` represents total number of rows, not total quantity. I think it should represent the total quantity. 

NOTE: `totalBooks` shown on warehouse/note list (for the respective entity) show correct quantity

fadwamahmoud:
Fixes #800


ikusteu (changes requested):
Looks good! Can you do the same for outbound note as well?

ikusteu (approved):

PR #842, opened 2025-03-13
Allow zero price to be shown in bookPriceCell

### LINKED ISSUE #801: potential bug/proposal: custom item price defaults to a non value: make it 0
When adding a custom item, we're presented with a form. If we fill only the title field, the price is a non value (`null`/`undefined`) and displayed as `N/A`.

I think we should either:
- not allow form submission if the price is 0
- default to 0 -- accept 0 in the form, not discard it

### LINKED ISSUE #802: potential bug/proposal: inventory: have book prices default to 0
If there's no book data for a book present in inbound/outbound note, the price is shown as `N/A`. This makes it challenging when calculating totals (for committing/printing of receipt).

Proposed solutions:
- price the book as 0 (not really useful in real world, imo)
- disallow committing of a note containing a book without the price defined (I think this makes more sense)

fadwamahmoud:
Fixes #801 
Fixes #802

### DIFF
--- a/apps/e2e/helpers/table.ts
+++ b/apps/e2e/helpers/table.ts
@@ -57,7 +57,7 @@ export function getInventoryTable(parent: DashboardNode, view: InventoryTableVie
 
 const defaultInventoryRowValues: InventoryRowValues = {
 	isbn: "",
-	price: "N/A",
+	price: "€0.00",
 	quantity: 1,
 	title: "N/A",
 	year: "N/A",
--- a/apps/web-client/src/lib/components/Tables/InventoryTables/BookPriceCell.svelte
+++ b/apps/web-client/src/lib/components/Tables/InventoryTables/BookPriceCell.svelte
@@ -22,5 +22,5 @@
 {:else}
 	<!-- Price only is shown regardless of the row being a book row without discount applied, or a custom item row -->
 	{@const { price } = data}
-	<span data-property="full-price">{price ? `€${price?.toFixed(2)}` : "N/A"} </span>
+	<span data-property="full-price">{price || price === 0 ? `€${price?.toFixed(2)}` : "N/A"} </span>
 {/if}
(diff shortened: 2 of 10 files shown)


apps/e2e/helpers/table.ts
```diff
@@ -57,7 +57,7 @@ export function getInventoryTable(parent: DashboardNode, view: InventoryTableVie
 
 const defaultInventoryRowValues: InventoryRowValues = {
 	isbn: "",
-	price: "N/A",
+	price: "€0.00",
 	quantity: 1,
```
ikusteu, line 61: I think this should be done at query level (`COALESCE(book.price, 0) AS price`).

ikusteu (changes requested):
I gave it a quick look:
- the e2e updates look good
- as for the fallback: make sure book price falls back to 0 in all queries related to inventory side of things -- this will take case of the fallbacks added to the UI (which can be reverted then)

fadwamahmoud:
> I gave it a quick look:
> 
>     * the e2e updates look good
> 
>     * as for the fallback: make sure book price falls back to 0 in all queries related to inventory side of things -- this will take case of the fallbacks added to the UI (which can be reverted then)

not sure which ones should be reverted

apps/web-client/src/lib/components/Tables/InventoryTables/BookPriceCell.svelte
```diff
@@ -22,5 +22,5 @@
 {:else}
 	<!-- Price only is shown regardless of the row being a book row without discount applied, or a custom item row -->
 	{@const { price } = data}
-	<span data-property="full-price">{price ? `€${price?.toFixed(2)}` : "N/A"} </span>
+	<span data-property="full-price">{price || price === 0 ? `€${price?.toFixed(2)}` : "N/A"} </span>
```
ikusteu, line 25: No need for fallback here (`N/A`) -- prices should always be defined (fall back to 0) at query level. If, by chance, this fails, the e2e will pick it up (expecting `€0` and not `N/A` -- or in faulty case `€undefined`)

ikusteu:
Besides the UI fallbacks. This PR seems to touch on both custom item and book prices. It mentions bookPriceCell (in the title) and says it fixes #801 

I think it's perfectly fine it kills two birds with one stone (book and custom item prices), but make sure:
- that is communicated in PR message
- the appropriate tickets get closed by merging

ikusteu (approved):

PR #843, opened 2025-03-14
Review/reactivity 1

### LINKED ISSUE #825: bug: orders: new supplier order is not reactive
If trying to create a new supplier order, in new-order view, we see possible supplier order lines. If I add an additional customer order line (in a different tab), I don't see the update (in new-order tab) until I refresh the page.

### LINKED ISSUE #836: bug: orders: customers reactivity
The customer list doesn't get updated (without reload) if I add lines to a customer

ikusteu:
Fixes #825 
Fixes #836 (doesn't really fix, but tests that the observed bug isn't there)

### DIFF
--- a/apps/e2e/helpers/fixtures.ts
+++ b/apps/e2e/helpers/fixtures.ts
@@ -130,7 +130,8 @@ type OrderTestFixture = {
 	/**
 	 * Data:
 	 *
-	 * id: 1, name: "Sup1", email: "sup1@gmail.com"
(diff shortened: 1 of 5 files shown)


silviot (approved):
Works great!

And I'm approving this!

But I can't help pointing out that this PR does the job in about 30 lines of code. And there are slightly less than 200 lines for e2e tests.

Note that I understand this is a kind of a special case, and needs to be different from other tests. I mention this to share my overall "background" concern.

PR #845, opened 2025-03-14
Coalesce price when caclculating totalCoverPrice and totalDiscountedP…

### LINKED ISSUE #803: bug: inventory: note history crashes if total prices aren't defined
Related to #801 and #802. If a note contains a book (or custom item) without a defined price, the total isn't calculated (it's `null`). This crashes the `history/notes/:date` view (if at least one note in the list doesn't have a defined value).

The error reports `null.toFixed()` is not a function (which makes sense) 

Proposed solutions:
- [ ] (quick fix) - default to 0 in such cases (or account notes with `null` values for totals as 0, allowing at least some insight into the totals of other notes)
- [ ] (fix the root cause) - see #801, #802

fadwamahmoud:
Fixes #803

### DIFF
--- a/apps/e2e/integration/history.spec.ts
+++ b/apps/e2e/integration/history.spec.ts
@@ -431,6 +431,13 @@ test("history/isbn - search results", async ({ page }) => {
 			authors: "Author 2",
 			publisher: "Publisher 2",
 			year: "2022"
+		},
+		{
+			isbn: "3333333333",
+			title: "Book 3",
+			authors: "Author 3",
+			publisher: "Publisher 3",
+			year: "2023"
 		}
 	]);
 });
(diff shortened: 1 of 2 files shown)


ikusteu (changes requested):
Looks good, can you add a test?

silviot:
...or add an assertion to an existing test? Ivan and myself had quite a discussion today about what should be preferred (I'm pushing for fewer tests that do more assertions).

ikusteu:
- the E2E is a bit of an overkill here.
- the commit containing the `COALESCE` update DOES fix the issue, but replicating that in the unit test took quite some work
- merging a branch into another non-main shouldn't close the issue
- it makes sense to detach this from the other 0 - price related work (as it does fix a separate issue)

so...I'm closing this, in favour of #860 -- it contains the fix commit (cherry picked), contains the work to this single issue, adds unit tests and is open against `main`

PR #849, opened 2025-03-18
Fix/possible orders granularity

### LINKED ISSUE #826: bug / proposal: orders: new supplier order line partitioning
When creating a supplier order, we're presented with possible lines (with checkboxes) for choice of the lines we want to include in the order: 1/4, 1/2, 3/4. However, the ratio is calculated based on order lines. I think there could be some improvements there:

~~- when calculating the cut, we might want to either take into account total books (and divide on that measure), or use total price and fit the first n **books** (not lines) that fit the price
- allow for partial selection (the former wouldn't work without this): say there's a possible line for a book, with quantity 100, currently we can either order full 100, or not order (the line) at all~~

**edit:**
- don't allow for explicit selection (quantity field), just partition by book count
- see #850 for enhancement

### LINKED ISSUE #828: bug: orders: new supplier order - total selected counts lines, not quantity
It would make more sense if total selected quantity is counted, not number of selected lines

ikusteu:
Fixes #826 
Fixes #828

### DIFF
--- a/apps/e2e/integration/supplier_order.spec.ts
+++ b/apps/e2e/integration/supplier_order.spec.ts
@@ -9,7 +9,7 @@ import {
 	createSupplierOrder,
 	finalizeReconciliationOrder
 } from "@/helpers/cr-sqlite";
-import { testOrders } from "@/helpers/fixtures";
+import { depends, testOrders } from "@/helpers/fixtures";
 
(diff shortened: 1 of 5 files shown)


ikusteu:
Everyone agreed on this functionality (as a stepping stone towards #850 ). Merging as it's been open forever.

PR #861, opened 2025-03-21
Enhance/empty reconciliation order

### LINKED ISSUE #854: It should be possible to reconcile an order without a single book delivered
Sometimes you send out an order to a supplier and then realize there was a mistake of some kind. You then proceed to cancel the order.

We should support this use case somehow.

The simplest thing that comes to mind is showing the `Compare` and then `Commit` buttons even if there are no book scanned for the reconciliation, i.e. when not a single book for the order has been delivered from the supplier.

ikusteu:
Fixes #854

### DIFF
--- a/apps/e2e/integration/reconciliation_order.spec.ts
+++ b/apps/e2e/integration/reconciliation_order.spec.ts
@@ -1575,8 +1575,7 @@ testOrders(
 	}
 );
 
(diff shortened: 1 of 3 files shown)


ikusteu:
Does what it's supposed to do + I have an e2e test to back that up.

PR #864, opened 2025-03-24
Enhance/new supplier order price partitioning

### LINKED ISSUE #850: enhance: orders: suppliers: new-order: partition by price
In new-supplier-order view, when presented with possible order lines, we're also presented with the choice of selecting a subset of possible orders: 1/4, 1/2, 3/4, all

We made the selection pick a slice based on book count (see #849). 

The proposed enhancement is to:
- calculate the full price of all possible orders (already in)
- select the slice based on price: 1/4 = Math.flor(total possible price / 4) --> Go down the list of possible orders and fit as many as possible in the selected price

ikusteu:
Fixes #850


ikusteu:
Does what it's supposed to and has tests to back that up. (If something's wrong, we can fix later)

PR #876, opened 2025-03-25
Feature/differentiate unassociated publishers

### LINKED ISSUE #873: Differentiate between unassociated publishers and those associated to different suppliers
Make sure user can tell the difference in the UI
If user is trying to change the association on one of the publishers show a confirmation dialog

fadwamahmoud:
Fixes #873


apps/web-client/src/routes/orders/suppliers/[id]/+page.svelte
```diff
@@ -207,6 +217,37 @@
...
+							<div class="h-max-[164px] w-full overflow-y-auto">
+								<tbody>
+									{#each publishersAssignedToOtherSuppliers as publisher}
+										<tr class="hover focus-within:bg-base-200">
+											<td class="px-2">{publisher}</td>
+											<td class="px-2 text-end"
+												><button
+													on:click={() => {
+														confirmationPublisher = publisher;
+														confirmationDialogOpen.set(true);
```
Comment (ikusteu) [inline apps/web-client/src/routes/orders/suppliers/[id]/+page.svelte:239]: When you resolve conflicts, there will be an error (probably coming from updated deps) stating that `div`, can't be nested under `table`, `thread`, `tbody`, etc. 
You can use the other two tables in `main` as a template (it was annoying and painful)

ikusteu (changes requested):

PR #883, opened 2025-03-28
Feat/import export json

### LINKED ISSUE #880: db sync: import / export: json
Implement a DB import / export functionality in JSON:

- [ ] add logic to dump the DB as JSON
- [ ] add logic to seed the DB using JSON data
- [ ] replace the UI (old -- SQLocal) to utilise the new functionality

ikusteu:
Fixes #880 

The settings page UI is under construction somewhat:
- the old DB management section is used, with the subset of functionality
- the list of DBs shows only the current DB (with selection disabled) - this is a TODO for a later date
- there's no DB delete functionality - this is a TODO for a later date
- to "import" the DB, click 'Import' and drag the `json` file to the drop area
- to "export" the DB, hover over the (only) DB line item (in DB selection area) and click the download button

NOTE: as #880 specifies, the DB is dumped in `json` format
NOTE: the "import"/"export" aren't full DB import/export, but rather data dumps:
- no schema is dumped
- no crsqlite data is dumped
- only the inserted data is dumped
- loading of the data doesn't create a new DB, but rather populates the existing one
- in case of conflict, the imported data takes precedence (`ON CONFLICT DO UPDATE SET '<column>' = excluded.<column>`)

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/import_export.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/import_export.ts
@@ -37,7 +37,7 @@ export async function loadData(db: DB, data: DatabaseDump): Promise<void> {
 		const stmt = [
 			`INSERT INTO ${table} (${keys.join(", ")}) VALUES`,
 			Array(rows.length).fill(placeholderLine).join(",\n"),
-			`ON CONFLICT SET ${keys.map((key) => `${key} = EXCLUDED.${key}`).join(", ")}`
+			`ON CONFLICT DO UPDATE SET ${keys.map((key) => `${key} = EXCLUDED.${key}`).join(", ")}`
 		].join("\n");
 
 		await db.exec(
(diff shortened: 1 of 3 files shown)


ikusteu:
> I tried this on https://test.libroc.co/feat/import-export-json/settings/
> 
> It produced a nice JSON as expected, but this error popped up in the console:
> 
> ```
> import_export.ts:50 Uncaught (in promise) NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
>     at HTMLButtonElement.<anonymous> (import_export.ts:50:87)
> ```

Right, forgot about that...The initial code was AI assisted (if I remember), I've just adjusted it now (as it made sense). I saw the error, but it didn't crash, so I went on with the implementation. Forgot to massage this.

ikusteu:
Infeasible/not-doable in this manner. Working on pure SQL import (as a single file -- needs OPFS, closing this as it's not an appropriate route)

PR #886, opened 2025-03-29
Feat/import export

### LINKED ISSUE #885: db sync: import / export: sql
#883 adds import/export functionality in JSON format, as a next step, we should make it possible to create the same functionality for `sql`

ikusteu:
Includes #883 
Fixes #885 

Both formats are kept (json and sql).
There are two buttons: one for json, one for sql export

The import/export functions accept additional options (for future use):
- dump only user data (tables) or full (including crsql master, etc.)
- when constructing sql (either loading json, or constructing sql dump): it could include schema definitions or data-only

Currently wired in import/export dumps user only data and loads data only (to an existing DB). The rest of the functionality is preparation for full DB imports

### DIFF
--- a/apps/web-client/package.json
+++ b/apps/web-client/package.json
@@ -37,7 +37,9 @@
 		"just-compare": "~2.3.0",
 		"nanoid": "^5.0.5",
 		"svelte-local-storage-store": "~0.6.4",
-		"typesafe-i18n": "~5.26.2"
+		"typesafe-i18n": "~5.26.2",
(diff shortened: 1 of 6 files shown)


silviot:
How can I do a round trip export → edit the exported data → import to test this?

ikusteu:
Infeasible/not-doable in this manner. Working on pure SQL import (as a single file -- needs OPFS, closing this as it's not an appropriate route)

PR #893, opened 2025-04-02
Feature/dictionaries orders

### LINKED ISSUE #748: Feat: finish adding app dictionaries
WIP: needs detail

`typesafe-i18n` is setup, and it is being used in a couple places. We need to finish this

fadwamahmoud:
Fixes #748 
Pt 2 Dictionaries for orders: suppliers and its contents


delasous (changes requested):

pkg/shared/src/i18n/en/index.ts
delasous, line 1: - the dialog objects should either be nested in a page where they are displayed, or nested in a top-level `common` object if they are used in multiple locations
- the pages within the `suppliers_page` should be moved to the top level. A few of these keys - `new_order_page` and `order_list_page` particularly - could receive the same treatment of giving `label`, `stat` or `placeholder_box` context treatment to the keys

delasous (changes requested):
The dictionary already feels much better, but I'd say it could use one more pass over with the comments I've outlined above

PR #897, opened 2025-04-06
Feature/dictionaries warehouses

### LINKED ISSUE #748: Feat: finish adding app dictionaries

fadwamahmoud:
Fixes #748 

Pt 3: Warehouses and Outbound


apps/web-client/src/lib/dialogs.ts
```diff
@@ -1,55 +1,66 @@
...
+
 export interface DialogContent {
 	onConfirm: (closeDialog: () => void) => void;
 	title: string;
 	description: string;
 }
 
+let t: TranslationFunctions;
+
+LL.subscribe((LL) => {
```
delasous, line 12: We shouldn't have a subscription without an unsubscription. I think this file was setup as an intermediary dictionary for dialog strings - do we need it anymore? Aren't we passing the same strings directly to the dialogs on the pages? If not, and we still need this, the setup needs to be rethought to avoid this loose subscription

pkg/shared/src/i18n/en/index.ts
```diff
@@ -137,11 +141,225 @@ const inbound_note = {
...
+		total_delivered: "Total delivered",
+		delete_dialog: {
+			confirmDeleteDialogHeading: "Delete Reconciliation Order",
+			confirmDeleteDialogDescription:
+				"Are you sure you want to delete this reconciliation order? This action will delete all the scanned lines."
+		}
+	}
+};
+
+const warehouses = {
```
delasous, line 237: Create separate `warehouse_list_page` and `warehouse_page` objects

pkg/shared/src/i18n/en/index.ts
```diff
@@ -137,11 +141,225 @@ const inbound_note = {
...
+	warehouse_id: {
+		display_label: {
```
delasous, line 239: `column_headers`, or whatever other standard label you've used for another table?

pkg/shared/src/i18n/en/index.ts
```diff
@@ -137,11 +141,225 @@ const inbound_note = {
...
+	stats: {
+		books: "books",
+		discount: "discount"
+	},
+	labels: {
+		new_note: "New note",
+		view_stock: "View Stock"
+	}
+};
+const dialog_title = {
```
delasous, line 270: I'd flip this and keep the title and description next to the specific dialog `commit_inbound_dialog: { title, description }` or `dialogs: { commit_inbound: { title, description }}`. If the dialogs appear in a few places, I'd place them in an object called `common`. If they appear on a single page, nest them there

delasous (changes requested):

Issue #910, opened 2025-04-11
optimise: page load times


ikusteu:
Filled with test data, the DB is extremely sluggish at points. This unit of work is aimed at optimising query performance and making sure the load functions (ideally) take less than 300ms (human eye and all that).

#909 creates indexes that would have been there if we were allowed to use foreign keys, with the following benchmarks:

## Current benchmarks

Route: /stock:
  - getPublisherList: 125 ms
  - getStock: 168 ms
  - **total load: 126 ms**

Route: /inventory/warehouses:
  - getAllWarehouses: 533 ms
  - **total load: 533 ms**

Route: /inventory/warehouses/[id]:
  - getWarehouseById: 2 ms
  - getStock: 431 ms
  - getPublisherList: 150 ms
  - **total load: 583 ms**

Route: /inventory/inbound:
  - getActiveInboundNotes: 1157 ms _(note: with all notes set to committed = 0, probably won't be as many in production)_
  - **total load: 1158 ms**

Route: /inventory/inbound/[id]:
  - getNoteById: 3 ms
  - getNoteEntries: 3 ms
  - getPublisherList: 146 ms
  - **total load: 152 ms**

Route: /outbound:
  - getActiveOutboundNotes: 774 ms _(note: with all notes set to committed = 0, probably won't be as many in production)_
  - **total load: 774 ms**

Route: /outbound/[id]:
  - getNoteById: 2 ms
  - getAllWarehouses: 372 ms
  - getNoteEntries: 3 ms
  - getStock: 5 ms
  - getNoteCustomItems: 1 ms
  - getPublisherList: 134 ms
  - **total load: 518 ms**

Route: /history/date:
  - getPastTransactions: 824 ms
  - **total load: 845 ms**

Route: /history/date/[date]:
  - getPastTransactions: 753 ms
  - **total load: 754 ms**

Route: /history/isbn:
  - searchBooks: 136 ms

Route: /history/isbn/[isbn]:
  - getPastTransactions: 5 ms
  - getBookData: 1 ms
  - getStock: 2 ms
  - **total load: 8 ms**

Route: /history/notes:
  - getPastNotes: 30 ms
  - **total load: 30 ms**

Route: /history/notes/[date]:
  - getPastNotes: 8 ms
  - **total load: 9 ms**

Route: /history/notes/archive/[id]:
  - getNoteById: 2 ms
  - getNoteEntries: 1 ms
  - getNoteCustomItems: 1 ms
  - **total load: 5 ms**

[...]

PR #916, opened 2025-04-15
Optimise/some query optimisations

ikusteu:
An effort to improve app load times (queries and such). As of now this is still WIP.

It has:
- updates to publisher retrieval - created an index - major speedup
- updates to warehouse list queries (and stock queries in general) - created a materialised calc table `book_stock` - major speedup to retrieval
- updates to note list queries - created a metarialised table `note_total_books` - major speedup
- data is consistent (even under sync) - the materialised tables aren't being synced (to avoid double addition/subtraction) and are updated by data triggers

Some challenges:
- data triggers make the sync teeeediously slow
- the infinite scroll is broken on note list views (starts at infinity 😂) so the load is really slow (due to paing, the load fn takes 250-280ms)
- warehouse total stock doesn't show anything (total stock = 0) after sync (even though a similar scenario passes in tests)

I would, by no means, recommend merging this in as it's a broken work in progress

**edit:** 
- tests are added for data consistency (didn't work at first)
- benchmarks coming soon: this should take care of most of the lag in #910 (provided I get it to work...)


silviot:
I have a hunch that infinite scroll is costing more than it provides. A standard set of pagination links would be ok with fewer problems.

ikusteu:
Stale, not doable/feasible in this way. I took a divergent route which had shown more potential/utility.

Issue #904, opened 2025-04-09
Unconfuse the outnote UI


silviot:
The UI for outgoing notes has a couple of confusing points:

* The "Custom item" button is not shown when the note is empty - it should 
* The "commit" button is not where I expect it to be. I keep pressing "custom item" instead

I propose we always display the "custom item" button, and that we display it on the left, in a different colour.

silviot:
Fixed in #917

PR #919, opened 2025-04-16
Automatic warehouse selection

### LINKED ISSUE #905: Out note warehouse detection not working
To reproduce:

* create a warehouse
* create and commit an in note with book 666 (Freakonomics)
* create an out note
* add book 666

Image

Even though librocco marks the book as present in "New warehouse", the `not-found` element is selected.

fadwamahmoud:
Fixes #905 and #904

Todo:
- [x] Tests
- [x] Dictionaries


silviot:
This assigns books to a warehouse having them even if the default warehouse is a different one, as long as the default does not have the book. As requested.

There are a couple more things it needs to do:

* Allow de-selecting the default warehouse
* If the book is not present in the currently selected default warehouse we should force a selection from the user

silviot (approved):

PR #926, opened 2025-04-22
Fix/playwright config

ikusteu:
For the most part, this PR serves to revert some changes to `playwright.config.ts` which I deem unnecessary, as well, as limiting for local env.

Furthermore, there was some ambiguity (which I couldn't put my finger on) around the port the app is served from:
- @silviot had made the determining of the base url (port `4173` or `5173`) easier and more automated
- the problem: it didn't seem to work (😂)
- the actual problem: we had two sources of truth for `baseURL`
- the solution: removed `e2e/integration/constants.ts` and made all tests use `e2e/constants.ts` (where the port check is made)
- works like a charm

Oh, and yeah: fix ambiguity with `CI` env variable:
- if `export CI=false`,  `process.env.CI == "false" == true`
- however if `export CI=false` and `IS_CI = process.env.CI === "true"`, `IS_CI == false`

### DIFF
--- a/apps/e2e/constants.ts
+++ b/apps/e2e/constants.ts
@@ -1,6 +1,6 @@
 import * as net from "net";
 
-export const IS_CI = process.env.CI === "true";
+export const IS_CI = /^(?:1|true|on)$/i.test(process.env.CI?.trim() ?? "");
 
 export function getPort(): Promise<number> {
 	const testSocket = new net.Socket();
--- a/apps/e2e/constants.ts
+++ b/apps/e2e/constants.ts
@@ -21,8 +21,7 @@ export function getPort(): Promise<number> {
 }
 
 const port = await getPort();
-export const baseURL = `http://localhost:${port}`;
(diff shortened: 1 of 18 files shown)


apps/e2e/constants.ts
```diff
@@ -1,5 +1,7 @@
 import * as net from "net";
 
+export const IS_CI = process.env.CI === "true";
```
silviot, line 3: I propose we also allow `True`, ` true`, `1`, `On`, `ON` and so on

```suggestion
export const IS_CI = /^(?:1|true|on)$/i.test(process.env.CI?.trim() ?? "");
```

silviot (approved):
LGTM

I requested a small change to be able to use more "truthy" values for `CI` (I understand Github Actions will always use `true`, but I find it nice to also be able to use `True` or `TRUE`.

Removing the second source of truth (the other `constants.ts` file) is very good!

PR #938, opened 2025-04-23
optimise: sluggish outbound scanning:

ikusteu:
* don't calculate 'totalBooks' for warehouses loaded in outbound view (not used and speeds up queries by orders of magnitute)
* create publisher index on 'book' table - much faster 'getPublishers' query

Works on my machine. Please try on yours.

The benchmarks before after (on outbound note page):
- `getAllWarehouses` - ~400ms -> ~5ms
- `getPublisherList` - ~75ms -> 14ms
- total load: ~500ms -> ~30ms


silviot (approved):
Testing this takes ages, so I'll merge it without clicking around. The code looks sound and tests are green.

PR #946, opened 2025-04-27
fix: optimise outbound load fn for empty notes:

### LINKED ISSUE #943: Improve outbound note view performace
Navigating to an existing outbound note in a populated db (from production) is extremely slow: 17 seconds on my not-that-slow ryzen 7.

This also happens when creating a new outnote. The user will press the button multiple times in frustration, resulting in multiple outbound notes created.

ikusteu:
* skip 'getStock' if there are no entries in the note (unnecessary and crazy expensive in case of an empty note)

Fixes #943 

In the load function we used to run something along the lines:
```ts
const _entries = await getNoteEntries()
const stock = await getStock({ isbns: entriesIsbns })
const entries = combineEntriesWithAvailableStock()
```
If entries is empty the expanded version looks something like this:
```ts
const _entries = []

// This is equivalent to running: getStock()
const stock = await getStock({ isbns: [] })

const entries = [] 
```
Running `getStock` with empty filters (and `[]` is akin to `undefined` in terms of `getStock` filters) retrieves the entire stock -- a crazy expensive query in filled DB. This explains why the load function of empty notes was terribly slow. In my experience, `getStock` for any reasonable number of entries in sufficiently fast (quite fast for up to 10, 20 entries), but let me know what your experiences are.

All in all, this definitely is an improvement, and, in my (aforementioned) experience it fixes the issue completely

### DIFF
--- a/apps/web-client/src/routes/outbound/[id]/+page.ts
+++ b/apps/web-client/src/routes/outbound/[id]/+page.ts
@@ -1,7 +1,7 @@
 import { redirect } from "@sveltejs/kit";
 
 import type { PageLoad } from "./$types";
-import type { Warehouse, DB, NoteCustomItem, NoteEntriesItem } from "$lib/db/cr-sqlite/types";
+import type { Warehouse, NoteCustomItem, NoteEntriesItem } from "$lib/db/cr-sqlite/types";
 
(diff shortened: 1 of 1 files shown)


silviot (approved):
This optimization fully solves the problem I reported. Thanks!

PR #951, opened 2025-05-03
Optimise/general 1

ikusteu:
Some general optimisations of app performance

- crate stock cache:
  - the full stock is cached
  - warehouse total books are derived from the cache
  - warehouse stock (for each warehouse) is derived from the cache
  - the cache contains promises (pending or resolved) so as to be able to return non-blocking data (promises handled in the UI)
  - the cache is invalidated at root `+layout.svelte` on each change to `note` table (we're aiming for committed `0` -> `1` - affecting the stock, but no finer granularity at this point)
  - additional to listening to DB updates to relevant tables for particular views, we're listening to stock cache `onInvalidated` as well (for views where matters) in order to recalculate the displayed data with respect to stock
  - the stock is disabled between views (and enabled on each view that requires it)
  - the stock is lazy: it will be invalidated even if disabled (just won't trigger a re-query until necessary)
- use cached stock to calculate `totalBooks` for each warehouse in `/inventory/warehouses` and `/history/warehouse` lists
- in the aforementioned views: the view gets loaded immediately whereas the expensive part (warehouse totals) is returned as a promise and, until resolved, `...` placeholder is rendered (click -> instant navigations)
- in warehouse stock view (`/inventory/warehouses/[id]`) the stock is derived from the cache, with book data being added at each load (somewhat cheap to calculate, based on limited ISBNs, as provided by stock, but fully reactive to changes to book data)
- when navigating `/inventory/warehouses` -> `/inventory/warehouses/[id]` preload the data (run load function - get book data) on "hover" for even snappier loading/navigation
- create an index on `note.committed_at` -> used for `/history/date` (speeds up `getPastTransactions` **~600ms** -> **~40ms**


silviot:
@coderabbitai full review

apps/web-client/src/lib/db/cr-sqlite/books.ts
```diff
@@ -75,6 +75,39 @@ async function _upsertBook(db: DB, book: BookData) {
 	);
 }
 
+async function _getMultipleBookData(db: DB, ...isbns: string[]): Promise<Array<Required<BookData> & { updatedAt: Date | null }>> {
+	const placeholders = isbns.map((_, i) => `(?, ${i})`).join(",\n");
+	const query = `
+		WITH input_list(isbn, ord) AS (
```
silviot, line 81: TIL about `WITH` in SQL. Nice! I always used a literal `IN` in those cases.

apps/web-client/src/lib/db/cr-sqlite/books.ts
```diff
@@ -175,5 +208,6 @@ const processRawBookRes = (res: RawBookRes): Required<BookData> & { updatedAt: D
 
 export const upsertBook = timed(_upsertBook);
 export const getBookData = timed(_getBookData);
+export const getMultipleBookData = timed(_getMultipleBookData);
```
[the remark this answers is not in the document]
ikusteu: yup, we're on the same page, but:
- `timed` was developed beforehand (just applied here)
- `@cached` would also be nice once the query-based caching is developed
- I went with the approach: 
  - refactor caching to be query based (instead of current data caching)
  - implement decorators for cached queries
  - implement the same for `timed` (I got the idea while thinking about caching, but saw `timed` as the perfect candidate as well)

apps/web-client/src/lib/db/cr-sqlite/books.ts
ikusteu: Felt the same way I saw it for the first time, thanks @fadwamahmoud

apps/web-client/src/lib/db/cr-sqlite/stock_cache.ts
```diff
@@ -0,0 +1,95 @@
...
+	if (!get(valid)) {
+		query.set(execQuery(db));
+	}
+};
+
+export const disable = () => dbStore.set(null);
+
+export const invalidate = () => {
+	// Invalidate the cache
+	valid.set(false);
```
silviot, line 70: Why do we have a boolean to tell us if the cached value is stale or valid?

My goto solution for this would be to remove the old state when it needs invalidation.

Are there edge cases I'm not thinking about?

apps/web-client/src/lib/db/cr-sqlite/stock_cache.ts
```diff
@@ -0,0 +1,95 @@
...
+	// Set the DB -- effectively enabling the cache
+	dbStore.set(db);
+
+	// If cache invalidated while not active, rerun the query
+	if (!get(valid)) {
+		query.set(execQuery(db));
+	}
+};
+
+export const disable = () => dbStore.set(null);
```
silviot, line 66: Related to the comment on `valid`: shouldn't we also set `valid` to `false` here given the current architecture?

ikusteu: No, it's a matter of keeping the caching logic as lazy as possible.

Consider this case:
- view A uses the cached data, view B doesn't 
- we switch from A -> B (cache is disabled)
- we switch to view C (which also uses cached data), it re-enables the cache and uses the cached value

Now, consider this:
- same setup: A, C - use the cache, B doesn't
- we're on view A, the cache gets invalidated
- since A uses the cache, we execute the (expensive) query to recalculate the state
- we switch from A -> B (cache is disabled)
- while on view B the cache gets invalidated, the query is not executed as it's expensive and unnecessary
- we switch to view C - we see that the cache is invalid, so we execute the expensive query

apps/web-client/src/routes/+layout.svelte
```diff
@@ -98,12 +110,19 @@
 		if (get(syncActive)) {
 			sync.sync(get(syncConfig));
 		}
+
+		// Control the invalidation of the stock cache in central spot
+		// Invalidate on every change to note (notes being committed)
+		// TODO: Implement more fine-grained control over this - potentially using a separate table to control committed state
+		disposer = dbCtx.rx.onRange(["note"], () => stockCache.invalidate());
 	});
```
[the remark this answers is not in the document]
ikusteu: This API doesn't support that. The first param to the callback function is actually a change type (INSERT, UPDATE, DELETE) and doesn't contain any information of the actual change.

apps/web-client/src/routes/inventory/warehouses/[id]/+page.ts
```diff
@@ -20,18 +25,42 @@ const _load = async ({ parent, params, depends }: Parameters<PageLoad>[0]) => {
...
+			discount: 0,
+			publisherList: [] as string[],
+			entries: new Promise<GetStockResponseItem[]>(() => {})
+		};
 	}
 
+	// Disable the stock cache to prevent the expensive stock query from blocking the
+	// DB for other (cheaper) queries necessary for the page load.
+	stockCache.disable();
+
```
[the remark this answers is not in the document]
Comment (ikusteu) [reply apps/web-client/src/routes/inventory/warehouses/[id]/+page.ts:41]: If an, unlikely, error happens here, we have bigger problems than caching as:
- the view will throw an error
- that error would need to be handled somehow
- every time a view needing cached data is loaded, the cache is enabled

silviot:
I have a hard time understanding the "enabled"/"disabled" state of the cache.

I usually expect this from a cache:

* it receives a request
* checks if it has the result ready - if it does it returns that and work is done
* if it doesn't have the result ready it fetches/computes it with a possibly expensive operation
* when it's notified that some cached result is stale it deletes it

and I'm not sure I understand the purpose of "enabling"/"disabling" the cache.
I have the impression that it's used to coordinate multiple potentially concurrent operations. If that's the case I feel like it should be more explicitly documented.

@coderabbit can you summarize for me how the cache enabling/disabling is used in this PR?

silviot:
I think I understand a bit better now.

I'm a simple man. If you tell me "I disabled the cache" my brain forms this thought:

> there used to be a cache enabled, to speed up repeated queries. Now it's disabled, so repeated queries are not sped up anymore

But what this "enable"/"disable" really means is:

> the cache has just been disabled. It means you won't be able to get any result from it, until it's enabled again

@coderabbit Did I nail it?

silviot:
To me the `enabled`/`disabled` naming in this context smells. I think it's misleading.
Like Peter Griffin was misled on a boat, I might be misled by this in the future:

> _Captain_: Who pooped on the poop deck?
> _Peter_: I did!
> _Captain_: You are disgusting!
> _Peter_: And you are misleading!

@coderabbit Can we find a better name for this feature?

silviot:
@ikusteu I propose to use `enableRefresh`/`disableRefresh` in place of `enable`/`disable` for the cache.

silviot (approved):

PR #954, opened 2025-05-03
Cleanup/small db cleanup

ikusteu:
This looks like a lot, but doesn't really do much, it:
- moves `$lib/db/cr-sqlite` -> `$lib/db` (as cr-sqlite is not the main and only implementation)
- removes `pouchdb` dependency 

This was done in an effort to remove stale DB code, but it turns out that that code had already been removed


ikusteu:
I'd pause this for now and rebase when all of the other PRs are in as it's a small update (repo organisation), but will produce conflicts at other (more elaborate) PR diffs

PR #969, opened 2025-05-24
Feat/sync progress 2

ikusteu:
Comes in after #968 

Control the sync process by controlling the transport:
- we wrap the (default) transport with our additional logic
  - as changes arrive via transport object, they arrive as a (potentially huge) list of changes
  - our wrapper splits the list into chunks (configurable by setting `maxChunkSize`) and feeds them to the changes applying process (`InboundStream` object) as chunks, consistent with the expected structure
  - this also speeds up the sync by an order of magnitude (wasn't intentional, but cool)

Update the way sync progress is communicated to the main thread: instead of signalling start/stop, it signals the current progress (`nProcessed` and `nTotal`).

Improve UX re: syncing - show a progress bar in the sync-in-progress window

Potential next steps: since we have this, and the sync is somewhat performant and not blocking, we might not need the progress dialog (blocking everything), but rather:
- show sync progress in the footer
- keep the unload preventing logic (while the sync is in progress)


silviot:
This improves UX immensely. And also performance.

I believe there are two low hanging fruits here:

* show numbers, besides the loading bar
* add a "Disable sync" button to the modal. It should finish processing the current chunk and discard the rest.

silviot (approved):

PR #976, opened 2025-06-11
Load all customers and filter them on search

### LINKED ISSUE #929: Shape up UI for customers when there are 1000-2000
Currently there is no "search" button. How will the app work with 2k customers?

fadwamahmoud:
WIP: Partly Fixes #929

### DIFF
--- a/apps/web-client/src/lib/forms/CustomerSearchForm.svelte
+++ b/apps/web-client/src/lib/forms/CustomerSearchForm.svelte
@@ -1,0 +1,39 @@
+<script lang="ts">
+	import { superForm } from "sveltekit-superforms/client";
+	import { QrCode } from "lucide-svelte";
+
+	import { testId } from "@librocco/shared";
+
+	import type { FormOptions, SuperValidated } from "sveltekit-superforms";
+	import type { CustomerSearchSchema } from "$lib/forms/schemas";
+
+	export let icon = QrCode;
+	export let placeholder = "Search for a customer by name";
+
+	export let data: SuperValidated<CustomerSearchSchema>;
+	export let options: FormOptions<CustomerSearchSchema>;
+
+	export let input: HTMLElement | undefined = undefined;
+
+	const form = superForm(data, {
+		...options
+	});
+
+	const { form: formStore, enhance } = form;
+</script>
+
+<form use:enhance method="POST" id="scan-form" class="h-full w-full">
+	<label class="input-bordered input flex flex-1 items-center gap-2">
+		<svelte:component this={icon} />
+		<input
+			name="fullname"
+			id="fullname"
+			required
+			autocomplete="off"
+			data-testid={testId("customer-search-form")}
+			bind:this={input}
+			bind:value={$formStore.fullname}
+			{placeholder}
+		/>
+	</label>
+</form>
--- a/apps/web-client/src/routes/orders/customers/+page.svelte
+++ b/apps/web-client/src/routes/orders/customers/+page.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { onMount, onDestroy } from "svelte";
-	import { Plus } from "lucide-svelte";
+	import { Plus, SearchIcon } from "lucide-svelte";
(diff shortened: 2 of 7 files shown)


apps/web-client/src/lib/forms/CustomerSearchForm.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+	import { superForm } from "sveltekit-superforms/client";
+	import { QrCode } from "lucide-svelte";
+
+	import { testId } from "@librocco/shared";
+
+	import type { FormOptions, SuperValidated } from "sveltekit-superforms";
+	import type { CustomerSearchSchema } from "$lib/forms/schemas";
+
+	export let icon = QrCode;
```
silviot, line 10: I don't think the qr code icon is a good fit here.

apps/web-client/src/routes/orders/customers/+page.svelte
```diff
@@ -78,6 +98,24 @@
...
+			<CustomerSearchForm
+				bind:input={searchField}
+				placeholder="Search for customers by name"
+				data={defaults(zod(customerSearchSchema))}
+				options={{
+					SPA: true,
+					dataType: "json",
+					validators: zod(customerSearchSchema),
+					validationMethod: "submit-only",
+					resetForm: true,
```
silviot, line 111: I don't think `resetForm` is a good fit here.

silviot (changes requested):
This is way better, but still has a couple of problems:

* the search field looks like an ISBN input field: it has the qrcode icon at the left
* and behaves like one: it's immediately cleared after a search
* [EDIT] search by user id is not currently possible - it should

As a user, I don't know what search I'm looking at. It's also unclear how to go back to all users (clear the search).

apps/web-client/src/routes/orders/customers/+page.svelte
```diff
@@ -9,26 +9,48 @@
...
+		maxResults += 10;
+	};
+	const scroll = createIntersectionObserver(seeMore);
+	let maxResults = 10;
 
+	$: filteredOrders = customerOrders
+		.filter(({ completed }) => completed === (orderFilterStatus === "completed"))
+		.filter(({ fullname, displayId }) => {
+			if (!$search) return true;
+			return fullname.toLowerCase().includes($search.toLowerCase()) || displayId.toLowerCase().includes($search.toLowerCase());
```
silviot, line 50: One more thing I just noticed: here if `$search` has value ` doe, Jane` I'd like it to also match `Jane Doe`.

I suggest:
* putting code for this (we need to find a name for the function that checks is a given user should be included or not in the search results - `functionThatChecksIfAGivenUserShouldBeIncludedOrNotInTheSearchResults` has its merits, but we can probably find a better one) in a separate `.ts` file, and import it from here.
* add unit tests for common cases

[This conversation might help](https://kagi.com/assistant/eb466a68-ac3a-4c65-9826-da78c4b0fed2)

apps/web-client/src/routes/orders/customers/+page.svelte
```diff
@@ -78,6 +98,23 @@
 
 <Page title="Customer Orders" view="orders/customers" {db} {plugins}>
 	<div slot="main" class="flex flex-col gap-y-6 overflow-x-auto py-2">
+		<div class="p-4">
+			<CustomerSearchForm
+				bind:input={searchField}
+				placeholder="Search for customers by name"
```
delasous, line 104: Please add this to the dictionary and reference via `t`

apps/web-client/src/routes/orders/customers/+page.svelte
delasous: Opening a new ticket for this

apps/web-client/src/routes/orders/customers/+page.svelte
delasous: Will tackle in #988 after this goes in

delasous (approved):

PR #978, opened 2025-06-15
Opfs/test concurrency

ikusteu:
This is a playground created to test out concurrent access of `OPFSAnyContextVFS`. We weren't sure if concurrent access would be possible.

## The setup:
- create a DB wrapper around `wa-sqlite` exposing some minimal implementation akin to `crsqlite-wasm` one
- support `OPFSAnyContextVFS` (preferred solution)
- support `OPFSCoopSyncVFS` (in case the former doesn't allow for concurrent access)
- create a worker (housing the DB) and a wrapper around the worker exposing the DB interface similar to direct interaction (`wkr_db_interface.exec` - akin to `db.exec`)

### Repo setup

Clone and set up the repo, and run
```sh
# get a clean slate
rm -rf ./**/node_modules 
# set up submodules
git submodule update --init --recursive
# install the packages (this will, ideally, make the wa-sqlite objects)
rush update --purge
```

In case the `postinstall` (of `rush update`) fails, and it probably will, inspect [https://github.com/librocco/librocco/blob/opfs/test-concurrency/apps/web-client/scripts/setup_wasm.sh] and take it from there

### Interactions:

Go to `/preview/debug/opfs` and use the console:
```js
// Initialise the local DB
local = await initDB("foo", "opfs-any-context")
// Create a table and insert some data
await local.exec("CREATE TABLE customer")
// IMPORTANT NOTE: the VALUES (...) have to be set explicitly, we don't support binding parameters (?)
await local.exec("INSERT INTO customer (id, name) VALUES (1, 'John')")
```
Open another tab and do
```js
// Initialise the local DB
local = await initDB("foo", "opfs-any-context")
// Read the data 
await local.execO("SELECT * FROM customer")
// should log out [{ id: 1, name: "John" }]
```
Play around with adding/reading from across tabs -- should work

To test further with the worker, from any of the two tabs run:

```js
// Set up the worker DB
// 
// This instructs the worker to open the DB (dbid = "dev")
wdb = await db_worker.open("dev", "opfs-any-context")

// Exec something within the worker
await wdb.exec("SELECT * FROM customer") 
// Should behave in the same way as the main process one
```

## Conclusion

The `OPFSAnyContextVFS` allows for concurrent access (at least that's what I inferred from playing with this), so I didn't test the `OPFSCoopSyncVFS` (as it's more complicated - having to run the DB from the worker -- no direct access -- that much I did test out awhile back)

## Additional observations

There seems to be some discrepancy between `crsqlite-wasm` wrapper and the way I've interacted with `wa-sqlite` -- some functions just didn't work (or had different signature), so additional effort should be applied in order to make the crsqlite wrapper(s) work with the up-to-date `wa-sqlite`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,6 @@ common/autoinstallers/*/.npmrc
 .aider*
 
 caddy-data
+
+# Ignore the symlink for the WASM binary (it should be created by the build process)
+*/**/wa-sqlite-async.wasm
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "3rd-party/wa-sqlite"]
+	path = 3rd-party/wa-sqlite
+	url = git@github.com:librocco/wa-sqlite
--- a/apps/web-client/package.json
+++ b/apps/web-client/package.json
@@ -13,7 +13,7 @@
 		"build:prod-rootdir": "VITE_GIT_SHA=$(git rev-parse --short HEAD) BASE_PATH= vite build",
 		"build:e2e": "VITE_GIT_SHA=$(git rev-parse --short HEAD) PUBLIC_IS_E2E=true vite build",
 		"format": "prettier --write .",
-		"postinstall": "svelte-kit sync",
+		"postinstall": "svelte-kit sync && ./scripts/setup_wasm.sh",
 		"lint": "prettier --check . && eslint .",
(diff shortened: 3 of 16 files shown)


silviot:
The test was succesful. We now know we'll be able to use the OPFS version of wa-sqlite with the cr-sqlite extension, as soon as we have it.

PR #982, opened 2025-06-17
Setup/wa sqlite environment

### LINKED ISSUE #981: Use wa-sqlite from a repository instead of the one from the npm registry
For now we'll point it to `master` of https://github.com/vlcn-io/wa-sqlite

ikusteu:
Fixes #981 

It doesn't really point to `master` of https://github.com/vlcn-io/wa-sqlite, but rather to `vlcn-io/latest-release` https://github.com/librocco/wa-sqlite -- a branch in our repo, created to keep the commit of latest release `@vlcn.io/wa-sqlite` -- the same one used in https://github.com/vlcn-io/js repo (to build out `crsqlite-wasm` we're using in our app).

Additionally, as we need to build out the entire stack, this features a fork of https://github.com/vlcn-io/cr-sql (`vlcn-io/latest-release` of https://github.com/librocco/cr-sqlite to be precise -- same reason as above)

In contains a copy/paste of `crsqlite-wasm` package, as we need to point it to our "local" submodules, rather than their official counterparts + it adds some build logic to achieve the similar build process as the one in https://github.com/vlcn-io/js

Finally, it includes a fork of https://github.com/emscripten-core/emsdk -- used to manage `emcc` version (as `@vlcn.io/wa-sqlite` uses an old version of `emcc`)

### DIFF
--- a/3rd-party/crsqlite-wasm/scripts/build-wasm.sh
+++ b/3rd-party/crsqlite-wasm/scripts/build-wasm.sh
@@ -2,6 +2,12 @@
 
 mkdir -p dist
 
+# This env variable is required by the crsql build process
+# (as it's, and I quote, "only meant to be built as a submodule of @vlcn.io/js")
+#
+# We're providing a dummy value here (this has nothing to do with the underlaying build, but here we are)
+export CRSQLITE_COMMIT_SHA="$(git rev-parse HEAD)"
+
 cd ../emsdk
 ./emsdk install 3.1.45
 ./emsdk activate 3.1.45


3rd-party/crsqlite-wasm/scripts/build-wasm.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# This env variable is required by the crsql build process
+# (as it's, and I quote, "only meant to be built as a submodule of @vlcn.io/js")
+#
+# We're providing a dummy value here (this has nothing to do with the underlaying build, but here we are)
+export CRSQLITE_COMMIT_SHA="$(git rev-parse HEAD)"
+
+cd ../emsdk
+./emsdk install 3.1.45
+./emsdk activate 3.1.45
```
silviot, line 13: Why do we need this specific `emcc` version?

3rd-party/crsqlite-wasm/scripts/build-wasm.sh
```diff
@@ -0,0 +1,20 @@
...
+source ./emsdk_env.sh
+
+cd ../wa-sqlite
+make
+cp dist/crsqlite.wasm ../crsqlite-wasm/dist/crsqlite.wasm
+cp dist/crsqlite.mjs ../crsqlite-wasm/src/crsqlite.mjs
```
silviot, line 19: Why not like this?
```suggestion
cp dist/crsqlite.mjs src/crsqlite.mjs
```

3rd-party/crsqlite-wasm/scripts/build-wasm.sh
ikusteu: No idea from global point of view.

From in practice point of view:
- I can't remember the exact issue (the built binary wasn't compatible with the JS glue -- it would break at runtime)
- it was fixed by (selectively) copying the [script](https://github.com/vlcn-io/js/blob/main/build-wasm.sh) used in `@vlcn.io/js` package (called from the [root Makefile](https://github.com/vlcn-io/js/blob/main/Makefile))

3rd-party/crsqlite-wasm/scripts/build-wasm.sh
```diff
@@ -0,0 +1,20 @@
...
+export CRSQLITE_COMMIT_SHA="$(git rev-parse HEAD)"
+
+cd ../emsdk
+./emsdk install 3.1.45
+./emsdk activate 3.1.45
+source ./emsdk_env.sh
+
+cd ../wa-sqlite
+make
+cp dist/crsqlite.wasm ../crsqlite-wasm/dist/crsqlite.wasm
```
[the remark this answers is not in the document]
ikusteu: The pseudo-script, including the working directory for every step

```sh
#NOTE: This is run from root of crsqlite-wasm package (as it's called via `package.json`-defined script)
librocco/3rd-party/crsqlite-wasm: mkdir -p dist

# This env variable is required by the crsql build process
# (as it's, and I quote, "only meant to be built as a submodule of @vlcn.io/js")
#
# We're providing a dummy value here (this has nothing to do with the underlaying build, but here we are)
librocco/3rd-party/crsqlite-wasm: export CRSQLITE_COMMIT_SHA="$(git rev-parse HEAD)"

librocco/3rd-party/crsqlite-wasm: cd ../emsdk
librocco/3rd-party/emsdk: ./emsdk install 3.1.45
librocco/3rd-party/emsdk: ./emsdk activate 3.1.45
librocco/3rd-party/emsdk: source ./emsdk_env.sh

librocco/3rd-party/emsdk: cd ../wa-sqlite
librocco/3rd-party/wa-sqlite: make
librocco/3rd-party/wa-sqlite: cp dist/crsqlite.wasm ../crsqlite-wasm/dist/crsqlite.wasm
librocco/3rd-party/wa-sqlite: cp dist/crsqlite.mjs ../crsqlite-wasm/src/crsqlite.mjs
```

ikusteu:
> I think we should try not to vendor in anything, but keep things in their respective repositories. I'm not sure what problem exactly granted the vendoring in of `crsqlite-wasm`.
> 
> I [opened a PR](https://github.com/librocco/librocco/pull/989) that solves some of the issues that prevent this one from being green.

`crsqlite-wasm` is not a standalone, but rather the whole [https://github.com/vlcn-io/js] monorepo. I wanted to reduce the number of moving parts as I remember having to jump through additional hoops in order to set up the monorepo (in hindsight, that was awhile back and the "hoops" may have been my lack of knowledge of make process -- which I've definitely upgraded since that). However, this might be worth revisiting, as the way the monorepo is structured it contains `wa-sqlite` and `emsdk` as submodules, so it would reduce the complexity in our setup.

Like I said, I went with this approach to keep it as minimal as possible and discarded the idea of using the `vlcn.io/js` monorepo due to some (potentially no-longer-relevant) past sentiment.

3rd-party/crsqlite-wasm/scripts/build-wasm.sh
silviot: Same mistake as above. I missed the `cd ../wa-sqlite`. Sorry for the noise.

PR #987, opened 2025-06-18
Chore: improve lucide-imports to reduce build time

delasous:
I noticed [this issue](https://github.com/lucide-icons/lucide/issues/1475) with lucide imports awhile ago. This adds an import alias for lucide and updates imports in all components/controllers/pages. We should now use this pattern to import icons `import IconName from $lucide/icon-name`. This speeds up our production build. It's a small update but easy to do with an LLM


ikusteu:
Build times aren't that horrid for me, so I'll take your word for it. I don't see a reason to object this PR.

PR #996, opened 2025-06-23
Create normalization function for search queries

### LINKED ISSUE #994: Improve user search: "Jane Doe" and "Doe, Jane" should both match "Jane Marie Doe".
From [this comment](https://github.com/librocco/librocco/pull/976#pullrequestreview-2938562364):

One more thing I just noticed: here if `$search` has value ` doe, Jane` I'd like it to also match `Jane Doe`.

I suggest:
* putting code for this (we need to find a name for the function that checks is a given user should be included or not in the search results - `functionThatChecksIfAGivenUserShouldBeIncludedOrNotInTheSearchResults` has its merits, but we can probably find a better one) in a separate `.ts` file, and import it from here.
* add unit tests for common cases

[This conversation might help](https://kagi.com/assistant/eb466a68-ac3a-4c65-9826-da78c4b0fed2)

fadwamahmoud:
Fixes #994


silviot (changes requested):
The approach of normalizing both the needle and the haystack and then check if the haystack contains the needle is a bit obscure in my opinion, and open for subtle bugs.

I had [a new conversation with LLMs](https://kagi.com/assistant/ccc85058-aa22-4f6e-960a-f6cac13a08da) about it where I state my concerns.

One big concern with current code: I believe that, if we have a cusomer with name `Anne Josephine Beatrice Julie Jane Mary Doe` (I know, I know! bear with me) then these strings won't match:
* Josephine Doe
* Beatrice Mary
* Anne Doe

since, once normalized, they're not contained in the normalized version of the customer name.
I did not test this, so I might be wrong.

Note that name would be `Anne Beatrice Doe Jane Josephine Julie Mary` once normalized.
It's not straightforward to test (by editing a test file and adding my case), and this in itself is another problem.
Ideally the tests should show what needle matches what haystack.

In conclusion, this PR needs a couple of changes:
* `Mary Doe` should match `Anne Josephine Beatrice Julie Jane Mary Doe`
* tests should explicitly state a cusotmer name and a search string, and assert whether it's a match or not
* I suggest the approach outlined in the last message of the conversation with the LLM

apps/web-client/src/lib/utils/misc.ts
```diff
@@ -40,3 +40,34 @@ export function normalizeName(name: string) {
...
+			.trim()
+			.split(" ")
+			.filter((part) => part.length > 0); // Remove any empty parts resulting from normalization
+	};
+
+	const needleConditions = getConditions(needle);
+
+	// If after normalization the needle has no conditions (e.g., needle was empty or just punctuation)
+	if (needleConditions.length === 0) {
+		return needle.trim().length === 0;
```
silviot, line 60: What is the reasoning behind this? Wy not returning `true` when `needleConditions` has length 0?

As far as I can tell, it makes these two queries yield different results: `   ` (only whitespace) and `  . ` (possibly whitespace and punctuation).

apps/web-client/src/lib/utils/misc.ts
```diff
@@ -40,3 +40,34 @@ export function normalizeName(name: string) {
...
+	}
+
+	// Normalize the haystack for checking (lowercase and standardize punctuation/spacing)
+	// This ensures "O'Malley" in haystack can match "OMalley" or "O Malley" from needle.
+	const normalizedHaystack = haystack
+		.toLowerCase()
+		.replace(/[^\w\s]/g, " ") // Replace punctuation with a space
+		.replace(/\s+/g, " ") // Collapse multiple spaces
+		.trim();
```
silviot, line 69: I'm under the impression only `.toLowerCase()` is doing something useful here. What is the purpose of replacing punctuation, collapsing spaces and trimming?

apps/web-client/src/lib/utils/misc.ts
```diff
@@ -30,3 +30,44 @@ export const extractBookData = <T extends BookData>(entry: T): BookData => ({
...
+	// Helper to normalize and tokenize the needle string into conditions
+	const getConditions = (str: string) => {
+		return str
+			.toLowerCase()
+			.replace(/[^\w\s]/g, " ") // Replace punctuation with a space
+			.replace(/\s+/g, " ") // Collapse multiple spaces into one
+			.trim()
+			.split(" ")
+			.filter((part) => part.length > 0); // Remove any empty parts resulting from normalization
+	};
```
silviot, line 54: There's no reason for this function to be a closure. It can be moved outside of `matchesName` for increased readability and testability (even if we don't have a test right now, we might always add one in the future if we find a bug with this function).

Issue #1006, opened 2025-06-25
Separate Warehouse Selection in case of no stock


fadwamahmoud:
Warehouse Dropdown list:
- Button to "Force Remove" from a warehouse where there is no stock
    - Opens up a dialog (the second dialog that would open on clicking commit)
- Remove Warehouses where there is no stock from list
- ~~Colour code warehouse dropdown list instead of entire row~~

delasous:
I'd also suggest:
- The whole row still stays coloured, we just go for a lighter red. This is to be used when a warehouse selection needs to be made (either because there are multiple, or because it doesn't exist in any).
- Remove the use of yellow to style the first case (in stock but multiple warehoues)
- Remove the use of red to style rows that will create a reconciliation note
- Add the count of available copies next to the warehouse name, e.g "Warhouse (2) 2 copies"
    - "2 copies" should probably be positioned below with lighter, smaller text
    - & pluralised properly using our i18n tooling
- We should play with the label for "Force Remove" - "Force warehouse"? "Override warehouse/stock"? "Force withdrawal"? and add some copy which explains how it works which is accessible via a little "(?)" popover (Edible has these in the MenuItemForm). This could draw on the existing messaging in the renconciliation confirmation dialog to provide examples of what happens
    - To explain "Remove Warehoueses where there is not stock from list" to anyone reading these proposals for the first time: the idea is that we don't give the user a list of all available warehouses by default, and confirm that they understand the consequences of "taking it from a warehouse that doesn't have stock" later. Instead, we assume that "they know better" and want to override => we force them to explicitly open the dialog and choose the warehouse override, and remove all of the additional warnings that we've had to provide after (red row, additional confirmation step)

fadwamahmoud:
1 Stock Management                                                 
    • On the outbound note page, fetch and store the initial stock levels for   
      all warehouses.                                                           
    • Utilize a StockMap? to serve as an       
      in-memory representation of the items being added to the current note.    
      This map will track the quantities of each ISBN assigned to each warehouse
      within the note.                                                          
 2 Implement Core Stock Validation Logic                                        
    • Create a central function, e.g., assignItemToWarehouse(isbn, quantity,    
      warehouseId?), to be invoked when scanning a book or updating an item's   
      quantity or ==warehouse==.  (re-validate the stock when wh is updated => add the quantity to previous wh in StockMap and subtract quantity from new wh in stockMap and create new row if needed)                                                            
[...]

PR #1014, opened 2025-06-26
Only display available warehouses in warehouseSelect

fadwamahmoud:
#1006 

What's done so far:
 - only show  available warehouses in warehouseSelect
 - show a dropdown of unavailable warehouses when "Force withdrawal" is clicked on each row
 - show a summary of what happens when user clicks confirm in withdrawal dialog
 - persist book_transaction row when confirmed
 - on quantity update => separate available and unavailable quantities into separate rows
 - when a default wh is selected, if scanned book does not exist in that (or any) wh don't assign the default wh to the scanned book row (current behaviour: if book doesn't exist in any warehouse, it *is* assigned to default wh, if book exists in different other wh but not default it's not assigned to default)
 - when assigning a different warehouse, check if the already scanned quantity exists in next warehouse, if not create separate row 
 - e2e tests
- Show "Force Withdrawal" button in the WarehouseSelect menu, not in the table cell
- Use PageCenterDialog for force withdrawal
- Dictionaries

 
TBD:
- Assert for quantities in e2e tests => not sure why not working

## Summary by CodeRabbit

* **New Features**
  * Added support for forced withdrawals when stock is insufficient, including a dedicated dialog for selecting warehouses.
  * Outbound note tables now display a "Type" column indicating transaction type ("normal" or "forced").
  * Warehouse selection UI now shows available stock per warehouse and allows forced withdrawal actions.
  * Enhanced alert messages and UI feedback for insufficient stock scenarios.

* **Improvements**
  * Warehouse selection and transaction assignment logic is now stock-aware, auto-splitting rows when quantities exceed available stock.
  * Outbound note and warehouse dropdowns update dynamically after warehouse deletion.
  * Enhanced localisation with new strings for force withdrawal actions and alerts.

* **Bug Fixes**
  * Fixed row styling and UI behaviour for out-of-stock and forced withdrawal scenarios.

* **Tests**
  * Added comprehensive end-to-end tests for outbound note flows, warehouse stock handling, and forced withdrawal cases.


apps/web-client/src/routes/outbound/[id]/+page.svelte
```diff
@@ -654,6 +700,40 @@
 	</div>
 </Page>
 
+{#if $forceWithdrawalDialogOpen}
+	{#if forceWithdrawalDialogData}
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.svelte:773]: These could be combined?

apps/web-client/src/routes/outbound/[id]/+page.svelte
```diff
@@ -654,6 +700,40 @@
...
+		{@const { row, unavailableWarehouses } = forceWithdrawalDialogData}
+		<div use:melt={$forceWithdrawalDialogPortalled}>
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.svelte:775]: There is a PageCenterDialog in `lib/components/Melt` that does this

apps/web-client/src/routes/outbound/[id]/+page.ts
```diff
@@ -54,6 +54,17 @@ const _load = async ({ parent, params, depends }: Parameters<PageLoad>[0]) => {
...
+  "978-3-16-148410-0" => Map {
+    1 => { displayName: "Main Warehouse", quantity: 10 },
+    2 => { displayName: "Downtown Branch", quantity: 5 }
+  },
+  "978-1-40-289462-6" => Map {
+    1 => { displayName: "Main Warehouse", quantity: 8 }
+  }
+}
+	 */
 	const isbnAvailability = new Map(isbns.map((isbn) => [isbn, new Map<number, { displayName: string; quantity: number }>()]));
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.ts:78]: Can you justify the two Maps used here? Why not an object of `{ [isbn]: Array<{ displayName: string, quantity: number }> }`

delasous (changes requested):
- I would say that we want the "Force Withdrawal" button to be shown in the `WarehouseSelect` menu, not in the table cell. To do this, I would add a slot at the foot of that menu. I would also use a slot on the `OutboundTable` for the warehouse cell, so that the WarehouseSelect and its button slot can all be composed on the page itself. 

- The way you have setup the Melt dialog is fine. I would then just add the trigger to button that's passed to the WarehouseSelect slot.

- Please add new strings to dictionaries

delasous (commented):
It's taking shape. I haven't had time to look at it thoroughly though - just click around, but I've noticed a couple things:
1. After selecting the warehouse via "force withdrawal", I can reopen the dialog to change my selection. But if I don't, "Confirm" should be disabled.
2. Its confusing reading the copy count as e.g ": 1", I'd minimally move it into an outlined badge after the name. The default value for the select is just the Warehouse Name, but if I select an option it can become Warehouse Name: 1 - we should make this consistent
3. After selecting the force withdrawal warehouse the row is still red. I know this is detailed in another ticket, but its confusing - I just acted, why is it still red? I feel like this functionality should be tackled here too...
4. The temp "no stock" behaviour you've described isn't quite working: I don't seem to get the alert on the first overstock selection. To replicate:
    1. I add a book "123" which is in "New Warehouse" 
    2. I add a second book "123", it creates a new line. I select "New Warehouse", it merges the rows and makes it red to show it will create a reconciliation note.
    4. It's only on the next time I try to increment the row quantity, that I see an alert
    5. Now that I think about it, I shouldn't be able to just select "New Warehouse" here, I should have to use force withdrawal? But if I open the dialog there are no options to select there... - or is this what #1029 is intended to introduce?
    6. Additionally, if I try to add 3 new copies via the new line, and select "New Warehouse", it adds one copy to the first row, and keeps two in the existing row. Both of the rows are now red, and I am very confused.

I think we should pause for a second to detail these cases and how we will handle them / what we expect to see at the end of each of them, because without a clear picture on how we are meant to handle the behaviour related to point 4, I think this is in danger of making the UX worse.

delasous:
I am seeing some new behaviour which seems buggy:
- I add a book "123" which I know is in a warehouse. The row is red, it says "Select Warehouse", the type column is already "forced". I select New Warehouse from the list, it splits the row... the new row is not red, it has a "normal" type AND the red row is still there. I can keep repeating this action on it, duplicating row after row...
    - If I do this with a quantity of 2, two rows are split
- This same behaviour happens in a "No stock" scenario - I add book "8812912" (which is in no warehouse), I have to open force withdrawal (nice), but after forcing withdrawal the row is duplicated...
- Additionally 
    -  I had two warehouses each with 1 copy of "123". In the Outbound note: if I add "123" I have to select between the two in the WarehouseSelect. I now want to see the behaviour of having it available in only 1 warehouse, so I delete 1 warehouse. I go back to the Outbound note. The row doesn't have a warehouse selected by default as I would expect as it should only be available in 1. Instead it has two options "New Warehouse: 1" and "Null: 1"...
    - an odd case I know, but I think it highlights something wrong with the logic in the page load?

fadwamahmoud:
@coderabbitai review

apps/web-client/src/routes/outbound/[id]/+page.svelte
```diff
@@ -633,6 +744,22 @@
...
+								<WarehouseSelect {row} {rowIx} warehouseList={warehouses} on:change={(event) => editWarehouse(event, row)}>
+									<button
+										let:open
+										use:melt={$forceWithdrawalDialogTrigger}
+										slot="force-withdrawal"
+										data-testid={testId("force-withdrawal-button")}
+										on:m-click={() => {
+											openForceWithdrawal(row);
+											open.set(false);
+										}}>{tOutbound.labels.force_withdrawal() || "Force Withdrawal"}</button
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.svelte:758]: Why do we need to fallback to "Force Withdrawal"?

PR #1030, opened 2025-07-03
Fix/local package resolution

ikusteu:
Fixes an issue when built tarball (one of the vlcn packages) references another (excluded from top level package.json):
- extend the `build_vlcn.sh` script to also build `wa-sqlite` (required by `crsqlite-wasm`)
- use Rush package resolution overrides:
  - create `common/config/rush/pnpm-config.json` (new way of handling pnpm config for Rush), [see here](https://rushjs.io/pages/configs/pnpm-config_json/)
  - remove `pnpmOptions` from `rush.json` ☝️ both can't exist at the same time
  - alias (vlcn) packages (built tarballs) within `pnpm-config.json` (and reinstate versions in web-client's `package.json` -- gets overridden anyway, but cleaner this way)

Create aliases for dev mode:
- (locally) I'd work with a commit aliasing vlcn imports (in `web-client`) to submodules (not built tarballs), for quicker dev feedback loop
- made this work both for dev and local tests (aliased), as well as production (including `build:e2e`) and CI tests (`test:ci`), so that within the former env, the paths are aliased, whereas in latter env the built tarballs are used
- this way we get a full check, with the agility of local dev (and without me having to add/remove the commit taking care of aliases each time)


apps/web-client/src/i18n/de/index.ts
```diff
@@ -0,0 +1,8 @@
+import type { Translation } from '../i18n-types.js'
```
silviot, line 1: Was this file committed by mistake?

silviot: I pushed a fix for this

silviot:
One more thing I just noticed: these lines in `common/config/rush/pnpm-lock.yaml` have your specific paths in them:

```
        specifier: file:/Users/ivan/dev/JS/librocco/3rd-party/artefacts/vlcn.io-crsqlite-wasm-0.16.0.tgz
        version: file:../../3rd-party/artefacts/vlcn.io-crsqlite-wasm-0.16.0.tgz
```

...which means that as soon as I run `rush update` my local git copy gets dirty.

<details>

<summary>I asked Gemini to comment on this</summary>
For your project, the core issue is that `pnpm` intentionally converts `file:` paths for your tarball dependencies into absolute paths within the `pnpm-lock.yaml` file. This is expected behaviour.

Here is the best advice to solve this and ensure your lockfile is consistent across different machines:

### The Recommended Solution: Use a `pnpm` Hook

Your best option is to create a `.pnpmfile.cjs` hook. This file will programmatically replace the absolute paths with a consistent relative path every time you run `rush update`. This is the favourite method as it directly fixes the issue without changing your project's structure.

1.  **Create the file:**
    `common/config/rush/.pnpmfile.cjs`

2.  **Add this code:** This script intercepts the lockfile creation and replaces the problematic path.

    ```javascript
    // common/config/rush/.pnpmfile.cjs
    'use strict';

    module.exports = {
      hooks: {
        afterAllResolved: (lockfile, context) => {
          const tarballFileName = 'vlcn.io-crsqlite-wasm-0.16.0.tgz';

          for (const importerKey in lockfile.importers) {
            const importer = lockfile.importers[importerKey];
            if (!importer.specifiers) continue;

            for (const depName in importer.specifiers) {
              const specifier = importer.specifiers[depName];

              if (specifier.startsWith('file:') && specifier.endsWith(tarballFileName)) {
                // Use a relative path from the lockfile's location.
                const relativePath = 'file:../../../3rd-party/artefacts/' + tarballFileName;
                context.log(`Replacing absolute path for ${depName} with: ${relativePath}`);

                importer.specifiers[depName] = relativePath;

                if (importer.dependencies && importer.dependencies[depName]) {
                    importer.dependencies[depName] = relativePath.replace('file:', 'link:');
                }
              }
            }
          }
          return lockfile;
        }
      }
    };
    ```

3.  **Update the lockfile:** Run `rush update --full` to apply the changes.

### Relative Path Base

Relative paths within `common/config/rush/pnpm-lock.yaml` are resolved from the **directory containing the lockfile itself**: `common/config/rush/`.
</details>

Issue #1029, opened 2025-07-03
Separate Warehouse Selection in case of no stock pt.2


fadwamahmoud:
#1006 got a bit too big so,
- Split row on assigning a warehouse with partially availability, currently the user gets an alert that the warehouse they've selected doesn't contain enough stock
- Show confirmation dialog before row-splitting
- update wh available quantities on scanning (in dropdown)

PR #1047, opened 2025-07-15
Only display available warehouses in warehouseSelect

silviot:
Resurrect #1014 

#1006 #1009 #1029

What's done so far:
 - only show  available warehouses in warehouseSelect
 - show a dropdown of unavailable warehouses when "Force withdrawal" is clicked on each row
 - show a summary of what happens when user clicks confirm in withdrawal dialog
 - persist book_transaction row when confirmed
 - on quantity update => separate available and unavailable quantities into separate rows
 - when a default wh is selected, if scanned book does not exist in that (or any) wh don't assign the default wh to the scanned book row (current behaviour: if book doesn't exist in any warehouse, it *is* assigned to default wh, if book exists in different other wh but not default it's not assigned to default)
 - when assigning a different warehouse, check if the already scanned quantity exists in next warehouse, if not create separate row 
 - e2e tests
- Show "Force Withdrawal" button in the WarehouseSelect menu, not in the table cell
- Use PageCenterDialog for force withdrawal
- Dictionaries

 
TBD:
- Assert for quantities in e2e tests => not sure why not working

## Summary by CodeRabbit

* **New Features**
  * Added support for forced withdrawals when stock is insufficient, including a dedicated dialog for selecting warehouses.
  * Outbound note tables now display a "Type" column indicating transaction type ("normal" or "forced").
  * Warehouse selection UI now shows available stock per warehouse and allows forced withdrawal actions.
  * Enhanced alert messages and UI feedback for insufficient stock scenarios.

* **Improvements**
  * Warehouse selection and transaction assignment logic is now stock-aware, auto-splitting rows when quantities exceed available stock.
  * Outbound note and warehouse dropdowns update dynamically after warehouse deletion.
  * Enhanced localisation with new strings for force withdrawal actions and alerts.

* **Bug Fixes**
  * Fixed row styling and UI behaviour for out-of-stock and forced withdrawal scenarios.

* **Tests**
  * Added comprehensive end-to-end tests for outbound note flows, warehouse stock handling, and forced withdrawal cases.


delasous:
@fadwa I left these comments on the other PR and it doesn't look like they have been resolved:

1. I'm still seeing some buggy behaviour with "Overstock" scenario's. If I know I have 1 copy of  "123" in "New Warehouse" and I: 
    1. Add another copy by scanning "123" - it splits the row out correctly but I can select it "New Warehouse" from the list of options and the options in Force Withdrawal are empty 
    2. Add another copy by editing the Quantity col to 2 - it splits the row but it immediately selects "New Warehouse" from the list of options. Again, this option is in the WarehouseSelect, not the Force Withdrawal Dialog which has no options.
2. Regarding the Force Withdrawal dialog:
    1. If I open it for the row "123" which has "New Warehouse" selected, I see "New Warehouse" in the force select option. I think this select should read the placeholder message in this case I've described in another comment, where I have selected a forced warehouse and I reopen the dialog but there is nothing to change.
    2. The commit button should be disabled here, and i n
3. Our column space is precious, I think the "Type" col should be moved into the WarehouseSelect.
4. The "no. of books" icon isn't very clear. I think it would be best to display it as copy "X book(s)"

Issue #992, opened 2025-06-20
Simplify the 'booked with supplier' functionality


ikusteu:
Continues after #936 @silviot can you provide the ticked description (suggested simplifying of the implemented functionality)

Issue #991, opened 2025-06-20
Test 'booked with supplier' functionality


ikusteu:
Cont'd from #936 

Test the implemented functionality

delasous:
The Reconciliation UI/UX is being rebuilt. The "booked with supplier" functionality will be included in this effort. #1146 will track the new tasks when the design has settled

PR #1049, opened 2025-07-15
Feat/booked with supplier

### LINKED ISSUE #936: In order reconciliation, allow users to mark books as "booked at the supplier"
Some suppliers keep tabs of ordered books that were not delivered. In this case the non delivered books should not be ordered again.

We should support this scenario.

Proposal:

on the very last screen of reconciliation we show books that were not delivered. We should allow users to check the ones that don't need reordering. If the user selects some books a new supplier order will be created, so that they won't be ordered again.

silviot:
Resurrect #993 

Fixes #936 

- creates a column on reconciliation comparison table - checkbox for the functionality
- when finalising a reconciliation order, it checks for underdelivered lines marked as 'booked with supplier' and opens "continuation orders" (supplier pseudo-orders)

This needs to be tested + Silvio had mentioned some potential tweaks / simplifications, see:
- #991
- #992

**NOTE:** e2e failing. In spirit of cutting off, I'm not doing this right now. I'll take another stab later, if I have the time, after I'm done with `crsqlite-wasm` (and friends) built from source work.

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/order-reconciliation.ts
@@ -42,7 +42,7 @@ import type {
 import { timed } from "$lib/utils/timer";
 
 import { multiplyString } from "./customers";
-import { createSupplierOrder } from "./suppliers";
+import { createSupplierOrderTxSafe } from "./suppliers";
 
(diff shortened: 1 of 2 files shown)


silviot:
One new thing: books that are booked with the supplier should be marked with a new `booked` state, almost equivalent to `ordered`, but useful to the bookshop people.

PR #1052, opened 2025-07-21
Improve Default Warehouse Selection

fadwamahmoud:
Improve isbn scanning ux by aligning the default warehouse select element with the isbn scan element

### DIFF
--- a/apps/web-client/src/routes/outbound/[id]/+page.svelte
+++ b/apps/web-client/src/routes/outbound/[id]/+page.svelte
@@ -591,7 +591,7 @@
 			</div>
 
 			<div class="flex h-full flex-wrap items-center gap-2">
-				<div class="flex flex-grow items-center gap-2">
+				<div class="flex grow items-center max-sm:w-full">
 					<ScannerForm
 						data={defaults(zod(scannerSchema))}
 						options={{
@@ -627,7 +627,7 @@
 					use:melt={$customItemDialogTrigger}
 					on:m-click={() => openCustomItemForm()}
 					on:m-keydown={() => openCustomItemForm()}
-					class="btn-neutral btn max-sm:w-full sm:mx-2"
+					class="btn-neutral btn max-sm:btn-sm max-sm:w-full"
 				>
 					<Plus />
 					{tOutbound.labels.custom_item()}</button


apps/web-client/src/routes/outbound/[id]/+page.svelte
```diff
@@ -619,7 +603,28 @@
 							await handleAddTransaction(isbn, 1, warehouseId);
 						}
 					}}
+					placeholder={"Scan to select books from"}
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.svelte:606]: Add to dicts, and add "..." at the end

apps/web-client/src/routes/outbound/[id]/+page.svelte
```diff
@@ -619,7 +603,28 @@
...
 				/>
+				<select
+					id="defaultWarehouse"
+					name="defaultWarehouse"
+					class="select-bordered select select-sm h-full"
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.svelte:611]: When the default option "Any warehouse" is displayed, style it italic

apps/web-client/src/routes/outbound/[id]/+page.svelte
```diff
@@ -619,7 +603,28 @@
...
+					{#each warehouses as warehouse}
+						<option value={warehouse.id}>{warehouse.displayName}</option>
+					{/each}
+				</select>
+				<!-- <div id="button-container" class="flex h-24 w-full items-center justify-start px-8"> -->
+				<button
+					use:melt={$customItemDialogTrigger}
+					on:m-click={() => openCustomItemForm()}
+					on:m-keydown={() => openCustomItemForm()}
+					class="btn-neutral btn mx-2">{tOutbound.labels.custom_item()}</button
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.svelte:625]: I feel like a + icon before the text would be helpful here to give it more context

delasous (commented):
Looks good. Two other things besides the small comments:
1. Remove the "Custom item" button from below the empty state placeholder
2. On small screen sizes, wrap the "Custom item" button next to the input-select to the line below and make it full width + `btn-sm`

delasous (approved):
My comment about the button size might have been unclear, sorry: 
- keep the select the same size as the input on large and small screen sizes
- make the custom item button small only at small screen sizes

PR #1053, opened 2025-07-21
Enhance/local package resolution

ikusteu:
Updated (history conformant version of #1030)

This also includes some fixes, all in all, this:
- overrides `@vlcn.io` dependencies (wherever in the dep tree they may be) to use the `3rd-party/artefacts` tarballs
- it makes sure the lockfile doesn't include my absolute path to each respective file, but rather a relative one
- it allows for aliasing of submodules as dependencies (in which case the submodules are used, not the tarballs)
- ☝️ this is enabled by (explicitly) passing `USE_SUBMODULES=true` env variable (the default uses the built tarballs, regardless of the environment)

There's one thing I've noticed: whenever we build the tarballs, some are always different (which ones seems really arbitrary), even if the source (submodules) hadn't changed...however, this is:
- an issue for another unit of work
- potentially quite complex
- **does** cause some bloat (in terms of git lfs), but probably won't be used as often as to justify the previous point


silviot:
[Reproducible builds](https://en.wikipedia.org/wiki/Reproducible_builds) are not trivial. I'm not surprised we get different tarballs after building the same source.

apps/web-client/svelte.config.js
```diff
@@ -5,6 +5,42 @@ import { preprocessMeltUI } from "@melt-ui/pp";
 import sequence from "svelte-sequential-preprocessor";
 
 const BASE_PATH = process.env.BASE_PATH ?? "";
+const USE_SUBMODULES = process.env.USE_SUBMODULES === "true";
+
+/**
+ * A helper used to provide aliases for vlcn.io packages in dev mode:
+ * - if in dev mode, we're aliasing vlcn.io imports to submoduled packages (under '3rd-party/js')
+ * - if in production mode, no aliases are provided: installed dependencies built from tarballs ('3rd-party/artefacts') are used
```
silviot, line 13:
```suggestion
 * A helper used to provide aliases for vlcn.io packages in dev mode:
 * - if `USE_SUBMODULES` env var is truthy, we're aliasing vlcn.io imports to submoduled packages (under '3rd-party/js')
 * - if `USE_SUBMODULES` env var is undefined or falsy, no aliases are provided: installed dependencies built from tarballs ('3rd-party/artefacts') are used
```

PR #1055, opened 2025-07-22
Wa sqlite/checkpoint 3 wip

ikusteu:
This is WIP for now. 

In terms of wa-sqlite reconciliation, this, as **checkpoint 3** is ready.

However, it includes the OPFS work in librocco, which produces less than ideal results.

1. Everything works (to an extent):
  - it defines our own `initWasm` which allows for usage of `opfs-any-context-vfs`, by doing `sqlite3.open(name, "c", "opfs-any-context-vfs")` (the "c" bit is the default in existing code)
  - we use the aforementioned open logic to use the opfs in the app and it works (FS transparency and all)
  - it applies the similar logic in the sync-worker (refactoring some of the `ws-browserdb` logic -- used to instantiate the DB in the worker -- to use opfs)
  - the sync works in-between tabs as well as in between chorme <-> mozilla (via sync server)
  - updating the DB on the server doesn't work immediately and one needs to refresh in order for the updates to become visible (this isn't the case with browser <-> browser sync, which is weird since it all goes through the same server, however, this might be some mismatch in my existing DB setup -- this is the _to and extent_ bit)

2. It's excruciatingly slow:
  - well maybe not as slow as `PouchDB`, but not too fast either
  - sync seems to suffer the same problem as `PouchDB`: slower and slower as time goes by
  - refreshing the page (and navigation) is noticeable and unpleasant (not snappy)
  - reactivity could be faster (scanning a book is also noticeable -- not snappy)
  - sync between browsers is janky - sometimes it even takes so long as to flash the progress bar for that 1 single change
  
All in all: we have the OPFS, but it's not ideal and should probably be looked into a bit more


ikusteu:
#1064 contains either this or corrected version of this

ikusteu:
#1064 contains either this or corrected version of this

Issue #1008, opened 2025-06-25
Communicate Reconciliation in Commit message


fadwamahmoud:
we should update the info in the first dialog 

Keep "2 books will be removed from stock"

(Some way to express "Reconcilation Note"), e.g 1 book will be added to a warehouse & removed...

fadwamahmoud:
1. Trying to commit an outbound note with an unselected warehouse should show a popover from commit button
2. stock mismatch dialog content should be moved into first dialog contained in a <details> element that expands 
- no of books actually be removed from stock should be listed first
- summary of the rest (reconciliation note) in the details 
- "reconciliation" is not a user facing term

Issue #1007, opened 2025-06-25
Show "No warehouse selected" warning early


fadwamahmoud:
- on clicking commit, if one of the rows has no warehouse selection, instead of opening a dialog, we show a warning message below/next to commit button
- non-selected row would already be red

delasous:
This should also remove the second dialog that pops up

silviot:
Would it be easy to flash the rows and the button three time when it's pressed but the form doesn't validate?

PR #1062, opened 2025-07-30
Merge confirmation and reconciliation dialogs

fadwamahmoud:
- getOutOfStockEntries function was exported to allow for validation before attempting to commit a note

- outbound note page was refactored to change the commit workflow. Instead of a single action that could throw an error, clicking "Commit" now performs a series of pre-checks:                                   

 1 It first validates if any items are missing a warehouse assignment. If so, an
   inline error alert is displayed on the page, and the commit process is       
   halted.                                                                      
 2 If all items have a warehouse, it then checks for out-of-stock items. If any 
   are found, a confirmation dialog is displayed with the reconciliation details in a `<details/>` element.                             
 3 If both checks pass, a standard confirmation dialog is shown.                

TBD:
Updating e2e tests

### DIFF
--- a/apps/e2e/integration/navigation_temp.spec.ts
+++ b/apps/e2e/integration/navigation_temp.spec.ts
@@ -22,7 +22,7 @@ test("should navigate using the side nav", async ({ page }) => {
 
 	// Navigate to 'outbound' page
 	await page.getByRole("link", { name: "Sale" }).click();
-	await expect(page.getByRole("main").getByRole("heading", { level: 1, name: "Outbound" })).toBeVisible();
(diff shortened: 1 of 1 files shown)


apps/web-client/src/routes/outbound/[id]/+page.svelte
```diff
@@ -603,6 +596,12 @@
 					</DropdownWrapper>
 				</div>
 			</div>
+			{#if no_warehouse_selected_commit_self.length}
```
Comment (delasous) [inline apps/web-client/src/routes/outbound/[id]/+page.svelte:599]: This takes up a lot of space and doesn't disappear, how about [the popover](https://www.melt-ui.com/docs/builders/popover) that hover's next to the commit button?

delasous:
Fixes #1007
Fixes #1008

delasous (approved):

PR #1066, opened 2025-08-03
Create note and warehouse with empty state after deletion

### LINKED ISSUE #1038: Bug: warehouse & note deletion error
I'm seeing some unexpected behaviour when deleting Outbound/Sale notes. What I observe:
1. I create a new note, add a line and then delete the note. If I then immediately click New Note, a new note is created with the previous state I just deleted.
2. This old state is kept in sequence. I create a note, add a line, create a second note with not lines, create a third with a line. I then delete the 2nd and 3rd note. The next two new notes that I create will have the same state as the ones I just deleted: 0 lines and 1 line.
3. While stepping through the above scenarios, sometimes a click on the New Note button will cause the app to crash.

fadwamahmoud:
- Use PageCenterDialog instead because old dialog caused a blank page to load when trying to navigate to a different page after deleting a note
- I added the bug in the delete note test case but to be honest it wasn't failing in the first place not sure why
- Delete notes associated with deleted warehouses and add test case

Fixes #1038

### DIFF
--- a/apps/e2e/helpers/entityList.ts
+++ b/apps/e2e/helpers/entityList.ts
@@ -33,7 +33,8 @@ export function getEntityList(_parent: DashboardNode, view: EntityListView): Ent
 			);
 		}
 
-		if (numBooks) await locator.getByText(`${numBooks} books`).waitFor();
+		if (numBooks === 1) await locator.getByText(`${numBooks} book`).waitFor();
+		if (numBooks > 1) await locator.getByText(`${numBooks} books`).waitFor();
 		if (discount) await locator.getByText(`${discount}% discount`).waitFor();
 		if (totalCoverPrice) await locator.getByText(`Total cover price: ${totalCoverPrice.toFixed(2)}`).waitFor();
 		if (totalDiscountedPrice) await locator.getByText(`Total discounted price: ${totalDiscountedPrice.toFixed(2)}`).waitFor();
--- a/apps/e2e/integration/note-transactions/outbound_note_transactions.spec.ts
+++ b/apps/e2e/integration/note-transactions/outbound_note_transactions.spec.ts
@@ -924,7 +924,7 @@ test("should auto-assign to the available warehouse after a warehouse with stock
 	await entries.assertRows([{ isbn, quantity: 1, warehouseName: "Warehouse 2" }]);
 });
 
-test("should not show transaction after the assigned warehouse is deleted", async ({ page }) => {
+test("should show transaction with no assigned warehouse after the assigned warehouse is deleted", async ({ page }) => {
 	// Setup: Create a transaction and assign it to a warehouse.
 	const dbHandle = await getDbHandle(page);
 	const isbn = "1234567890";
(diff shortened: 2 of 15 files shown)


delasous (commented):
- I'm still seeing the bug with warehouses when I click through the app (not sure what the difference is between this and the test case you've added)
- `routes/inventory/inbound/+page.svelte` could use the same fix with the dialog as you've done here with the outbound page

apps/e2e/helpers/entityList.ts
```diff
@@ -33,7 +33,8 @@ export function getEntityList(_parent: DashboardNode, view: EntityListView): Ent
 			);
 		}
 
-		if (numBooks) await locator.getByText(`${numBooks} books`).waitFor();
+		if (numBooks === 1) await locator.getByText(`${numBooks} book`).waitFor();
+		if (numBooks > 1) await locator.getByText(`${numBooks} books`).waitFor();
```
delasous, line 37: Small comment: you can use typesafe-i18n/`t` for stuff like this

apps/e2e/integration/note-transactions/outbound_note_transactions.spec.ts
```diff
@@ -942,8 +941,8 @@ test("should show no warehouse selected after the assigned warehouse is deleted"
 	// Action: Delete the assigned warehouse.
 	await dbHandle.evaluate(deleteWarehouse, 1);
 
-	// Assertion: Verify the transaction row now shows no warehouse selected.
-	await entries.assertRows([{ isbn, quantity: 1, warehouseName: "" }]);
+	// Assertion: Verify the transaction row now does not exist
+	await entries.assertRows([]);
```
delasous, line 945: Do we want this behaviour? 
I think it makes sense to clear all In Note's and their transactions that are associated with the warehouse, but an Out Note transaction could still exist, and a user could just select a new option? 
I would say this calls for two updates:
- Update the warehouse delete dialog message to notify the user that all associated Purchase notes will be deleted
- Update the "deleteWarehouse" handler so that book_transaction rows are not deleted if they are for outbound notes...
- and that this behaviour still stands, with an outbound note row showing "no warehouse selected" after

apps/web-client/src/routes/inventory/inbound/+page.svelte
```diff
@@ -160,16 +155,21 @@
...
+	}}
+	description={$LL.common.delete_dialog.title({ entity: noteToDelete?.displayName })}
+	title={$LL.common.delete_dialog.description()}
+>
+	<ConfirmDialog
+		on:confirm={() => {
+			handleDeleteNote(noteToDelete.id);
+			noteToDelete = null;
+		}}
+		on:cancel={() => open.set(false)}
```
delasous, line 172: You forgot to add `noteToDelete = null` here. Write a function once and re-use it. Same applies in `outbound/+page.svelte`

apps/web-client/src/lib/components/Melt/PageCenterDialog.svelte
delasous, line 1: Not sure I agree about the direction of the dialog set up. Lets chat about it

PR #1067, opened 2025-08-04
Vfs benchmark/first run

ikusteu:
This is a draft and shouldn't find it's way into `main`, but opening like this as it's a convenient place to provide instructions.

This includes all the work related to multi vfs selection (and it's the one I'm pulling from and submitting in chunks).
Some points:
- starting the app will run it with the default VFS (`idb-batch-atomic`)
- in order to use a different VFS (while running manually) you should store a localStorage entry:
  - key: `vfs`
  - value: one-of-the-adapters-mentioned-below
- available VFS implementations are: 
  - run in main thread: `idb-batch-atomic`, `opfs-any-context`
  - run within a dedicated worker (using Comlink wrapper): `opfs-adaptive-vfs`, `opfs-coop-sync`
- to run full "benckmark" (not really a benchmark yet, but if run with Playwright UI, you can see the duration for every step  in the UI), run: `rushx test:vfs:open` (from `e2e` folder) - this will open the Playwright UI with projects set up (different browser + vfs pairs)
- running the tests using Playwright UI:
  - top left corner shows `projects`, clicking on it will open a project selection (browser + vfs pairs)
  - you can run them all, but that will take a long long time, so running a single test, just with different implementations should give you a flavour
- if you want to check the code, the entry point(s) are:
  - `web-client/src/lib/db/cr-sqlite/db.ts:getDB` - this points to underlaying implementations
  - `web-client/src/lib/workers/sync-worker` - starts the DB for sync worker (uses the same underlaying initialisation)

Updated branch (including some fixes along the way): #1069


ikusteu:
Stale

PR #1068, opened 2025-08-04
Feat/multi vfs support

ikusteu:
Includes #1065 

- sets up VFS selection (in the code, local storage should be used to select the vfs)
- sets up logic for in-worker DBs (using Comlink to expose the expected DB interface)

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/core/worker-db.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/core/worker-db.ts
@@ -1,0 +1,104 @@
+import * as Comlink from "comlink";
+
+import type { DBAsync, StmtAsync, TMutex, TXCallback, UpdateType, _TXAsync } from "./types";
+
+import DBWorker from "./worker-db.worker?worker";
+import type { MsgInit } from "./worker-db.worker";
+
+export async function getWorkerDB(dbname: string, vfs: string): Promise<DBAsync> {
+	const wkr = await initWorker(dbname, vfs);
+
+	const ifc = Comlink.wrap<DBAsync>(wkr);
+	const [__mutex, siteid, filename, tablesUsedStmt] = await Promise.all([ifc.__mutex, ifc.siteid, ifc.filename, ifc.tablesUsedStmt]);
+
+	return new WorkerDB(ifc, __mutex, siteid, filename, tablesUsedStmt);
+}
+
+function initWorker(dbname: string, vfs: string) {
+	const name = [dbname, vfs].join("---");
+	const wkr = new DBWorker({ name });
+
+	return new Promise<Worker>((resolve, reject) => {
+		const listener = (e: MessageEvent) => {
+			console.log("[worker message] msg received", JSON.stringify(e.data));
+			const isInitMsg = (e: MessageEvent): e is MessageEvent<MsgInit> => e.data?._type === "wkr-init";
+			if (!isInitMsg(e)) return;
+			switch (e.data.status) {
+				case "ok": {
+					console.log("[worker message] ready!");
+					wkr.removeEventListener("message", listener);
+					return resolve(wkr);
+				}
+				case "error": {
+					console.log("[worker message] error!");
+					wkr.removeEventListener("message", listener);
+					const err = new Error(e.data.error);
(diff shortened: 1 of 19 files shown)


ikusteu:
UPDATE: added a commit from a different branch (#1069) fixing an error related to multi-vfs support (a particular non-serialisable value not being wrapped in `Comlink.proxy`) which wouldn't get caught in current CI (due to no worker db being the default one and this branch not doing full E2E test for every VFS)

Anyway, this should be ready to go

apps/web-client/src/lib/db/cr-sqlite/core/worker-db.ts
```diff
@@ -0,0 +1,104 @@
...
+
+	createFunction(name: string, fn: (...args: any) => unknown, opts?: Record<string, any>) {
+		return this.remote.createFunction(name, Comlink.proxy(fn), opts);
+	}
+
+	onUpdate(cb: OnUpdateCallback): () => void {
+		// NOTE: everything done over the wire is a Promise, whereas 'onUpdate' signature expects the unsubscribe function
+		// to be returned immediately, so we create a function that (internally) waits for the unsubscribe and calls it
+		const res = this.remote.onUpdate(Comlink.proxy(cb));
+		return () => res.then((unsubscribe) => unsubscribe()); // Everything done over the wire is a Promise
```
delasous, line 90: Should we be handling if this rejects?

apps/web-client/src/lib/db/cr-sqlite/core/worker-db.ts
```diff
@@ -0,0 +1,104 @@
...
+					}
+					return reject(err);
+				}
+			}
+		};
+		wkr.addEventListener("message", listener);
+	});
+}
+
+class WorkerDB implements DBAsync {
```
delasous, line 47: Some suggestions from C-GPT:

You construct a new Worker in initWorker, then return a WorkerDB that only holds the Comlink interface, not the actual Worker reference. There’s a close() that calls remote.close(), but you never guarantee the underlying worker is terminated; nor do you Comlink.releaseProxy on returned proxy objects (e.g., prepared statements) when they’re done. Over time, that can leak both sides.

Suggested fix
- Have WorkerDB keep:
    - the Worker instance,
    - any long-lived proxied objects (e.g., stmts you return), and
    - a dispose() that:
        - calls remote.close() (DB close),
        - worker.terminate(),
        - iterates & Comlink.releaseProxy on any outstanding proxy refs.
- For prepare(), return a small wrapper with an explicit finalize() that releases the proxy, and document expected usage. (If your StmtAsync already has finalize(), ensure it’s actually invoked in app code). You’re already returning Comlink.proxy(stmt), so the above is a small structural addition to ensure cleanup

apps/web-client/src/lib/db/cr-sqlite/core/worker-db.worker.ts
```diff
@@ -0,0 +1,111 @@
...
+
+import { getCrsqliteDB } from "./init";
+
+export type MsgInitOk = { _type: "wkr-init"; status: "ok" };
+export type MsgWkrError = { _type: "wkr-init"; status: "error"; error: string; stack?: string };
+export type MsgInit = MsgInitOk | MsgWkrError;
+
+async function start() {
+	try {
+		const [dbname, vfs] = self.name.split("---") as [string, VFSWhitelist];
```
delasous, line 13: Could be helpful to note that this is constructed in line 18 of `./worker-db.ts`. Took me a sec to figure this out

ikusteu:
@delasous saw your comments, and I'd like to tackle the last two before this goes in, but am having a break atm, so I'd prefer it if this wasn't merged before.

ikusteu:
@delasous I wouldn't go with the fixes ChatGPT gave you, at least not with this unit of work:
- the issues seem to be edge case (memory leaks in JS)
- the suggested fix would work (in theory), but it would hold reference to every (proxied) object created, thus preventing GC of those objects
- implementing this might be possible with `WeakMap`s, but I don't think it's prudent to dwell on this at this time.

The points are valid though and I'll open up issues to revisit.

PR #1071, opened 2025-08-07
Feat/db management touchups

ikusteu:
Some touchups regarding DB management: select db, create new db, delete db, import / export db.

I made some choices to reduce the complexity as much as possible:
- the aforementioned functionality is only available with a VFS supporting OPFS
- only `.sqlite3` files are supported (for import), anything else will be **silently ignored**
- the imported DB is named the same way as the imported file (e.g. `dev.sqlite3` will produce a DB with the same name)
- if importing a DB with the same name as an existing one, the **existing one is overwritten (silently)** so beware

I've done this as so since every other, more user friendly, avenue would introduce additional complexity and wouldn't be trivial to implement. Any additional touchups should then be made as an additional unit of work.


ikusteu:
Check for existing DBs in your OPFS, from browser console, run:
```js
dir = await window.navigator.storage.getDirectory()
for await (const [fname] of dir.entries()) {
  console.log(fname)
}
```
If you see the file with the same name as the requested DB, delete it and try again. Let me know if this works or not either way.

(to delete it just do `await dir.removeEntry("dump.libraio(1).sqlite3")`

PR #1088, opened 2025-08-28
Feat/additional vfs support

ikusteu:
# 🛸 PR WITHOUT DESCRIPTION: ALIEN HOSTAGE SITUATION 🛸

I did not provide a description for this PR. I had a good reason though.
The only good reason.
A group of aliens threatened to kill me if I put a single word in this PR description.

> *If you're reading this template instead of a real description, good news: there are no actual aliens! The PR author is probably just making coffee. Or fighting ninjas. Or both.*

---
*Pro tip: Aliens are surprisingly reasonable about bullet points and code snippets* 🖖


delasous:
Could use a description? I will check out alongside #1069

delasous (approved):

PR #1094, opened 2025-10-01
add a debounce on sync dialog showing:

### LINKED ISSUE #1093: Do not show modal for small sync operations
Currently whenever there's a sync operation, a modal is shown and the whole UI flashes.
The modal is there mainly for the initial sync that takes a long time. Most of sync events are not like that: they are very fast.

Ideally we should not show the modal when a quick sync operation is performed.

ikusteu:
* don't show the dialog immediately when syncing (show with timeout)
* make the (debounce) timeout configurable k

Fixes #1093

### DIFF
--- a/apps/web-client/src/routes/+layout.svelte
+++ b/apps/web-client/src/routes/+layout.svelte
@@ -187,7 +187,23 @@
 	} = createDialog({
 		forceVisible: true
 	});
-	$: $syncDialogOpen = $progress.active;
+
+	// In order to prevent sync dialog flashing for moinor changes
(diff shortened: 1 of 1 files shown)


ikusteu:
@silviot nah...it's a known ambiguity (at least to me...): the return type of ` setTimeout` is shown as `NodeJS.Timeout` (sometimes), sometimes as `number`...idk...`any` works in this case as it's obvious

delasous:
@silviot leaving this for you to tweak timeout if needed

PR #1107, opened 2025-10-10
Stock search improvements

### LINKED ISSUE #1106: Improve stock search
It has a couple of problems:

* The 13 character limit should not apply to stock search
* for a book to match, the complete search term must be present in either author, title or isbn - you won't find _Blade Runner_ by searching _Philip Dick_
* searching for "Rocky 3" will match the 3 against the ISBNs too, which doesn't make much sense

silviot:
Fixes #1106

* Removes 13 char limitation
* handles search by isbn separately (only when search term is exclusively numeric)
* divides search term into words and searches books where all of them appear either in title or author


ikusteu (approved):
Just waiting for green CI

PR #1114, opened 2025-10-14
Apply the optimization also to the initial sync

silviot:
* add a bit of doc about the sqlite file fetching and re-identifying
* use the fetch-and-reidentify strategy also for initial sync
* detect a 404 or a bad sqlite file after fetching, and don't use it if it's faulty

### DIFF
--- a/.github/workflow.templates/build-crsqlite.lib.yml
+++ b/.github/workflow.templates/build-crsqlite.lib.yml
@@ -8,13 +8,12 @@ steps:
   - uses: actions/checkout@v4
     with:
       submodules: recursive
-      lfs: false
+      lfs: true
       fetch-depth: 0 #! We need all the history to check artifact date
   - name: Test if artefacts are stale
     id: ts
     shell: bash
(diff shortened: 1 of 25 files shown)


ikusteu (approved):
I think this should go in. 
I see the `(...)/core/init.ts` had been reverted to use `initWasm`, so not much change there and I've changed that API in a different branch (both in `vlcn-io/js` as well as our usage), so I won't propose any additional changes here.

@silviot I don't think there's an issue with type definitions of `initWasm` (there definitely is one incorrect typing, but that shouldn't cause problems in `(...)/cr-sqlite/core/init.ts`). 

One potential cause for inconsistency I see is that you may have run `./scripts/build_vlcn.sh` without `rush update` afterwards. If you have, I don't know what's causing this, but I've definitely run and rerun the build script countless times (due to some issues I had) and didn't experience this issue. 

However, this uses the somewhat old (stitched-up) `vlcn.io/js` API which I've changed in #1109 anyway

PR #1117, opened 2025-10-16
Feat/app wide error boundary

### LINKED ISSUE #902: New button in the debug page: raise exception
To properly test sentry setup we need to be able to manually trigger an exception. A button in the debug page should work.

### LINKED ISSUE #1040: Improve error boundaries - when an exception occurs the app becomes frozen
If you trigger #1039 the application becomes completely unresponsive.

We need to be more defensive, and have somthing in place to prevent a single exception from freezing the whole app.

\---

Add error.html at root of source and/or investigate if anything else needs to be done to make sure app is still usable/can be recovered after exception

ikusteu:
Fixes #902 
Fixes #1040 

Not sure if it fixes those issues to the fullest. Current sutiation:

- create `ErrorView` (component) - used to render errors
- create `+error.svelte` - renders `ErrorView` with `load` time error (this catches all errors that happen at load time)
- create runtime error catching logic:
  - in root `+layout.svelte`
  - add `<svelte:window on:error={handleRuntimeError}>`  (doesn't prevent the error, merely a way to react)
  - store error in local state (variable)
  - if error - display `ErrorView` (this time with `runtime` variant)
- add 2 buttons in `/debug` page: 
  - trigger load error (navigating to `/load_error` route - simply throws an error during `load` function
  - trigger runtime error 
  
This should:
- help catch load time errors and provide feedback (something **did** in fact crash) -- allowing us to add a spinner for loading (avoiding blank screen when loading, and showing the error if crash)
- help trigger errors for Sentry (my understanding is that all errors thrown will get reported)

Things to consider:
- currently every runtime error "crashes" (displays `ErrorView`) - this makes sense as we should probably handle all expected errors and have unexpected errors crash, but it may be too strict
- how do we communicate the unknown error to the user (pure UX perspective)

### DIFF
--- a/apps/web-client/src/routes/debug/+page.svelte
+++ b/apps/web-client/src/routes/debug/+page.svelte
@@ -293,7 +293,7 @@
 	const triggerLoadError = () => goto("#/load_error");
 
 	const throwError = () => {
-		throw new Error("Kaboom! Runtime error");
+		throw new Error($LL.debug_page.labels.runtime_error());
 	};
 
 	let error = false;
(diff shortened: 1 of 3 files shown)


apps/web-client/src/routes/+error.svelte
delasous: I don't agree with these changes for loading errors. The existing `error.svelte` that is being wiped here is much less jarring than the `ErrorView` and it gives the user a chance to navigate away. I'd say here we'd at least want to add the Reload button.

delasous (approved):

PR #1128, opened 2025-11-06
color the time badges on customer order lines to indicate order line status

### LINKED ISSUE #1118: Color code rows in customer book list: 1 pt
In the customer order view, being able to tell at a glance how many books are in which state is very useful.

We should introduce some kind of color coding. Maybe changing the background color for the whole row.

We need four codes. The current ones are:

* no change for pending
* `#FFD700` `yellow` ordered to the supplier
* `#2E8B57` `green` delivered from the supplier
* `#DC143C` `red` collected by customer

ikusteu:
Fixes #1118

### DIFF
--- a/apps/web-client/src/routes/orders/customers/[id]/+page.svelte
+++ b/apps/web-client/src/routes/orders/customers/[id]/+page.svelte
@@ -328,29 +328,24 @@
 									<td>{publisher}</td>
 									<td>
 										{#if status === OrderLineStatus.Collected}
-											<div class="badge-primary badge-outline badge bg-red-400 text-xs font-semibold">
+											<div class="badge-outline badge orderline-collected text-xs font-semibold">
 												{$LL.customer_orders_page.status.collected()} -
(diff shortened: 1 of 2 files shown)


silviot:
I think we should give these colors some names. This PR uses them literally in the code, and it works.

But I think we should have them configurable somewhere. We might also want two different variants for light and dark theme.

ikusteu:
> I think we should give these colors some names. This PR uses them literally in the code, and it works.
>
> But I think we should have them configurable somewhere. We might also want two different variants for light and dark theme.

I was thinking the same thing (you were explicit in the issue spec though, so I went with this).

@delasous care to chime in here? (I understand this should best be solved from design tokens point of view, which I know nothing about)

ikusteu:
Righht...you were probably specific to leverage nice MD formatting (not related to the issue impl.), we could just make them red/gree/yellow (e.g. `bg-red-300` or something)

PR #1159, opened 2025-12-10
Lock some deps for acute unblocking of development

ikusteu:
Bumps zod `3.24.1` -> `3.25.0` (required by sveltekit-superforms). This just started being an issue when building/using dev server. I couldn't find a blame that caused it, but this seems to work (respects the `peerDependency` without breaking).

**edit:** includes additional effort to fix linting and runtime breaking, thus:
- updates (and locks) `zod` to version `3.25.0`
- locks svelte to version `5.35.3` - newer (minor) versions were causing runtime to break
- locks prettier to version `~3.4.2` - this didn't play well with `svelte-`something plugins (failing at formatting/linting) so it's locked for the time being

Also added: `LOCKED_DEPS.md` at the root of the repo, documenting locked deps and reason for doing so.

**Ivan:** I've also sneaked in a small `PLAYWRIGHT_FULLY_PARALLEL` functionality (adding an env variable for opt-in `fullyParallel: true` testing -- speeds thing up for me considerably, but might cause issues on slower machines -- hence opt-in. **shouldn't affect any of the aforementioned work**)

### DIFF
--- a/apps/e2e/helpers/cr-sqlite.ts
+++ b/apps/e2e/helpers/cr-sqlite.ts
@@ -43,7 +43,7 @@ export function getDbHandle(page: Page) {
 			} else {
 				// Creating a separate function, as we want to run the listener only once and then remove it
 				const finalise = () => {
-					(window.removeEventListener("db_ready", finalise), res());
+					window.removeEventListener("db_ready", finalise), res();
 				};
 				window.addEventListener("db_ready", finalise);
 			}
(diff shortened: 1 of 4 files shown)


ikusteu:
Playwright matrix seems to be broken...but that's an issue for another day

LOCKED_DEPS.md
```diff
@@ -0,0 +1,24 @@
+# Locked deps
+
+This is a file listing depencencies whose versions we had to lock as they were breaking.
```
silviot: What does "locking a dependency" mean? If it means "only allowing a single version", then we're not locking prettier when we specify `~3.4.2`.

PR #1161, opened 2025-12-15
Set `ensureConsistentVersions` to `true` and align all dependencies versions across packages in the monorepo

silviot:
Upgrades dependencies and enforces version consistency across the monorepo:

  - Zod: 3.25.0 → 3.25.76 (fixes typing issues)
  - @vlcn.io/crsqlite: 0.16.1 → 0.16.3 (prevents the need to compile in the CI)
  - TypeScript: ^5.5.0 → ^5.7.3
  - Various tooling: prettier, eslint parsers, vite, vitest

  Adds version consistency enforcement and updates CI workflows to use rush install instead of rush update. Fixes typing problems introduced by zod upgrade.

[Claude's analysis of the zod problem is interesting](https://gist.github.com/silviot/31763e2a45331805d495e492ef50353b): it deems the inclusion of both zod 3 and for into 3.25.x as the root cause of our problems, I believe.

### DIFF
--- a/apps/e2e/helpers/cr-sqlite.ts
+++ b/apps/e2e/helpers/cr-sqlite.ts
@@ -43,7 +43,8 @@ export function getDbHandle(page: Page) {
 			} else {
 				// Creating a separate function, as we want to run the listener only once and then remove it
 				const finalise = () => {
-					window.removeEventListener("db_ready", finalise), res();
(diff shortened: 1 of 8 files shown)


silviot:
Note to self: stop letting agents run wild. Keep the steering wheel in your hands!

PR #1234, opened 2026-04-03
feat(db): SharedWorker — fix multi-tab OPFS lock contention

silviot:
## Opening librocco in a second tab caused a 30-second hang

Every time a user opened librocco in a new tab while another tab was already running, the second tab would hang for 30 seconds and then show an error dialog. This happened silently in production — the user saw a broken app with no clear explanation.

The root cause is deep in how the database works. Librocco uses OPFS (the browser's Origin Private File System) for local storage via the `sync-opfs-coop-sync` VFS. OPFS uses an exclusive synchronous file lock — only one JavaScript context can hold the lock at a time. When a second tab opened, it had to negotiate with the first tab (via BroadcastChannel) to take over the lock. Chrome throttles JavaScript in background tabs by 30 seconds or more, so by the time the first tab could respond, the second tab had already timed out.

**This PR fixes the problem at its source** by moving the database into a SharedWorker. A SharedWorker has a single shared instance across all tabs of the same origin. Instead of each tab competing for the OPFS lock, they all connect to one worker that holds the lock permanently. The second tab connects in ~50ms instead of timing out.

---

### Before: one DedicatedWorker per tab

```
Tab 1                Tab 2
  │                    │
  ▼                    ▼
DedicatedWorker    DedicatedWorker
  │                    │
  └── OPFS lock ───────┘  ← contention
```

Tab 2 enters the `sync-opfs-coop-sync` cooperative handoff protocol, broadcasting a "please release" message to Tab 1. Chrome throttles Tab 1's background JS. Tab 2 waits 30s. `ErrDBInitTimeout`.

### After: one SharedWorker per DB

```
Tab 1    Tab 2    Tab 3
  │        │        │
  └────────┴────────┘
           │
      SharedWorker
           │
        OPFS lock  ← one holder, forever
```

All tabs connect via `MessagePort`. The SharedWorker is identified by `<dbname>---<vfs>` name — the browser guarantees one instance per origin per name. No lock negotiation needed. This applies to both the main DB and the demo DB, which share the same `getWorkerDB` code path.

---

### `worker-db.worker.ts` — dual-mode entry point

The same worker file now handles both SharedWorker and DedicatedWorker (iOS fallback). Duck-typing detects the runtime: `"onconnect" in self` is only true in `SharedWorkerGlobalScope`.

- **SharedWorker path**: `onconnect` handler, one DB singleton via `getOrInitDB()`, Comlink exposed per-port
- **DedicatedWorker path** (unchanged): original `start()` function
- **`getOrInitDB()` singleton**: 30s internal timeout, leaked-handle cleanup if a race resolves after timeout, `_dbPromise = null` reset so a fresh SharedWorker after page reload gets a clean start

### `worker-db.ts` — complete rewrite

- **`getSharedWorkerDB()`**: creates a `?sharedworker` import, waits for the "wkr-init" handshake, wraps with Comlink
- **`getDedicatedWorkerDB()`**: original code, renamed
- **`disconnectAllPorts()`**: replaces `terminateAllWorkers()` — closes `MessagePort`s (SharedWorker) and terminates `Worker`s (iOS DedicatedWorker)
- **`WorkerDB` constructor**: takes `teardown: () => void` instead of a `Worker`, abstracting port-close vs worker-terminate
- **Graceful fallback**: if SharedWorker init fails (e.g. Firefox, which doesn't support `createSyncAccessHandle` in SharedWorkers, or Playwright headless Chromium in the test environment), `getSharedWorkerDB` catches the error and falls back to `getDedicatedWorkerDB` automatically

### Browser compatibility

| Environment | Path | Multi-tab |
|---|---|---|
| Chrome 108+ | SharedWorker | Fixed ✓ |
| Firefox | SharedWorker init fails → DedicatedWorker fallback | Same as before (Firefox lacks SharedWorker SAH) |
| iOS Safari | `typeof SharedWorker === "undefined"` → DedicatedWorker | Single-tab as before |
| Playwright E2E (Chromium) | SharedWorker | Fixed ✓ |
| Playwright E2E (Firefox) | DedicatedWorker fallback | Same as before |

### Tests

**`apps/e2e/integration/multi-tab.spec.ts`** (new) — 4 tests covering the SharedWorker behavior:
1. Two tabs open the same DB — both reach `db_ready` without timeout
2. Closing tab 1 does not break tab 2
3. Reloading a tab reconnects within timeout
4. Three tabs — all reach `db_ready`

These pass on both Chromium and Firefox in CI (Playwright doesn't throttle background tabs, and DedicatedWorker coop-sync is fast enough without active sync contention).

**`apps/e2e/integration/sync.spec.ts`** — The existing multi-tab sync test keeps its `asyncify-idb-batch-atomic` VFS override. That test verifies sync protocol behavior (status consistency, reconnect, data propagation) in a two-tab scenario where an active sync connection is running. Firefox's DedicatedWorker fallback + active sync contention causes the OPFS handoff to stall, so the VFS override keeps the test stable on all browsers. The actual multi-tab OPFS contention fix is what `multi-tab.spec.ts` covers.

### Open deferred work

One item remains in `TODOS.md`: the `wkr-close` protocol for `debug-export`. `writeDbBytesToOPFS` currently calls `disconnectAllPorts()` then waits a fixed 300ms before writing, relying on browser GC to release the SharedWorker's OPFS handle. A deterministic `wkr-close` / `wkr-close-ack` message exchange would remove the race, but it's a developer-flow-only path and the current 300ms works in practice.

---

## Test plan

- [x] All existing E2E tests pass (CI green, all 6 shards)
- [x] New `multi-tab.spec.ts` — 4 tests pass (Chromium + Firefox)
- [x] Web Client unit tests pass
- [x] Lint and typecheck clean
- [x] Open librocco in 2 tabs — both reach `db_ready` (manual verification via preview deploy)
- [ ] iOS Safari — single tab still works (manual, no CI coverage)

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -1,0 +1,34 @@
+# TODOS
+
+Deferred work captured during engineering review.
+
+---
+
+## SharedWorker debug-export teardown
+
+**What:** `writeDbBytesToOPFS` in `src/lib/utils/debug-export.ts:78` calls `disconnectAllPorts()` then waits 300ms before writing to OPFS. With SharedWorker, the OPFS SyncAccessHandle release is async (happens when the worker GC runs after all ports close). The 300ms is a timing race — works in practice on fast machines but is not deterministic.
+
+**Why:** When the SharedWorker DB migration lands, this becomes a latent bug for the import-state-archive developer flow. On a slow device or under load, the 300ms may not be enough for the worker GC to release the handle, causing the OPFS write to fail.
+
+**Fix:** Add a `wkr-close` message protocol:
+1. Worker tracks `connectedPorts` count in `onconnect` / on `wkr-close` message
+2. On `wkr-close` message from main thread: decrement count; when count hits 0, explicitly `await db.close()` (releases SyncAccessHandle), then send `wkr-close-ack`
+3. Main thread: `disconnectAllPorts()` sends `wkr-close` to each port, awaits ack before proceeding with file write
+
+**Files:** `src/lib/db/cr-sqlite/core/worker-db.worker.ts`, `src/lib/utils/debug-export.ts`
+
+**Depends on:** SharedWorker DB migration landing first.
+
+---
+
+## Demo DB SharedWorker migration
+
+**What:** `initializeDemoDb` in `src/lib/app/db.ts` uses `DEMO_VFS` (same `sync-opfs-coop-sync`) but a different DB file (`DEMO_DB_NAME`). It currently calls `getDBCore()` which goes through the worker path, but demo DB init has its own lifecycle separate from the main DB SharedWorker.
+
+**Why:** After the main DB SharedWorker migration, demo DB still runs a DedicatedWorker. If a user opens demo mode in two tabs, they hit the same OPFS lock contention bug (just for the demo DB). Low probability but inconsistent.
+
+**Fix:** Apply the same SharedWorker pattern to demo DB initialization. Same `onconnect`/`getOrInitDB` approach, different dbname key (`DEMO_DB_NAME---<vfs>`).
+
+**Files:** `src/lib/app/db.ts` (`initializeDemoDb`), potentially `src/lib/db/cr-sqlite/core/worker-db.ts`
+
+**Depends on:** SharedWorker DB migration landing and proving stable in production (at least one release cycle).
--- a/TODOS.md
+++ b/TODOS.md
@@ -18,17 +18,3 @@ Deferred work captured during engineering review.
 **Files:** `src/lib/db/cr-sqlite/core/worker-db.worker.ts`, `src/lib/utils/debug-export.ts`
 
 **Depends on:** SharedWorker DB migration landing first.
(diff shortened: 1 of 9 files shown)


delasous:
I don't think this approach will work. The [`createSyncAccessHandle` is not available inside a SharedWorker](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createSyncAccessHandle). rhashimoto outlines this in [this discussion](https://github.com/rhashimoto/wa-sqlite/discussions/81) along with a proposed ["shared service" architecture](https://github.com/rhashimoto/wa-sqlite/tree/master/demo/SharedService) - where additional tabs are essentially deligating work to a "lead" tab which is in direct comms with the DB in the dedicated worker. The challenge with this approach is handling "leader migration", when the leader tab closes

chiruzzimarco:
Closing is favor of #1235 since the approach taken here is not going to work.

PR #1243, opened 2026-06-01
fix(stock): never surface negative stock in getStock (D-293/D-217)

silviot:
## Problem
The bookshop reported **negative copies** showing in magazzino search on the cash-register PC. Root cause (investigated + reproduced — see Linear **D-217**/**D-293**): stock is *derived* as `SUM(+inbound / -outbound)` per `(isbn, warehouse_id)`, and `getStock`'s `HAVING … != 0` only filtered **zero**, so a **negative** sum was returned verbatim and rendered to the cashier.

This goes negative whenever the ledger is unbalanced — most notably when a warehouse with committed sales is deleted: `deleteWarehouse()` removes the inbound (+) rows and relocates the outbound (−) sale lines to "no warehouse" (wh 0), leaving a net‑negative there with no matching inbound. (Verified deterministically with the real cr‑sqlite extension: a single `deleteWarehouse` strands `-2` at wh 0 on both converged peers.)

## Fix
Physical stock can never be negative, so **clamp the read layer**: `HAVING SUM(...) > 0` (was `!= 0`). The UI never shows "negative copies"; the underlying imbalance is left for **monitoring** to surface, not the cashier. This matches `getStock`'s own documented contract (which already claimed "negative stock is prevented").

> The deeper `deleteWarehouse` PK‑rewrite / soft‑delete redesign (D‑293) is intentionally **out of scope** here — it carries a product decision (keep vs. drop the sale line of a deleted warehouse). This PR fixes the user‑visible symptom safely and is independently correct.

## Forward-looking guard (replaces a weak test)
The earlier throwaway single‑scenario repro only pinned the one known bug. It's replaced with a real **invariant / property harness** (`sync-invariants.test.ts`) that drives the **actual** `warehouse`/`note`/`stock` operations across two synced cr‑sqlite peers and asserts **general invariants** — *getStock never returns a negative*, and *peers converge* — over both the incident scenario and a **seeded fuzz** of random operation/sync interleavings. This guards the whole class ("a converged‑but‑invalid state leaks negative stock"), not just this bug.

## Testing
- `rushx test:ci` for `stock`, `warehouse`, and the new `sync-invariants` suites — **19 passed** (incl. the incident scenario + 40‑iteration seeded fuzz).
- `rushx typecheck` — 0 errors. `rushx lint` — clean.

Refs: D-217, D-293, D-297

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/__tests__/sync-invariants.test.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/__tests__/sync-invariants.test.ts
@@ -1,0 +1,156 @@
+/**
+ * Sync invariant + property tests.
+ *
+ * Unlike a single hand-built scenario, these drive the REAL librocco operations
+ * (warehouse.ts / note.ts / stock.ts) across two synced cr-sqlite peers and assert
+ * GENERAL invariants that must hold no matter what sequence of operations or sync
+ * interleavings occur:
+ *
+ *   INV-1  getStock never surfaces a negative quantity (physical stock is never < 0).
+ *   INV-2  After syncing to quiescence, both peers agree on stock (convergence).
+ *
+ * This is the forward-looking guard: it catches the class of "a converged-but-invalid
+ * state leaks negative stock to the UI" — including the D-217/D-293 warehouse-deletion
+ * bug — rather than pinning one known reproduction. The deterministic scenario encodes
+ * the incident; the seeded fuzz explores interleavings no one wrote by hand.
+ *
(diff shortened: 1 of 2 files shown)


silviot:
Abandoning this approach. Clamping `getStock` to `> 0` hides a real data-integrity error instead of fixing it: a negative derived stock is a true signal that the ledger is unbalanced, and suppressing it at the read layer makes a corrupt database look healthy (errors should never pass silently). The correct fix is to prevent the imbalance at its source in `deleteWarehouse` (don't remove one side of a balanced transaction set), not to mask the symptom.

PR #1244, opened 2026-06-01
feat(sync-server): surface negative committed stock as a /health warning (D-296)

silviot:
## Why
Investigation of the negative-stock incident (Linear **D-217**/**D-293**) showed the cheapest, highest-value detector for the whole class is a **negative committed-stock invariant** — and it already fires on production today. This wires that invariant into the sync server so a recurrence is caught automatically instead of when the shop phones.

## What
Adds a `negative_stock` check to `checkDatabaseHealth`. Stock is derived as `SUM(+inbound/+reconciliation, −outbound)` per `(isbn, warehouse_id)` over committed notes; physical stock can never be `< 0`. A negative sum means the ledger is unbalanced — e.g. a warehouse with committed sales was deleted, stranding the outbound (−) lines with no matching inbound (D-293/D-217).

It surfaces in `GET /health` and `GET /:db/health` (each lists per-check `passed`/`severity`/`message`), so an external monitor (Uptime Kuma keyword match on the body) can alert.

### Deliberately a WARNING, not an error
`ok` is computed only from `error`-severity failures, and startup exits / `/health` returns 503 only when `!ok`. A data imbalance must **not** stop the shop or refuse startup, so this is `severity: "warning"`: visible to monitoring, invisible to availability. It also **skips gracefully** on non-librocco DBs (no `note`/`book_transaction` table).

## Testing
- `rushx test:ci` (sync-server): **28 passed**, including 3 new cases — negative (warning, db still `ok`), balanced (passes), and graceful-skip on a non-librocco schema.
- `tsc --noEmit`: no new errors in the changed files. prettier-clean.

Pairs with the read-layer clamp in #1243 (which stops the cashier *seeing* negatives) — this PR makes the underlying imbalance *observable*.

Refs: D-296, D-293, D-217

### DIFF
--- a/apps/sync-server/src/db-health.test.ts
+++ b/apps/sync-server/src/db-health.test.ts
@@ -59,6 +59,31 @@ function createCorruptDb(name: string): string {
 	return dbPath;
 }
 
+// Creates a minimal librocco-shaped DB (note + book_transaction CRRs) with a
+// committed inbound of +5 and a committed outbound sale. With `outboundQty` > 5
+// the derived stock for the (isbn, warehouse) pair goes negative.
+function createDbWithStock(name: string, outboundQty: number): string {
+	const dbPath = path.join(TEST_DIR, name);
+	const db = new Database(dbPath);
+	db.loadExtension(extensionPath);
+	db.exec(`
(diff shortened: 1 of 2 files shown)


silviot:
Closing — this puts a heavy invariant scan on `/health`, which is the wrong place. `/health` already runs a full `integrity_check` (~2s on the production DB) per call, and an invariant check should not run synchronously against the live serving DB on a polled endpoint. The right design (D-296) is a scheduled job over a DB snapshot pushed to monitoring — zero cost on `/health`, no contention with live sync.

PR #1245, opened 2026-06-01
fix(sync-server): vendor ws-server gap-recovery fix (D-302)

silviot:
## What
Brings the sync server's vendored `@vlcn.io/ws-server` up to include the **OutboundStream gap-recovery fix** (fork PR [librocco/vlcn-js#1](https://github.com/librocco/vlcn-js/pull/1), merged).

`OutboundStream.reset()` was a no-op, so when a peer detected a gap and sent `RejectChanges`, the server never rewound or re-sent — a peer that fell behind mid-connection then silently and permanently diverged while still reporting "synced". This is the delivery-layer mechanism behind the negative-stock incident (**D-217**); it was diagnosed by comparing two live replicas, which differed by ~36k changes yet converged once the missing changes were supplied (i.e. a delivery gap, not a CRDT merge bug).

## Changes
- `3rd-party/artefacts/vlcn.io-ws-server-0.2.2.tgz` — rebuilt with the fix (and uploaded to the R2 artefact store, keyed by its new sha256, so CI can pull it).
- `3rd-party/js` — repointed at `librocco/vlcn-js@main` (`e99c5d7`); updates `artefacts_version.txt` / `version-cached.txt`. Also corrects a pre-existing skew where the submodule referenced `vlcn-io/js` upstream while the artefacts already carried librocco's forked fixes.
- `pnpm-lock.yaml` — refreshed via `rush update`.
- `apps/sync-server/src/ws-server-reject-recovery.test.ts` — installed-tarball regression test (same pattern as `peer-coherence.test.ts`) so a stale/un-rebuilt artefact fails CI.

## Testing
- `rushx test:ci` (sync-server): **26 passed**, including the new regression test and the existing `peer-coherence` installed-tarball test.
- Fork change verified red→green in isolation against the real `OutboundStream` source.

Server-only; clients already send `RejectChanges` correctly. Deploying needs only the sync server. A separate data-cleanup (orphaned wh-0 rows from D-293) is tracked independently.

Refs: D-302, D-217, D-293

## Release Notes

* **Chores**
  * Updated submodule configuration for JavaScript dependencies and refreshed cached version hashes.
  * Updated Git LFS pointers for archived third-party artefacts with new references.

* **Tests**
  * Added regression test to verify correct synchronisation recovery behaviour when peers detect gaps in change streams and request resynchronisation.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "3rd-party/js"]
 	path = 3rd-party/js
-	url = git@github.com:vlcn-io/js
-	branch = wa-sqlite-updates
+	url = git@github.com:librocco/vlcn-js
+	branch = main
 [submodule "3rd-party/typed-sql"]
 	path = 3rd-party/typed-sql
 	url = git@github.com:vlcn-io/typed-sql
--- a/3rd-party/artefacts/version-cached.txt
+++ b/3rd-party/artefacts/version-cached.txt
@@ -1,2 +1,2 @@
-js repo hash: 435b75d4d9b499c89ff10b9e12e44c305643971d
+js repo hash: e99c5d700a7e4da9712ab10be1249a296cf72384
 typed-sql repo hash: bed447324d1d078d30eedc73504ac99e7874fe31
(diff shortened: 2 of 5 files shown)


chiruzzimarco:
@coderabbitai review

PR #1246, opened 2026-06-01
fix: delete warehouse stock symmetrically to avoid phantom negatives

silviot:
## Problem

`deleteWarehouse` removed only the **inbound** transaction legs (selected by note ownership) and then **reassigned the orphaned outbound legs to the sentinel warehouse `0`**. Those surviving negative legs surfaced as phantom **negative stock** attributed to warehouse 0.

## Fix

Both the positive (inbound / reconciliation) and negative (outbound) legs of a warehouse carry `book_transaction.warehouse_id = id`:

| note type | `note.warehouse_id` | `bt.warehouse_id` | stock sign |
|---|---|---|---|
| inbound to X | `X` | `X` | `+qty` |
| reconciliation touching X | `NULL` | `X` | `+qty` |
| outbound sold from X | `NULL` | `X` | `−qty` |

So for **committed** notes, deleting by `warehouse_id` removes the warehouse's entire stock contribution symmetrically — every `(isbn, warehouse)` group goes away together, nothing is re-tagged, and no negative quantity can be synthesised on warehouse 0 (or anywhere else). Multi-warehouse outbound/reconciliation notes keep their other-warehouse legs.

**Draft notes are the exception**: lines in an open note someone may be actively editing (e.g. an in-progress outbound note) are kept but unassigned (sentinel warehouse `0`) instead of deleted, preserving the previous UX for work-in-progress. This can't reintroduce phantom stock — drafts don't count toward stock (`getStock` filters on `committed = 1`) and commit validation rejects unassigned lines (`NoWarehouseSelectedError`).

```sql
DELETE FROM book_transaction WHERE warehouse_id = ?
  AND note_id IN (SELECT id FROM note WHERE committed = 1);  -- committed + and − legs, symmetrically
DELETE FROM book_transaction
  WHERE note_id IN (SELECT id FROM note WHERE warehouse_id = ?);  -- legs of draft notes owned by the warehouse
UPDATE book_transaction SET warehouse_id = 0 WHERE warehouse_id = ?;  -- draft lines elsewhere: unassign, don't delete
DELETE FROM note      WHERE warehouse_id = ?;
DELETE FROM warehouse WHERE id = ?;
```

## Tests

- **Regression** (red before the fix, green after): same ISBN stocked in two warehouses plus a committed outbound sourced from the deleted warehouse → after deletion only the other warehouse's stock survives, with no negative quantity and nothing on warehouse 0.
- **Draft preservation**: a line in an open outbound note pointing at the deleted warehouse survives as unassigned (`warehouseId: 0`) and contributes no stock.

The pre-existing e2e test "should show transaction with no assigned warehouse after the assigned warehouse is deleted" still passes — draft-note UX is unchanged.

### DIFF
--- a/apps/web-client/src/lib/db/cr-sqlite/__tests__/warehouse.test.ts
+++ b/apps/web-client/src/lib/db/cr-sqlite/__tests__/warehouse.test.ts
@@ -4,6 +4,7 @@ import { getRandomDb } from "./lib";
 
 import { upsertWarehouse, getAllWarehouses, getWarehouseById, getWarehouseIdSeq, deleteWarehouse } from "../warehouse";
 import { addVolumesToNote, createAndCommitReconciliationNote, createInboundNote, createOutboundNote, commitNote } from "../note";
+import { getStock } from "../stock";
 
 describe("Warehouse tests", () => {
 	it("creates a new warehouse, using only id, with default fields", async () => {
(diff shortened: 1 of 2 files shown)


chiruzzimarco:
@coderabbitai review

PR #1247, opened 2026-06-05
Add read-only sync server query endpoint

chiruzzimarco:
## Summary
- add `POST /:dbname/readonly-query` for integrations that need authoritative reads through the sync server DB connection
- enable the endpoint by default, with `READONLY_QUERY_API=false` available to disable it
- return 404 for missing database files before using the sync cache
- reject non-reader/write-attempting statements and run accepted queries under SQLite `PRAGMA query_only = ON`
- classify client query errors as 400, server/cache failures as 500, and log failures server-side
- document the endpoint and environment variable

## Verification
- `source ~/.nvm/nvm.sh && nvm use && node common/scripts/install-run-rush.js install`
- `source ~/.nvm/nvm.sh && nvm use && node ../../common/scripts/install-run-rushx.js test:ci` in `apps/sync-server` (32 tests passed)
- `source ~/.nvm/nvm.sh && nvm use && node common/scripts/install-run-rush.js build --to @librocco/sync-server` (build completed with existing bundle-size warnings; Rush exits non-zero on warnings)

### DIFF
--- a/apps/sync-server/README.md
+++ b/apps/sync-server/README.md
@@ -30,6 +30,24 @@ Backup runs are reused per schema target version, so repeated failed restarts do
 - `SCHEMA_FOLDER`: schema folder path (default `./schemas`)
 - `SCHEMA_NAME`: schema file name in `SCHEMA_FOLDER` (default `init`)
 - `IS_DEV=true`: enables dev-only RPC endpoints
+- `READONLY_QUERY_API=false`: disables `POST /:dbname/readonly-query` (enabled by default)
 - `SKIP_HEALTH_CHECK=true`: disables startup health checks
 - `STARTUP_MIGRATION_BACKUP_FOLDER`: optional explicit backup folder (default `DB_FOLDER/.startup-migration-backups`)
 - `STARTUP_MIGRATION_MAX_BACKUP_RUNS`: number of backup run directories to retain (default `5`)
+
+## Read-only query API
+
+The read-only query API is enabled by default and is available at:
+
+```http
+POST /:dbname/readonly-query
+Content-Type: application/json
+
+{
+  "sql": "SELECT * FROM book WHERE updated_at > ? ORDER BY updated_at ASC",
+  "bind": [1234567890]
+}
+```
+
+The endpoint only accepts SQLite reader statements and executes them with `PRAGMA query_only = ON`.
+It exists for external integrations that need authoritative reads through the sync server's active DB connection without enabling the dev-only generic `/exec` endpoint.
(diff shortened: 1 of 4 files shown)


silviot:
Reviewed the new `readonly-query` endpoint with a focus on failure behavior and observability. The write-protection itself is solid (`stmt.reader` + `PRAGMA query_only` covers `UPDATE ... RETURNING` and statement smuggling), but I found three issues in the error path — the first one is a real bug.

## 1. Querying a nonexistent dbname silently *creates* the database

`useReadonlyDb` → `dbCache.use` → on cache miss, `DBCache.getAndRef` unconditionally calls `dbFactory.createDB(...)`, which does `new Database(dbpath)` — better-sqlite3's default mode creates the file if missing, and the factory then applies the schema.

That's intentional for sync (rooms are provisioned on first connect), but for this endpoint it means:

- A typo'd dbname (`librocc0`) creates a fresh, empty, schema'd DB and the query returns **`200 {"rows": []}`** — a confidently wrong answer, not an error. An integration polling for updates would see "nothing new" forever.
- Stray DB files accumulate in `DB_FOLDER` and get picked up by the startup-migration/backup machinery on the next restart.

**Suggested fix:** check the db file exists before going through the cache and return `404` if not.

## 2. Every failure maps to 400

`sendReadonlyQueryError` returns `400 "Read-only query failed"` for everything, but the catch block sees genuinely different cases:

| failure | should be | is |
|---|---|---|
| bad SQL / non-reader statement / bad bind | 400 (caller's fault) | 400 ✓ |
| schema version mismatch from `getAndRef` | 5xx (server state) | 400 ✗ |
| `"Database callback did not return a result"` sentinel | 500 (server bug) | 400 ✗ |

An integration with standard retry logic (retry 5xx, alert on persistent 4xx) will misclassify transient server faults as "my query is wrong".

**Suggested fix:** prepare/reader/bind errors → 400 with the SQLite message verbatim (callers need it to fix their SQL); everything else → 500 with a generic body.

## 3. No server-side logging

Errors are serialized to the client and dropped — nothing lands in the server logs. The serialization also keeps only `err.message`, discarding better-sqlite3's `err.code` (`SQLITE_BUSY` vs `SQLITE_ERROR` are very different diagnoses) and the stack. `DBCache` already logs via `@vlcn.io/logger-provider`, so there's an obvious logger to reuse.

**Suggested fix:** `logger.error` with dbname + `err.code` + stack on the 5xx path; optionally a `logger.info` with dbname + truncated SQL on the 400 path so client-side query mistakes are diagnosable from the server too.

Minor: the `result === undefined` guard in `useReadonlyDb` can't actually fire (the callback always returns an array, and a throwing callback propagates out of `dbCache.use`), and its message would be confusing if surfaced to an HTTP client as a 400.

silviot (changes requested):
See previous comment

chiruzzimarco:
@silviot Thanky you, should be fixed now!

silviot (approved):

PR #1250, opened 2026-06-07
docs: sync-correctness audit (warehouse domain + general architecture)

silviot:
Adds `docs/architecture/sync-correctness-audit.md` — a prioritized assessment of the cr-sqlite/vlcn
sync stack and the warehouse domain, hunting for the same **silent-divergence** class as the recently
fixed `OutboundStream.reset` stub.

Headlines:
- **Critical (verified):** upload-path gap recovery is dead in two places — the server `InboundStream`
  applies a gapped changeset and advances its cursor past the gap (missing `return` after
  `rejectChanges`), and the client `WebSocketTransport` `throw`s on an incoming `RejectChanges`. A
  single skipped upload frame permanently loses a committed note storewide. (C1 reproduced against the
  shipped `@vlcn.io/ws-server` artefact.)
- **High:** stock-cache invalidation keyed on a cross-node wall clock (stale stock while connected);
  `deleteWarehouse`/`deleteNote` leaving phantom/orphan stock under concurrent inserts; multi-tab
  leadership; unauthenticated user-editable room name; WAL-mode bootstrap snapshot.
- **Architecture:** the design is sound; the systemic mistake is deriving correctness from wall-clock
  timestamps and snapshot-time SQL DELETEs in a CRDT, plus gap-recovery paths that had no tests.

Companion fixes already open: #1248 (warehouse totals / orphan legs, finding H4), #1249 (stock-cache
logical-version invalidation, findings H1/H2). The two critical engine bugs are tracked for the
`@vlcn.io/ws-server`/`ws-client` fork.


silviot:
Closing without merging. This was a point-in-time audit snapshot; `docs/architecture/` should hold durable architecture docs, not a dated bug list that goes stale the moment the fixes (#1248, #1249, fork PR) land. The findings live in Linear D-312 (kept current), the fix rationale in each PR + inline code comments, and the durable test methodology in the sync-convergence-test / sync-engine-regression-test skills. The full snapshot remains on the branch for reference.

PR #1274, opened 2026-08-08
feat(web-client): workstation-based default note names (D-608)

silviot:
New notes are now named after the workstation they're created on ("Sale Front desk", "Vendita Cassa") instead of the global "New Sale (2870)" counter. Each device gets a name: randomly generated on first use, stored in localStorage (never synced), editable on the settings page. Together with the per-site id blocks of #1273 this makes concurrent multi-device work legible: ids keep documents distinct, names make the origin readable at a glance. **One deliberate behavior change: the "(n)" suffix now counts only active (uncommitted) drafts**, so the counter restarts once same-named drafts are committed/renamed; committed notes in history are told apart by their timestamps, and the unbounded "(2870)"-style growth is gone.

## What changed

- New [`site workstation name`](https://github.com/librocco/librocco/blob/86d353658a3053df32642f83913a0330e94a45cd/apps/web-client/src/lib/utils/note-names.ts) flow: `deviceSettingsStore` gains `workstationName` (new field on the settings page, seeded lazily with a localized random default like "Workstation 374" / "Postazione 374").
- UI note creation passes a workstation-derived base name; [`getSeqNameForBase`](https://github.com/librocco/librocco/blob/86d353658a3053df32642f83913a0330e94a45cd/apps/web-client/src/lib/db/cr-sqlite/note.ts#L106-L131) suffixes it against active same-kind notes (LIKE-escaped, so names with `%`/`_` behave).
- i18n: `common.new_note_names.{sale,purchase}` and workstation settings labels in en/it/de.
- e2e seeds a deterministic workstation name ("Alpha") via `storageState`; naming-sequence specs rewritten to the new semantics; new `workstation_name.spec.ts` covers renaming via settings.

## Why it's correct

- The db-level create functions keep the legacy "New Sale (n)" naming when no base is passed, so API-level callers and existing tests are untouched; only UI creation paths opt in.
- Names are pure display data: no schema change, no sync impact (the workstation name itself never syncs; only the resulting note names do, like any display name).

## What this does NOT do

- Reconciliation notes keep their timestamp names; warehouse default names keep the legacy "New Warehouse (n)" sequence.
- Two devices left on the same default-random name would produce same-named (never same-id) notes; the settings field is the remedy.

## Tests

Unit: suffix scoping, active-only restart, inbound/outbound separation, LIKE-escaping ([note.test.ts](https://github.com/librocco/librocco/blob/86d353658a3053df32642f83913a0330e94a45cd/apps/web-client/src/lib/db/cr-sqlite/__tests__/note.test.ts)). E2E: rewritten naming-sequence specs + settings-rename flow, all passing locally (chromium).

---
Closes D-608

---
Stacked on #1273 (D-595 per-site id blocks); merges after it.

### DIFF
--- a/apps/e2e/integration/inbound_flow.spec.ts
+++ b/apps/e2e/integration/inbound_flow.spec.ts
@@ -27,9 +27,9 @@ test('should create a new inbound note, under the particular warehouse, on wareh
 	// Create a new note under "Warehouse 1"
 	await content.entityList("warehouse-list").item(0).getByRole("button", { name: "New Purchase" }).click();
 
-	// Check that we've been redirected to the new note's page
+	// Check that we've been redirected to the new note's page (named after the workstation)
 	await dashboard.view("inbound-note").waitFor();
-	await page.getByRole("main").getByRole("heading", { name: "New Purchase" }).first().waitFor();
+	await page.getByRole("main").getByRole("heading", { name: "Purchase Alpha" }).first().waitFor();
 });
 
 test("should display notes, namespaced to warehouses, in the inbound note list", async ({ page }) => {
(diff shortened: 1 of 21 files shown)


silviot:
Follow-up commit e9e5e32e (from the codex adversarial review): double-activation guard on the sale-commit confirm (a second concurrent run minted a duplicate reconciliation note = phantom stock), 60-char cap on the workstation name (schema + input), storage-failure fallback so note creation never blocks on localStorage, and a 10x wider random default-name space.

chiruzzimarco (approved):

PR #1275, opened 2026-08-08
feat(sync): client version handshake + server-side deny gate (D-609)

silviot:
Sync clients now announce their build version on connect, and the sync server can refuse the WebSocket upgrade (HTTP 401) for versions on a deny list — turning "reload every till together after a deploy" into an enforced invariant. **The gate is inert until configured**: `DENIED_CLIENT_VERSIONS` is unset in production today, so this PR changes nothing observable except a per-connection version log line; the lever is armed when the D-595 stack deploys. **One honest limit, per the codex risk review: the gate stops live mixing but does not quarantine a denied till's local writes** — it keeps selling offline and its rows sync once re-admitted, so the runbook says reload/remediate before lifting a denial.

## What changed

- Client: [`withClientVersion`](https://github.com/librocco/librocco/blob/a67aa6ce20f00c322c807f52a828dfb4ce5b7052/apps/web-client/src/lib/utils/sync-url.ts) appends `client_version=<git sha>` to the sync WS URL at the single point where sync starts (`AppSyncCore.start`).
- Server: new [`version-gate.ts`](https://github.com/librocco/librocco/blob/a67aa6ce20f00c322c807f52a828dfb4ce5b7052/apps/sync-server/src/version-gate.ts) plugs into the `authenticate` hook the vendored `@vlcn.io/ws-server` already exposes — no fork changes. Every connection's version+room is logged; listed versions get 401 before any sync protocol runs. Deny logs are rate-limited per version+room (a denied client retries every 1–3s).
- Config: `DENIED_CLIENT_VERSIONS` (comma-separated; special token `unversioned` denies pre-gate clients). `/health` reports the deny list the running process actually loaded.
- [README runbook](https://github.com/librocco/librocco/blob/a67aa6ce20f00c322c807f52a828dfb4ce5b7052/apps/sync-server/README.md): activation/rollback incl. the launcher env-snapshot gotcha, `unversioned` rollout ordering (client first, verify logged versions, then arm), staffed-window advice.

## Why it's safe

- Deny-list unset ⇒ identical to today's `noopAuth` path; old server + new client and new server + old client both work (extra query param is ignored; Caddy preserves the query string).
- Denied at the upgrade, so a blocked client exchanges zero sync messages; no partial-write state.
- Rejection loops land in the existing rapid-close "sync stuck" detection, so a blocked till shows stuck rather than nothing.

## What this does NOT do

- No client-side "your build is blocked, reload" message — the pre-upgrade 401 carries no readable reason in the browser WS API. Follow-up if wanted: a version-check HTTP endpoint the client polls when stuck.
- No quarantine of a denied build's local writes (see lead). True prevention needs remediation-before-readmission, which is a procedure, not a header.
- Deny, not read-only mode (ticket left this open; refusal is simpler and loud).

## Tests

[`version-gate.test.ts`](https://github.com/librocco/librocco/blob/a67aa6ce20f00c322c807f52a828dfb4ce5b7052/apps/sync-server/src/version-gate.test.ts): deny-list parsing, version extraction, gate decisions, deny-log rate limiting, plus an integration test driving a real HTTP upgrade through `attachWebsocketServer` with the gate in the positional `authenticate` slot (denied → 401, allowed → 101, unversioned+sentinel → 401). Client-side URL decoration unit-tested. Note: `ws-server-reject-recovery.test.ts` currently fails on `main` too (pre-existing tarball drift, unrelated — worth its own look).

Codex reviews: correctness pass clean; risk-focused pass drove the operability commit (health exposure, log rate-limit, runbook) — full findings in the PR conversation.

Closes D-609


silviot:
Codex review (gpt-5.6-sol) ran twice on this branch:

**Pass 1 — correctness (default review):** clean. "The version parameter is consistently added by the client and correctly evaluated through the WebSocket server authentication hook. No actionable correctness regressions were identified."

**Pass 2 — risk-focused (shop-must-go-on framing):**

- **[P1] A denied till silently enters an unlimited reconnect loop** — the pre-upgrade 401 isn't observable via the browser WS API; the till shows only generic reconnecting. *Partially addressed:* deny logs are now rate-limited server-side; a client-visible "build blocked, reload" state needs a new channel (follow-up).
- **[P1] The gate does not quarantine a denied build's local writes** — it keeps writing offline and those rows sync once the denial is lifted. *Addressed as procedure:* README runbook now requires reload/remediation before re-admission; the PR lead states this limit.
- **[P2] Env-only config is awkward to roll back under the launcher** (env snapshot). *Addressed:* `/health` now reports the loaded deny list; README documents the relaunch procedure.
- **[P2] `unversioned` can disconnect every old till at once on a server-first rollout.** *Addressed:* README rollout ordering — ship the announcing client first, verify logged versions from every register, then arm the sentinel in a staffed window.

With the deny list unset codex found no selling-path regression: query string survives Caddy, path regex still matches, old-server/new-client and new-server/old-client both compatible.

chiruzzimarco (approved):

